PR #362, opened 2016-07-05
scripts can be run from anywhere

### LINKED ISSUE #361: scripts should not require the cwd to be the scripts directory
I attempted to run `scripts/npm-link-demo-shell.sh` from <alfresco-ng2-components-clone-dir> and the script failed because it expects to be run from the scripts directory.

binduwavell:
updated all scripts except update-version.sh to be able to
run from anywhere. Previously you had to be in the scripts
directory to run them.

fixes #361


binduwavell:
I have no idea how coverage could decrease given I just forked master and patched these changes in and pushed them. And these changes have nothing to do with coverage... Coveralls is not my friend :)

PR #442, opened 2016-07-18
Dev eromano webscriptcomponent

eromano:


ng2-components/ng2-alfresco-datatable/src/data/object-datatable-adapter.ts
```diff
@@ -32,6 +32,25 @@ export class ObjectDataTableAdapter implements DataTableAdapter {
     private _rows: DataRow[];
     private _columns: DataColumn[];
 
+    static generationSchema(rowToExaminate: any) {
```
DenisVuyka: I think this could be named 'generateSchema'

PR #458, opened 2016-07-22
Process list component

wabson:


ng2-components/ng2-activiti-processlist/LICENSE
```diff
@@ -0,0 +1,13 @@
+Copyright 2016 Will Abson <will.abson@alfresco.com>
```
DenisVuyka: Should it be 'Alfresco Software' header instead?

wabson: The generator did this, it is fixed in 02ce562a

PR #460, opened 2016-07-22
About page with package versions

DenisVuyka:
- new About page with alfresco package versions


DenisVuyka:
Introduces new `About` page for demo app

screen shot 2016-07-22 at 11 51 11

PR #490, opened 2016-07-28
Improve scripts and script docs

wabson:


magemello:
cool staff, in some package.json the postinstall is missing....can we add it?

Issue #499, opened 2016-08-01
Task list demo doesn't start


VitoAlbano:
trying to start the demo for the ng2-activiti-tasklist component , the demo doesn't show up.

PR #558, opened 2016-08-11
Fix demo and unit test

mauriziovitale:
ref #499


mauriziovitale:
The test problems are referred to the old version of the alfresco js API on NPM. Using the last version, there are no test problems.

Issue #621, opened 2016-08-25
Attach control for linking to respository content from Alfresco


wabson:
To complement the new upload control added by @DenisVuyka we need to be able to select from content items that are already stored in Alfresco when such a source has been configured in the Activiti form instead of the local file source

hashraf88:
File does not display if previously uploaded and cleared
1. Upload a file
2. Save form
3. Navigate away and back to task
4. Clear previously uploaded file
5. Select same file or different file

**Expected results**
File chosen is displayed

**Actual results**
File chosen is not displayed

[upload widget.mp4.zip](https://github.com/Alfresco/alfresco-ng2-components/files/489388/upload.widget.mp4.zip)

hashraf88:
Above reported in #812

PR #622, opened 2016-08-25
Add new attached-file widget for linking to external files

wabson:
Support linked content from Alfresco that is never copied into Activiti but just linked from the process. Does not yet support copying the content into Activiti, but this would be a small improvement to make.

In the future the 'attach' and 'upload' widgets should be wrapped with a single widget, since it is possible for the Activiti 'upload' field to allow the user to choose between uploading a local file OR copying from Alfresco so both options need to be available to them.

Refs #621

### DIFF
--- a/ng2-components/ng2-activiti-form/src/components/widgets/attach/attach.widget.css
+++ b/ng2-components/ng2-activiti-form/src/components/widgets/attach/attach.widget.css
@@ -7,11 +7,9 @@
 }
 
 .attach-widget__file {
-    /*float: left;*/
     margin-top: 4px;
 }
 
 .attach-widget__reset {
-    /*float: left;*/
     margin-top: 4px;
 }
--- a/ng2-components/ng2-activiti-form/src/components/widgets/attach/attach.widget.css
+++ b/ng2-components/ng2-activiti-form/src/components/widgets/attach/attach.widget.css
@@ -1,0 +1,17 @@
+.attach-widget {
+    width:100%
+}
+
+.attach-widget__icon {
+    float: left;
+}
+
+.attach-widget__file {
+    /*float: left;*/
+    margin-top: 4px;
+}
+
+.attach-widget__reset {
+    /*float: left;*/
+    margin-top: 4px;
+}
--- a/ng2-components/ng2-activiti-form/src/services/form.service.ts
+++ b/ng2-components/ng2-activiti-form/src/services/form.service.ts
@@ -18,16 +18,13 @@
 import { Injectable } from '@angular/core';
 import { Observable } from 'rxjs/Rx';
 import 'rxjs/add/operator/map';
-import { AlfrescoApiService, AlfrescoSettingsService } from 'ng2-alfresco-core';
+import { AlfrescoApiService } from 'ng2-alfresco-core';
 import { FormValues } from './../components/widgets/core/index';
 import { FormDefinitionModel } from '../models/form-definition.model';
 import { EcmModelService } from './ecm-model.service';
 import { GroupModel } from './../components/widgets/core/group.model';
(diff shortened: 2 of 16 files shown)


ng2-components/ng2-activiti-form/src/services/form.service.ts
```diff
@@ -101,6 +106,67 @@ export class FormService {
...
+     *
+     * @param accountId
+     * @param nodeId
+     * @returns {null}
+     */
+    getAlfrescoNodes(accountId: string, nodeId: string): Observable<[ExternalContent]> {
+        let accountPath = accountId.replace('-', '/');
+        let headers = new Headers();
+        headers.append('Authorization', this.authService.getTicketBpm());
+        return this.http.get(
```
eromano: why do we need to use this http call instead of the JS-API?

ng2-components/ng2-activiti-form/src/services/form.service.ts
```diff
@@ -17,16 +17,21 @@
...
 import {Injectable} from '@angular/core';
 import {Observable} from 'rxjs/Rx';
-import {AlfrescoAuthenticationService} from 'ng2-alfresco-core';
+import 'rxjs/add/operator/map';
+import {AlfrescoAuthenticationService, AlfrescoSettingsService} from 'ng2-alfresco-core';
 import {FormValues} from './../components/widgets/core/index';
+import {ExternalContent} from '../components/widgets/core/external-content';
+import {ExternalContentLink} from '../components/widgets/core/external-content-link';
 
+import {Http, Response, RequestOptions, Headers} from '@angular/http';
```
eromano: If we use the JS-API we can remove all these import

ng2-components/ng2-activiti-form/src/services/form.service.ts
```diff
@@ -101,6 +106,67 @@ export class FormService {
...
+     * @param node
+     * @param siteId
+     * @returns {null}
+     */
+    linkAlfrescoNode(accountId: string, node: ExternalContent, siteId: string): Observable<ExternalContentLink> {
+        let headers = new Headers();
+        headers.append('Content-Type', 'application/json');
+        headers.append('Authorization', this.authService.getTicketBpm());
+        return this.http.post(
+            `${this.alfrescoSettingsService.bpmHost}/activiti-app/app/rest/content`,
```
eromano: same as line before

ng2-components/ng2-activiti-form/src/components/widgets/attach/attach.widget.css
```diff
@@ -0,0 +1,17 @@
+.attach-widget {
+    width:100%
+}
+
+.attach-widget__icon {
+    float: left;
+}
+
+.attach-widget__file {
+    /*float: left;*/
```
eromano: commented css

ng2-components/ng2-activiti-form/src/services/form.service.ts
```diff
@@ -209,6 +214,66 @@ export class FormService {
...
     }
 
+    /**
+     * Returns a list of child nodes below the specified folder
+     *
+     * @param accountId
+     * @param nodeId
+     * @returns {null}
+     */
+    getAlfrescoNodes(accountId: string, nodeId: string): Observable<[ExternalContent]> {
```
magemello: We have a node.service in the same folder, this method should be there

wabson: I had to use the Activiti service to get the nodes because there is no guarantee that the repository linked to in Activiti is the same one as the demo-shell, also these APIs are needed to create the 'related content' item in Activiti. The code above is now gone anyway and the JS API is used for getting the list of nodes as well as creating the related content.

PR #629, opened 2016-08-26
ECM node to BPM form & BPM form to ECM node #628

magemello:


demo-shell-ng2/app/components/activiti/form-viewer.component.html
```diff
@@ -1,3 +1,11 @@
 <div class="activiti-form-viewer" *ngIf="taskId">
     <activiti-form [taskId]="taskId"></activiti-form>
+    <!--<activiti-form [formName]="'activitiForms:patientFolder'"
```
eromano: do we need this commented HTML?

magemello: yep, It shows how the component works....I will put some logic to make the user choosing how the component has to behave

PR #651, opened 2016-08-31
add video support 'mp4'  'WebM' 'Ogg'

eromano:


DenisVuyka:
To improve mobile support (especially screen orientation change) you could use something like the following:

```css
.video-player {
  width: 100%;
  height: auto;
}

/* portrait */
@media screen and (orientation:portrait) {
	/* portrait-specific styles */
}
/* landscape */
@media screen and (orientation:landscape) {
	/* landscape-specific styles */
  .video-player {
    height: 100%;
  }
}
```

This was tested on iOS/Safari and proved to be working good.

eromano:
I have used a different approach but seems to be fine

PR #714, opened 2016-09-08
onError event for form component

magemello:


DenisVuyka:
You can replace 
```ts
(error) => {
this.handleError(error);
 }
```
with just 
```ts
this.handleError;
```

magemello:
tried this approach doesn't work...I get undefined

PR #733, opened 2016-09-12
Task without a Form is handled incorrectly

### LINKED ISSUE #724: Task without a Form is handled incorrectly
Provide support for displaying a 'missing' form for a Task. For the moment such task raises exception when clicked in the Task list.

DenisVuyka:
fixes #724 

due to some reason Activiti stores/serializes null as a “null” string
rather than value for a ‘formKey’ property; added checks for this case


DenisVuyka:
@rmukesh I suggest raising a separate issue for the case you've mentioned, thanks.

DenisVuyka:
@rmukesh thanks for pointing out, we have created a separate issue (#794) to address that

Issue #671, opened 2016-09-05
Regardless of search criteria search results are always same


hashraf88:
1. Search for any text
2. Notice results
3. Search for different text
4. Notice results

At 3 and 4 results are the same 

n.b. No error in console
[searchItems.mp4.zip](https://github.com/Alfresco/alfresco-ng2-components/files/454821/searchItems.mp4.zip)

PR #735, opened 2016-09-12
Fix wrong reference to search term

VitoAlbano:
#671


DenisVuyka:
@VitoAlbano Could you please create some unit test to cover this scenario and avoid regression in the future?

VitoAlbano:
@DenisVuyka  no problem!

Issue #726, opened 2016-09-10
Form Validation - Activiti form component


magemello:
Introduce form validation support

- Text
  - [x] Required
  - [x] Min Length
  - [x] Max Length
  - [x] RegEx Pattern
- Multiline Text
  - [x] Required
  - [x] Min Length
  - [x] Max Length
  - [x] RegEx Pattern
- Number
  - [x] Valid Number Value
  - [x] Required
  - [x] Min Value
  - [x] Max Value 
- Dropdown
  - [x] Required
- Typeahead
  - [x] Required
- Radio buttons
  - [x] Required
- People
  - [x] Required
- Group of people
  - [x] Required
- Attach
  - [x] Required

hashraf88:
In Activiti for widgets with max length set no text is allowed to be entered after max length reached.
In the Activiti component once the max length is reached more text is allowed to be entered however a message is displayed 'Should be x characters maximum.'
Is this ok or would Activiti behaviour need to replicated?

DenisVuyka:
We can't replicate full behaviour of Activiti components at this point (styles, input masks, etc.). For now just pure validation.

hashraf88:
Are we introducing validation (required) for the checkbox widget too? Currently it seems we do not

DenisVuyka:
Checkbox cannot have required validation as it has 2 states - checked and unchecked and both states are valid.

hashraf88:
Ah yes true

PR #752, opened 2016-09-15
#726 Form Validation

DenisVuyka:
refs #726 

Also fixes numerous bugs related to form widgets.

### DIFF
--- a/ng2-components/ng2-activiti-form/src/components/widgets/widget.component.ts
+++ b/ng2-components/ng2-activiti-form/src/components/widgets/widget.component.ts
@@ -36,6 +36,13 @@ export class WidgetComponent implements AfterViewInit {
         return this.field ? true : false;
     }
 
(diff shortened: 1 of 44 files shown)


ng2-components/ng2-activiti-form/src/components/widgets/widget.component.ts
```diff
@@ -36,6 +36,13 @@ export class WidgetComponent implements AfterViewInit {
         return this.field ? true : false;
     }
 
+    isRequired(): any {
```
mauriziovitale: Should return a boolean instead of any

DenisVuyka: This is a special case as I'm binding "required" attribute. This function returns "null" that upon binding removes "required" from html. Otherwise it will be "required"="false" and will still be treated as "required"

PR #820, opened 2016-09-26
Flag to disable csrf

magemello:


PREREQUISITES.md
```diff
@@ -99,4 +22,8 @@ $ node -v
 v5.12.0
 ```
 
+## Configure Nginx
+
```
enzor: This pr assumes that our users are starting Activiti on port 9999.
Same that they start alfresco on port 8080 and the demo shell on 3000.
This should be specified somewhere on the documentation how to configure this

PR #860, opened 2016-10-05
Fix dialogs for firefox and safari

magemello:


DenisVuyka:
This won't properly work unless dialog is direct child of the document body element. You can try getting dialog multiple times or clicking buttons - it stops working

magemello:
I tested  It and it works...because the change at the CSS property

Issue #850, opened 2016-10-05
Auto-create default process-instance filters


wabson:
In the task component if the user has no filters configured then a default set is automatically created by the service. The processlist needs to do the same, so that it is not necessary to use the Activiti UI to do this.

Issue #849, opened 2016-10-05
Add more processlist tests


wabson:
Tests were disabled following upgrade to Angular 2.0.0

PR #864, opened 2016-10-07
Auto-create default process-instance filters

wabson:
Refs #850


eromano:
The coverage of is only 61% activiti-process.service.spec.ts maybe you can add more tests

wabson:
Added more process service tests so that filter-related things are tested more exhaustively. Other things implemented by this service (get process instances, comments, tasks, cancel process instance etc.) will have tests added as part of #849

PR #868, opened 2016-10-10
Added fix for visibility on dropdown

VitoAlbano:


eromano:
Vito please fix all the code formatting , have to be compliant with the code already present.

Issue #799, opened 2016-09-22
Create a Reports list - Analytics component


magemello:
The component needs to display all the existing reports.

- [ ] Process definition heat map
- [ ] Process definition overview
- [ ] Process instances overview
- [ ] Task overview
- [ ] Task service level agreement

Issue #791, opened 2016-09-21
Process definition overview - Analytics component


magemello:
Reproduce the Process definition overview graph of Activiti.

hashraf88:
If a specific app is selected when accessing the process definition report all process definitions for all apps are displayed, occurs for all process statuses.

mauriziovitale:
@hashraf88  I'm not able to reproduce your scenario. Can you test it again ?

Issue #789, opened 2016-09-21
 Process instances overview - Analytics component


magemello:
Reproduce the Process instances heat map graph of Activiti.

hashraf88:
The process instance counts section of the report is missing.
![screen shot 2016-10-17 at 15 27 43](https://cloud.githubusercontent.com/assets/13200338/19441936/73109130-947f-11e6-892e-222774dbc75d.png)

hashraf88:
For the process steps total time section if the sum of the activity is 0 is doesnt need to be displayed as a key in on the pie chart. 

**Component**
![screen shot 2016-10-17 at 15 47 43](https://cloud.githubusercontent.com/assets/13200338/19442733/449b1a8e-9482-11e6-8f57-0b8772d84723.png)

**Activiti**
![screen shot 2016-10-17 at 15 47 29](https://cloud.githubusercontent.com/assets/13200338/19442530/7f7c755e-9481-11e6-9f93-680b9a643434.png)

magemello:
It is fine, we can keep it as it is.

Issue #788, opened 2016-09-21
Task overview - Analytics component


magemello:
Reproduce the Task overview graph of Activiti.

hashraf88:
Don't think the timestamp is required for the number of tasks divided by date interval bar chart 
**COMPONENT**
<img width="897" alt="screen shot 2016-10-17 at 16 16 12" src="https://cloud.githubusercontent.com/assets/13200338/19443601/1d113a5e-9485-11e6-8bf7-6a5931aae477.png">

**ACTIVITI**
![screen shot 2016-10-17 at 16 17 24](https://cloud.githubusercontent.com/assets/13200338/19443633/37830354-9485-11e6-9055-f71e291d72ed.png)

PR #871, opened 2016-10-10
Create report list and report charts

mauriziovitale:
Create a Reports list ref #799 
Process definition overview ref #791 
Process instances overview ref #789 
Task overview ref #788


eromano:
It seems to be  fine can you just  increase the coverage? thanks

PR #881, opened 2016-10-11
replace the getApiService with the apiService in the tagService

eromano:


ng2-components/ng2-alfresco-tag/src/services/tag.service.ts
```diff
@@ -29,32 +30,32 @@ export class TagService {
 
     /**
      * Constructor
-     * @param authService
+     * @param apiService
      */
-    constructor(public authService: AlfrescoAuthenticationService) {
+    constructor(private authService: AlfrescoAuthenticationService, private apiService: AlfrescoApiService) {
```
DenisVuyka: Do you need `authService` anywhere in the code with the migration to `apiService`? I guess you can get rid of the `authService` reference.

PR #891, opened 2016-10-12
Improve analytics component unit tests

mauriziovitale:
ref #791


ng2-components/ng2-activiti-analytics/src/components/widgets/duration/duration.widget.html
```diff
@@ -5,6 +5,7 @@
                    type="text"
                    pattern="-?[0-9]*(\.[0-9]+)?"
                    [attr.id]="field.id"
+                   [value]="field.value"
```
magemello: do we need both? value and ngModel?

mauriziovitale: Yes, we need both. Otherwise I can't unit test the component.

let numberElement: any = element.querySelector('#duration');
expect(numberElement.value).toEqual('0');

magemello: the problem is that you are unit testing it hoping that value will always be equal to ngModel, but if  someone changes the value passed in ngModel your unit test will keep been green because you are testing the other filed....

PR #901, opened 2016-10-13
Unit tests for Activiti Form component

DenisVuyka:
includes several bug fixes and code improvement based on unit testing results

### DIFF
--- a/ng2-components/ng2-activiti-form/src/components/activiti-form.component.spec.ts
+++ b/ng2-components/ng2-activiti-form/src/components/activiti-form.component.spec.ts
@@ -21,6 +21,7 @@ import { ActivitiForm } from './activiti-form.component';
 import { FormModel, FormOutcomeModel, FormFieldModel, FormOutcomeEvent } from './widgets/index';
 import { FormService } from './../services/form.service';
 import { WidgetVisibilityService } from './../services/widget-visibility.service';
+import { NodeService } from './../services/node.service';
 
 describe('ActivitiForm', () => {
 
@@ -28,6 +29,7 @@ describe('ActivitiForm', () => {
     let formService: FormService;
     let formComponent: ActivitiForm;
     let visibilityService:  WidgetVisibilityService;
+    let nodeService: NodeService;
 
     beforeEach(() => {
         componentHandler = jasmine.createSpyObj('componentHandler', [
@@ -39,7 +41,8 @@ describe('ActivitiForm', () => {
         window['componentHandler'] = componentHandler;
 
         formService = new FormService(null, null);
-        formComponent = new ActivitiForm(formService, visibilityService, null, null);
+        nodeService = new NodeService(null);
+        formComponent = new ActivitiForm(formService, visibilityService, null, nodeService);
     });
 
     it('should upgrade MDL content on view checked', () => {
@@ -635,12 +638,22 @@ describe('ActivitiForm', () => {
     });
 
     it('should load form for ecm node', () => {
-        spyOn(formComponent, 'loadFormForEcmNode').and.stub();
+        let metadata = {};
+        spyOn(nodeService, 'getNodeMetadata').and.returnValue(
+            Observable.create(observer => {
+                observer.next({ metadata: metadata });
+                observer.complete();
+            })
+        );
+        spyOn(formComponent, 'loadFormFromActiviti').and.stub();
 
-        formComponent.nodeId = '<id>';
+        const nodeId = '<id>';
+        formComponent.nodeId = nodeId;
         formComponent.ngOnInit();
 
-        expect(formComponent.loadFormForEcmNode).toHaveBeenCalled();
+        expect(nodeService.getNodeMetadata).toHaveBeenCalledWith(nodeId);
+        expect(formComponent.loadFormFromActiviti).toHaveBeenCalled();
+        expect(formComponent.data).toBe(metadata);
     });
 
     it('should disable outcome buttons for readonly form', () => {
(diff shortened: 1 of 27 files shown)


ng2-components/ng2-activiti-form/src/components/widgets/attach/attach.widget.spec.ts
```diff
@@ -0,0 +1,293 @@
...
+        };
+
+        widget.showDialog();
+        expect(modalShown).toBeTruthy();
+    });
+
+    it('should select folder and load nodes', () => {
+        let nodes = [{}];
+        spyOn(contentService, 'getAlfrescoNodes').and.callFake(() => {
+            return Observable.create(observer => {
```
wabson: Could `Observer.from()` simplify the creation? And possibly spy's `returnValue` rather than `callFake()`?

ng2-components/ng2-activiti-form/src/components/activiti-form.component.spec.ts
```diff
@@ -634,4 +634,74 @@ describe('ActivitiForm', () => {
         expect(visibilityService.refreshVisibility).toHaveBeenCalledWith(field.form);
     });
 
+    it('should load form for ecm node', () => {
+        spyOn(formComponent, 'loadFormForEcmNode').and.stub();
+
+        formComponent.nodeId = '<id>';
+        formComponent.ngOnInit();
+
+        expect(formComponent.loadFormForEcmNode).toHaveBeenCalled();
```
DenisVuyka: I think that complicates testing. This test checks `ngOnInit` and `nodeId` condition.

ng2-components/ng2-activiti-form/src/components/widgets/attach/attach.widget.spec.ts
DenisVuyka: Do you mean changing this block in all tests? We have plenty of this snippets copy-pasted everywhere.

ng2-components/ng2-activiti-form/src/components/widgets/display-value/display-value.widget.spec.ts
```diff
@@ -0,0 +1,542 @@
...
+
+    it('should require field to setup default value', () => {
+        widget.field = null;
+        widget.ngOnInit();
+        expect(widget.value).toBeUndefined();
+    });
+
+    it('should take field value on init', () => {
+        let value = '<value>';
+        widget.field = new FormFieldModel(null, { value: value });
```
wabson: Could the mock `FormFieldModel` instances used in these tests be declared in an external file to improve readability?

DenisVuyka: Not sure I got the question. The `FormFieldModel` is a model class used in Form renderer, it's not a mock.

ng2-components/ng2-activiti-form/src/components/widgets/widget.component.ts
```diff
@@ -63,6 +65,20 @@ export class WidgetComponent implements AfterViewInit {
         return false;
     }
 
+    setupMaterialTextField(elementRef: ElementRef, handler: any, value: string): boolean {
+        if (elementRef && handler) {
+            let el = elementRef.nativeElement;
```
wabson: Is there a benefit to using Angular element querying using `By` instead of `nativeElement.querySelector()` method?

DenisVuyka: Testing purposes only. It's sometimes enough mocking the elementRef (without component builder and dependency injectors) that has no 'nativeElement'

ng2-components/ng2-activiti-form/src/components/widgets/display-value/display-value.widget.spec.ts
wabson: I was proposing moving the instances of `FormFieldModel` that you create into a separate file and import them into the test via a meaningful name. This will aid readability of the tests in this file and allow you over time to refactor the instances further to remove duplicated values or use helpers to construct them.

DenisVuyka: I suggest we postpone that until classes/types are introduced for all form fields, that will allow using strongly-typed classes without need for mocks. This is something to address when we work on stencils.

ng2-components/ng2-activiti-form/src/components/activiti-form.component.spec.ts
wabson: Perhaps this needs further discussion among the team, then. My position is to have a minimal public interface on the class to allow for later refactoring without having to re-write the tests and ensures we can commit to not break public/protected methods that others bind to in a later (non-major) release. There should be ways to avoid this increasing complexity of the tests too much, e.g. refactoring private methods into external classes which can then be tested themselves via their public methods.

DenisVuyka: Test will be updated in the upcoming commit.

wabson:
Nice changes, thanks. I wonder with the observables you changed though is it possible to use `Observer.from(value)` though rather than `Observable.create(observer => { observer.next(value); observer.complete(); });`? This [guide](https://github.com/ReactiveX/RxJava/wiki/Creating-Observables) is for Java ReactiveX impl rather than rxjs but the methods seem the same and here they say use `create()` for advanced purposes only.

DenisVuyka:
You refer to the Java library, not JavaScript one, but they seem to have similar api as well: https://github.com/Reactive-Extensions/RxJS/blob/master/doc/api/core/operators/from.md

DenisVuyka:
@wabson Using `Observable.from` will require again discarding previous refactoring with `spyOn` and getting back to `callFake` as you can't properly use it within `returnValue`

DenisVuyka:
@wabson I can't make `Observable.from` work properly with TypeScript.

Issue #916, opened 2016-10-22
improve import alfresco-js-api 


eromano:
All the demo and the demoshell import the alfresco-js-api in the HTML as script and in the system.js through the loader. 
Improving the declaration file of the alfresco-js-api we can leave only the load inside systemjs.
Now if you are going to remove the import from the index the alfresco-js-API doesn't work.

PR #917, opened 2016-10-22
fix importing alfreso-js-api compatible with the version 0.3.6 #916

eromano:


eromano:
This PR is going to be green when the next version of alfresco-js-api will be released

DenisVuyka:
It will never work in its current state due to wrong imports:

screen shot 2016-10-29 at 09 26 57

eromano:
this commit should fix he problem https://github.com/Alfresco/alfresco-js-api/issues/78

PR #920, opened 2016-10-25
Fix Visibility check for equal dates

### LINKED ISSUE #896: if visibility condition of date widget set to equal date and smae date entered, condition found not to be met
Regarding date widget, if visibility conditions **before or on same date** or **after or on same date** are set on a widget and same date is entered on field condition is not met when should be.

In below example visibility condition is set to 'after or on same date' with value being '1-10-2016'
**Condition**
screen shot 2016-10-13 at 10 57 07

**Activiti**
screen shot 2016-10-13 at 10 56 20

**Component**
screen shot 2016-10-13 at 10 45 57

[form_with_all_widgets.json.zip](https://github.com/Alfresco/alfresco-ng2-components/files/526790/form_with_all_widgets.json.zip)

VitoAlbano:
Fix #896

### DIFF
--- a/ng2-components/ng2-activiti-form/src/components/widgets/date/date.widget.spec.ts
+++ b/ng2-components/ng2-activiti-form/src/components/widgets/date/date.widget.spec.ts
@@ -18,6 +18,9 @@
 import { ElementRef } from '@angular/core';
 import { DateWidget } from './date.widget';
 import { FormFieldModel } from './../core/form-field.model';
+import { FormModel } from './../core/form.model';
+import { CoreModule } from 'ng2-alfresco-core';
+import { ComponentFixture, TestBed, async } from '@angular/core/testing';
 
 describe('DateWidget', () => {
 
@@ -112,7 +115,7 @@ describe('DateWidget', () => {
     });
 
     it('should update field value on date selected', () => {
-        widget.field = new FormFieldModel(null, { type: 'date' });
+        widget.field = new FormFieldModel(null, {type: 'date'});
         widget.ngOnInit();
 
         let date = '13-3-1982';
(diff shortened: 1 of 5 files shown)


ng2-components/ng2-activiti-form/src/components/widgets/date/date.widget.spec.ts
```diff
@@ -136,11 +143,127 @@ describe('DateWidget', () => {
...
+
+    it('should send field change event when a new date is picked from data picker', (done) => {
+        let w = new DateWidget(null);
+        spyOn(w, 'setupMaterialTextField').and.callThrough();
+        w.field = new FormFieldModel(null, {type: 'date'});
+        w.ngOnInit();
+        w.datePicker.time = moment();
+        w.fieldChanged.subscribe((field) => {
+            expect(field).toBeDefined();
+            expect(field).not.toBeNull();
```
mauriziovitale: You should check not only if is defined and not null but also the value.

ng2-components/ng2-activiti-form/src/components/widgets/date/date.widget.spec.ts
```diff
@@ -136,11 +143,127 @@ describe('DateWidget', () => {
...
+
+    it('should send field change event when date is changed in input text', (done) => {
+        let w = new DateWidget(null);
+        spyOn(w, 'setupMaterialTextField').and.callThrough();
+        w.field = new FormFieldModel(null, {type: 'date'});
+        w.ngOnInit();
+        w.datePicker.time = moment();
+        w.fieldChanged.subscribe((field) => {
+            expect(field).toBeDefined();
+            expect(field).not.toBeNull();
```
mauriziovitale: You should check not only if is defined and not null but also the value.

mauriziovitale: You should check not only if is defined and not null but also the value.

Issue #371, opened 2016-07-06
Cannot tab into search results


hashraf88:
Cannot tab into search results and so cannot to read by screen reader and so cannot be selected 

![screen shot 2016-07-06 at 14 17 27](https://cloud.githubusercontent.com/assets/13200338/16619317/88575cae-4385-11e6-8137-ea4c823165be.png)

VitoAlbano:
depends on : #671

hashraf88:
Unable to test due to #934

PR #927, opened 2016-10-26
Tab into search results

wabson:
Refs #371


ng2-components/ng2-alfresco-search/src/components/alfresco-search-autocomplete.component.ts
```diff
@@ -40,15 +40,29 @@ export class AlfrescoSearchAutocompleteComponent implements OnInit, OnChanges {
...
     ngClass: any;
 
+    @Input()
+    maxResults: number = 5;
+
     @Output()
     preview: EventEmitter<any> = new EventEmitter();
 
     @Output()
+    focusEmitter: EventEmitter<FocusEvent> = new EventEmitter<FocusEvent>();
```
DenisVuyka: this will end with odd html markup as for me: `<element (focusEmitter)="handler">`
Can name be changed to something like `focus` or `onFocus` to be more html-friendly?

PR #996, opened 2016-11-03
Add browser support 

hashraf88:


hashraf88:
Not sure how that happend... will delete and redo

Issue #946, opened 2016-10-28
4 empty buttons on activiti app


hashraf88:
Within the activiti app in general there seems to be 4 empty buttons 
Unfortunately I cannot seem to see which element the WAVE tool is pointing to
N.b. the screenshot below is with no styles selected in the WAVE tool

![screen shot 2016-10-28 at 17 37 27](https://cloud.githubusercontent.com/assets/13200338/19814528/6fc282e4-9d36-11e6-8061-abdd7f9dd731.png)

mauriziovitale:
@hashraf88  We can't fix this accessibility bug because the empty buttons are generated by the material design date picker. 
@magemello can we close this issue ?

magemello:
@mauriziovitale , let's try with that:

.mddtp-button::after{content: 'date button'}

PR #1018, opened 2016-11-07
Fix date picker accessibility

mauriziovitale:
#946


ng2-components/ng2-activiti-analytics/src/components/widgets/date-range/date-range.widget.ts
```diff
@@ -99,6 +100,36 @@ export class DateRangeWidget extends WidgetComponent {
         });
     }
 
+    private fixDatePickerAccessibility() {
```
magemello: better name.....AddAccessibilityLabelToDatePicker

Issue #1054, opened 2016-11-10
Date does not display 


hashraf88:
Completed tasks date displays as {{created}} and active tasks display as 'invalid date'
Fine on Firefox and Chrome

![screen shot 2016-11-10 at 17 54 57](https://cloud.githubusercontent.com/assets/13200338/20188118/18f0d23e-a76f-11e6-9e70-0d0ef9938cd5.png)

PR #1057, opened 2016-11-10
Fix date conversion

mauriziovitale:
#1054

### DIFF
--- a/ng2-components/ng2-activiti-processlist/src/services/activiti-process.service.ts
+++ b/ng2-components/ng2-activiti-processlist/src/services/activiti-process.service.ts
@@ -23,6 +23,8 @@ import { Observable }     from 'rxjs/Observable';
 import 'rxjs/add/operator/map';
 import 'rxjs/add/operator/catch';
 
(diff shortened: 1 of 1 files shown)


ng2-components/ng2-activiti-processlist/src/services/activiti-process.service.ts
```diff
@@ -168,7 +170,7 @@ export class ActivitiProcessService {
         return Observable.fromPromise(this.authService.getAlfrescoApi().activiti.taskApi.listTasks(taskOpts))
             .map(this.extractData)
             .map(tasks => tasks.map((task: any) => {
-                task.created = new Date(task.created);
+                task.created =  moment(task.created, 'YYYY-MM-DD').format();
```
DenisVuyka: Can you please create a unit test to ensure we have this problem covered.

magemello: @DenisVuyka let's add it as part of #849

PR #1074, opened 2016-11-11
#828 remove conflicting and obsolete i18n blocks

### LINKED ISSUE #828: 'Mario' appears instead of 'sign in'
Only appears after initially starting the dmeo shell
1. Npm install in demo shell folder

screen shot 2016-09-28 at 10 22 18

DenisVuyka:
just removes conflicting blocks from the demo shell. the proper i18n customisations will be introduced later.
fixes #828

### DIFF
--- a/demo-shell-ng2/i18n/en.json
+++ b/demo-shell-ng2/i18n/en.json
@@ -20,48 +20,5 @@
                 "CUSTOM": "Custom action"
             }
         }
-    },
-
-    "DATATABLE": {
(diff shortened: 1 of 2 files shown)


magemello:
it fix the button but doesn't solve the problem of the loading order/time of the #828 , so let's keep this issue open

DenisVuyka:
@magemello I've created a separate issue to address this functionality: #1076

Issue #1066, opened 2016-11-11
start process dialog: cancel does not clear form attached to start event 


hashraf88:
1. Start a process (ensure form is attached to start event)
2. Enter into values into any fields
3. Click cancel

**Expected results**
Form is cleared of any previous inputs

**Actual results**
 Form is populated with those in step 3

PR #1110, opened 2016-11-16
Reset form when cancel button clicked

wabson:
- Remove use of any in component and add types
- Small refactoring of method name
- Add many more tests

Refs #1066

### DIFF
--- a/ng2-components/ng2-activiti-processlist/src/components/activiti-start-process.component.spec.ts
+++ b/ng2-components/ng2-activiti-processlist/src/components/activiti-start-process.component.spec.ts
@@ -107,7 +107,7 @@ describe('ActivitiStartProcessButton', () => {
             fixture.whenStable().then(() => {
                 let errorEl: DebugElement = debugElement.query(By.css('.error-message'));
                 expect(errorEl).not.toBeNull('Expected error message to be present');
-                expect(errorEl.nativeElement.innerText).toBe('START_PROCESS.ERROR.LOAD_PROCESS_DEFS');
+                expect(errorEl.nativeElement.innerText.trim()).toBe('START_PROCESS.ERROR.LOAD_PROCESS_DEFS');
             });
         }));
 
@@ -204,7 +204,7 @@ describe('ActivitiStartProcessButton', () => {
                 fixture.detectChanges();
                 let errorEl: DebugElement = debugElement.query(By.css('.error-message'));
                 expect(errorEl).not.toBeNull();
-                expect(errorEl.nativeElement.innerText).toBe('START_PROCESS.ERROR.START');
+                expect(errorEl.nativeElement.innerText.trim()).toBe('START_PROCESS.ERROR.START');
             });
         }));
 
--- a/ng2-components/ng2-activiti-processlist/src/components/activiti-start-process.component.spec.ts
+++ b/ng2-components/ng2-activiti-processlist/src/components/activiti-start-process.component.spec.ts
@@ -17,12 +17,13 @@
 
 import { ComponentFixture, TestBed, async } from '@angular/core/testing';
 import { By } from '@angular/platform-browser';
-import { DebugElement } from '@angular/core';
+import { DebugElement, SimpleChange } from '@angular/core';
 import { Observable } from 'rxjs/Rx';
-import { ActivitiFormModule } from 'ng2-activiti-form';
 import { AlfrescoTranslationService, CoreModule } from 'ng2-alfresco-core';
-import { FormService } from 'ng2-activiti-form';
+import { ActivitiFormModule, FormService } from 'ng2-activiti-form';
+
 import { TranslationMock } from './../assets/translation.service.mock';
+import { newProcess, fakeProcessDefs, fakeProcessDefWithForm, taskFormMock } from './../assets/activiti-start-process.component.mock';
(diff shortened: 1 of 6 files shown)


ng2-components/ng2-activiti-processlist/src/components/activiti-start-process.component.spec.ts
```diff
@@ -80,79 +72,217 @@ describe('ActivitiStartProcessButton', () => {
...
-    it('should display the correct process def details', (done) => {
-        fixture.detectChanges();
-        fixture.whenStable().then(() => {
-            let optionEl: HTMLOptionElement = debugElement.queryAll(By.css('select option'))[1].nativeElement;
-            expect(optionEl.value).toBe('my:process1');
-            expect(optionEl.textContent.trim()).toBe('My Process 1');
-            done();
+        it('should call service to fetch process definitions', () => {
+            fixture.detectChanges();
+            expect(getDefinitionsSpy).toHaveBeenCalled();
```
mauriziovitale: Instead of checking if the method is called, could you:
1 - add an onSuccess event inside the load method
2 - subscribe to it
3 - check if the HTML is correct

ng2-components/ng2-activiti-processlist/src/components/activiti-start-process.component.spec.ts
```diff
@@ -80,79 +72,217 @@ describe('ActivitiStartProcessButton', () => {
...
+            component.name = 'My new process';
+            component.showDialog();
+            fixture.detectChanges();
         });
+
+        it('should call service to start process if required fields provided', async(() => {
+            component.onProcessDefChange('my:process1');
+            component.startProcess();
+            fixture.whenStable().then(() => {
+                expect(startProcessSpy).toHaveBeenCalled();
```
mauriziovitale: Instead of checking if the method is called, could you subscribe to the existing start eventEmitter ?

ng2-components/ng2-activiti-processlist/src/components/activiti-start-process.component.spec.ts
```diff
@@ -80,79 +72,217 @@ describe('ActivitiStartProcessButton', () => {
...
+            });
+        }));
+
+        it('should avoid calling service to start process if required fields NOT provided', async(() => {
+            component.name = '';
+            component.startProcess();
+            fixture.whenStable().then(() => {
+                expect(startProcessSpy).not.toHaveBeenCalled();
```
mauriziovitale: Could be better check if the field is required when the value is empty and the start button is disabled

ng2-components/ng2-activiti-processlist/src/components/activiti-start-process.component.spec.ts
```diff
@@ -80,79 +72,217 @@ describe('ActivitiStartProcessButton', () => {
...
+            });
+        }));
+
+        it('should call service to start process with the correct parameters', async(() => {
```
mauriziovitale: Could you check the HTML as you did for the previous test ?
let selectElement = debugElement.query(By.css('select'));
expect(selectElement.children.length).toBe(3);

ng2-components/ng2-activiti-processlist/src/components/activiti-start-process.component.spec.ts
```diff
@@ -80,79 +72,217 @@ describe('ActivitiStartProcessButton', () => {
...
+                fixture.whenStable();
+            });
+
+            it('should initialize start form', () => {
+                expect(component.startForm).toBeDefined();
+                expect(component.startForm).not.toBeNull();
+            });
+
+            it('should load start form from service', () => {
+                expect(getStartFormDefinitionSpy).toHaveBeenCalled();
```
mauriziovitale: Could you check the HTML ?

ng2-components/ng2-activiti-processlist/src/components/activiti-start-process.component.spec.ts
wabson: I have left this test in after verifying that the HTML is also checked (check correct number/content of `option` elements in later tests. There is no need to add an additional output to detect when the results have been returned for the purposes of the test since `whenStable()` ensures that the test waits until all async activity has completed before making assertions on the HTML.

ng2-components/ng2-activiti-processlist/src/components/activiti-start-process.component.spec.ts
wabson: There is another test to ensure that the `start` output emits a value, but I still think it is important to ensure the service is called

ng2-components/ng2-activiti-processlist/src/components/activiti-start-process.component.spec.ts
wabson: Added more tests to check the state of the start button (disabled or not) in 3ef10db9

ng2-components/ng2-activiti-processlist/src/components/activiti-start-process.component.spec.ts
wabson: There is no change in the HTML on starting the process, but there is another test to check that the dialog closes OK and that a value is emitted from the output

Issue #666, opened 2016-09-02
Search Folder


eromano:
We could add also the folder to the search and in case of click open it in the document list

eromano:
```javascript
let opts = {
            include: ['path'],
            rootNodeId: nodeId,
            nodeType: 'cm:content'
        };
```
If you remove  ``` nodeType: 'cm:content' ```  from the option you will search among folder and files.

hashraf88:
Issues found are reported in the below:
#1240 
#1241

PR #1209, opened 2016-12-07
Allow navigation to folders from search results

wabson:
- Uses router to pass ID of the folder
     - Modified document list component to accept folder ID without path
     - Current limitations
     - Breadcrumb cannot currently be shown when navigating via folder id
     - Clicking between folders does not update the current route

Refs #666


magemello:
needs to be rebased

PR #1419, opened 2017-01-09
update material design lite theme

### LINKED ISSUE #1111: Upgrade custom material-design-lite theme
Our custom orange-blue theme does not support the latest features of MDL (e.g. [chips](https://getmdl.io/components/index.html#chips-section)) and may have bugs which are fixed in more recent versions. We should consider upgrading.

DenisVuyka:
Update MDL theme CSS
fixes #1111


wabson:
Seems after trying this out that the demo-shell branding at the top is now orange rather than blue? I seem to recall I had the same thing last time I tried.

screen shot 2017-01-09 at 18 09 45

Issue #1387, opened 2017-01-04
Images not displayed in document list component demo


hashraf88:
When starting the document list component (outside of demo shell) the images and icons are not displayed correctly

<img width="1381" alt="screen shot 2017-01-04 at 16 19 18" src="https://cloud.githubusercontent.com/assets/13200338/21649686/78f7f5c6-d29a-11e6-87ce-b9b7e1103b98.png">

magemello:
This issue is relative only to systemjs, and is present as well in the userinfo and login component.

hashraf88:
On just testing the document list I can still see the same issue as originally reported

PR #1421, opened 2017-01-09
#1387 fix img for systemjs and refactoring apiservice

magemello:
```
[ ] The commit message follows our [guidelines](https://github.com/Alfresco/alfresco-ng2-components/wiki/Commit-format)
[ ] Tests for the changes have been added (for bug fixes / features)
[ ] Docs have been added / updated (for bug fixes / features)
```

```
[ x] Bugfix
[ ] Feature
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] Documentation
[ ] Other... Please describe:
```

**What is the current behavior?** (You can also link to an open issue here)

#1387 , I have also refactored the authService and apiService... the injection pattern was used in wrong way.

**What is the new behavior?**

```
[ ] Yes
[ ] No
```

**Other information**:

### DIFF
--- a/ng2-components/ng2-alfresco-core/src/services/AlfrescoAuthentication.service.ts
+++ b/ng2-components/ng2-alfresco-core/src/services/AlfrescoAuthentication.service.ts
@@ -20,16 +20,14 @@ import { LogService } from './log.service';
 import { SettingsService } from './settings.service';
 import { StorageService } from './storage.service';
 import { AlfrescoApiService } from './alfresco-api.service';
-import { AuthService } from './auth.service';
 
 /** @deprecated AlfrescoAuthenticationService is deprecated. Use AuthService instead */
 @Injectable()
-export class AlfrescoAuthenticationService extends AuthService {
+export class AlfrescoAuthenticationService {
     constructor(settingsService: SettingsService,
                 apiService: AlfrescoApiService,
                 storage: StorageService,
                 logService: LogService) {
-        super(settingsService, apiService, storage, logService);
-        logService.warn('Warning: AlfrescoAuthenticationService is deprecated. Use AuthService instead.');
+        logService.error('ERROR: AlfrescoAuthenticationService is deprecated. Use AuthService instead.');
     }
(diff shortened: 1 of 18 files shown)


ng2-components/ng2-alfresco-core/src/services/AlfrescoAuthentication.service.ts
```diff
@@ -20,16 +20,14 @@ import { LogService } from './log.service';
 import { SettingsService } from './settings.service';
 import { StorageService } from './storage.service';
 import { AlfrescoApiService } from './alfresco-api.service';
-import { AuthService } from './auth.service';
 
 /** @deprecated AlfrescoAuthenticationService is deprecated. Use AuthService instead */
 @Injectable()
-export class AlfrescoAuthenticationService extends AuthService {
+export class AlfrescoAuthenticationService {
```
DenisVuyka: The idea was that both old and new service still function, and old one is raising a working. If we wrap AuthService into the AlfrescoAuthenticationService (and emulate apis) then both services will work the same.

magemello: If we keep it in this way it's going to be easy have double instance of the same class, with inspected behaviour, one for AlfrescoAuthenticationService and another for AuthService.

DenisVuyka: Yes, having "extends" is wrong, I was thinking over injecting AuthService into AlfrescoAuthenticationService, i.e.

```ts
@Injectable()
export class AlfrescoAuthenticationService {
  constructor(private authService: AuthService) {}

  login(username: string, password: string) {
     return this.authService.login(username, password);
  }
}
```

magemello: I think you will still get error at build time if you were using the old type...

Issue #1332, opened 2016-12-19
Process list always shows the running one after a new process is started


VitoAlbano:
**Type of issue:**  (check with "[x]")
```
- [ ] New feature request
- [x] Bug  
- [ ] Support request
```
**Current behavior:**
After start a new process which will be completed afterward, the process list shows the wrong tab 

**Expected behavior:**
After start a new process which will be completed afterward, the process list should show the just completed process.

**Steps to reproduce the issue:**

**Component name and version:** 

**Browser and version:** 

 
**Node version (for build issues):**  

**New feature request:**

jdosti:
After starting a new process, the "Process Details" table doesn't show the details of the process that was just created.

<img width="508" alt="screen shot 2017-01-16 at 15 45 55" src="https://cloud.githubusercontent.com/assets/24432311/21989473/e73913e6-dc02-11e6-8b67-ad883052f05c.png">

The process details are only shown on the "Running", "All" and "Completed" (If the process is completed) sections.

<img width="1388" alt="screen shot 2017-01-16 at 15 52 46" src="https://cloud.githubusercontent.com/assets/24432311/21989708/e37d31be-dc03-11e6-8990-063560af32c5.png">

Issue #1307, opened 2016-12-16
After creating a new process first item in process list is shown not newly created process


hashraf88:
Create a new process

**Expected results**
Process details of process just created is displayed 

**Actual results**
First item in process list is displayed

[process list.mp4.zip](https://github.com/Alfresco/alfresco-ng2-components/files/657225/process.list.mp4.zip)

hashraf88:
When creating a process in 'running' and 'all' process filter newly created process details are displayed but not when starting a process within the 'completed' filter.

hashraf88:
No longer applicatable due to #1470

PR #1470, opened 2017-01-13
Hide the process list when starting and after started new process

wabson:
- Ensures that the newly-started process is *always* shown
- No need to reload the list any more as not shown
- Filter component shows no selected filter when starting process
- User can exit Start Process by clicking a filter again

Refs #1332, #1307


demo-shell-ng2/app/components/activiti/activiti-demo.component.ts
```diff
@@ -190,13 +191,16 @@ export class ActivitiDemoComponent implements AfterViewInit {
     }
 
     navigateStartProcess() {
+        this.processFilter = null;
```
magemello: I see this 2 lines be repeated twice, can we put it in a private method that explains what they do?

Issue #1357, opened 2016-12-26
DocumentList documentation needs details on object binding


DenisVuyka:
**Type of issue:**  (check with "[x]")
```
- [x] New feature request
- [ ] Bug  
- [ ] Support request
```
**Current behavior:**

It is not clear for the moment what are the existing properties/objects custom columns can bind to.

**Expected behavior:**

Provide better documentation and examples on underlying entities and custom columns binding

**Steps to reproduce the issue:**

**Component name and version:** 

**Browser and version:** 

 
**Node version (for build issues):**  

**New feature request:**

Issue #400, opened 2016-07-11
Support row template for datatable


magemello:

davidcognite:
This should be expanded to include more flexibility in the content column. Aikau (and indeed YUI) support the concept of pluggable cell renderers, which adds a huge amount of flexibility, but an ng-content option using transclusion would be useful too.

DenisVuyka:
@davidcognite It is much easier taking enterprise-grade datatable component and wrap it with the angular application than implementing it from the scratch. Complete datatable implementation will require huge amount of resources (human and financial) with little to no output. We have paused on raw templates because there are too many moving parts there - sorting, column reordering, cell types, context menus, etc. Technically it is much cheaper for end-developers implementing own table rather than customizing ours.

magemello:
@DenysVuyka speaking with @davidcognite the two features that he would like to see in the datatable and then reflected in the document list are:
1. Ability to define custom type for a column
2. Ability to define custom template for a column

This 2 functionality in general make sense also for me.

@davidcognite next time to avoid misunderstandings please provide more specific use cases. Most of us don't know enough Aikau to understand what is a pluggable cell render.

DenisVuyka:
We can try implementing support for cell templates, something like this:

```html
<document-list>
    <columns>
        <column key="name">
            <column-template>
                <template>
                    <h1>hello world</ht>
                    <div>{{node.id}}</div>
                </template>
            </column-template>
        </columns>
    </columns>
</document-list>
```

magemello:
@DenisVuyka, is it possible make it less verbose? entry.data.getValue(entry.row, entry.col)} removing entry.row and entry.col...also can we make available all properties of the node inside the cell?

example:
```html
<content-column
        title="{{'DOCUMENT_LIST.COLUMNS.DISPLAY_NAME' | translate}}"
        sortable="true"
        class="full-width ellipsis-cell">
    <template let-entry="$implicit">
        <span>Id: {{entry.data.getNode().id)}}</span>
        <span>Name: {{entry.data.getNode().id)}}</span>
    </template>
</content-column>
```

DenisVuyka:
That may require huge refactoring of DataTable and DataAdapter.

magemello:
To consider the task done we need at least be able to access all the node information from the cell..

DenisVuyka:
DataRow holds the Node and provides access to all node properties:

```ts
export interface DataRow {
    isSelected: boolean;
[...]

PR #1549, opened 2017-01-26
documentation updates

DenisVuyka:
refs #1357
refs #400

### DIFF
--- a/ng2-components/ng2-alfresco-documentlist/README.md
+++ b/ng2-components/ng2-alfresco-documentlist/README.md
@@ -110,8 +110,8 @@ Follow the 3 steps below:
 Usage example of this component :
 
 **main.ts**
-```ts
 
+```ts
 import { NgModule, Component, ViewChild } from '@angular/core';
 import { BrowserModule } from '@angular/platform-browser';
 import { platformBrowserDynamic } from '@angular/platform-browser-dynamic';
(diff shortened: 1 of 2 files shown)


ng2-components/ng2-alfresco-documentlist/README.md
```diff
@@ -437,18 +452,53 @@ For a full list of available `format` values please refer to [DatePipe](https://
...
+    <!-- template body -->
+</template>
+```
+
+The format of naming is `let-VARIABLE_NAME="$implicit"` where `VARIABLE_NAME` is the name of the variable you want binding template data context to.
+
+Getting a cell value from the underlying DataTableAdapter:
+
+```ts
+context.data.getValue(entry.row, entry.col);
```
magemello: which one is correct?
- context.data.getValue(entry.row, entry.col);
- entry.data.getValue(entry.row, entry.col);

ng2-components/ng2-alfresco-documentlist/README.md
```diff
@@ -437,18 +452,53 @@ For a full list of available `format` values please refer to [DatePipe](https://
...
+context.row.getValue('createdByUser.displayName')
+```
+
+_You may want using **row** api to get raw value access. 
+Use **data** api to get values with post-processing, like datetime/icon conversion._
+
+Final example:
+
+```html
+<content-column title="Name" key="name" sortable="true" class="full-width ellipsis-cell">
```
magemello: Can I use something like that? If so can you add it to the examples?

```html
<content-column title="Name" key="name" sortable="true" class="full-width ellipsis-cell">
    <template let-entry="$implicit">
        <span>Hi! {{entry.row.getValue('createdByUser.displayName')}}</span>
        <span>Hi! {{entry.row.getValue('name')}}</span>
    </template>
</content-column>
```

ng2-components/ng2-alfresco-documentlist/README.md
DenisVuyka: First, as the step above it demos naming property as `context`

ng2-components/ng2-alfresco-documentlist/README.md
DenisVuyka: Yes, you can.

Issue #1581, opened 2017-01-31
Actions placement


kodermax:
**Type of issue:**  (check with "[x]")
```
- [X] New feature request
- [ ] Bug  
- [ ] Support request
- [ ] Documentation
```
**Current behavior:**

Now actions is located at the end of row

**Browser and version:** 
all
 
**New feature request:**
I'd like too see this button at the beginning of the row.
I think it be more comfortable

![image](https://cloud.githubusercontent.com/assets/590985/22467508/05001c7e-e7d6-11e6-8dab-9363c5bbcd9d.png)

magemello:
In this case, I think we can implement an option on the documentlist to decide if to display it on right or on the left.

PR #1590, opened 2017-02-03
ability to change position of the dropdown menu (left|right)

DenisVuyka:
refs #1581

### DIFF
--- a/ng2-components/ng2-alfresco-datatable/src/components/datatable/datatable.component.html
+++ b/ng2-components/ng2-alfresco-datatable/src/components/datatable/datatable.component.html
@@ -3,6 +3,10 @@
     class="mdl-data-table mdl-js-data-table full-width mdl-data-table-fix-firefox">
     <thead>
     <tr>
+        <!-- Actions (left) -->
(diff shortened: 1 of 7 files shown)


magemello:
Can we have a couple of unit tests that checks the action be the first column and last if is left or right?

ng2-components/ng2-alfresco-datatable/src/components/datatable/datatable.component.html
```diff
@@ -23,8 +27,8 @@
             <span *ngIf="col.srTitle" class="sr-only">{{col.srTitle}}</span>
             <span *ngIf="col.title">{{col.title}}</span>
         </th>
-        <!-- Actions -->
-        <th *ngIf="actions">
+        <!-- Actions (right) -->
```
magemello: Can we have a couple of unit tests that checks the action be the first column and last if is left or right?

DenisVuyka: Added unit tests

Issue #1715, opened 2017-03-10
Demoshell test failing, same for other applications


Salaboy:
**Type of issue:**  (check with "[x]")
```
- [ ] New feature request
- [x ] Bug  
- [ ] Support request
- [ ] Documentation
```
**Current behaviour:**

When I run npm test in demo-shell I'm getting:

> Alfresco-Angular2-Demo@1.2.0 test /Users/msalatino/Projects/alfresco/alfresco-ng2-components/demo-shell-ng2
> karma start

START:

webpack: Compiled successfully.
webpack: Compiling...

webpack: Compiled successfully.
10 03 2017 16:50:38.977:WARN [karma]: Port 9876 in use
10 03 2017 16:50:38.980:INFO [karma]: Karma v1.4.1 server started at http://0.0.0.0:9877/
10 03 2017 16:50:38.981:INFO [launcher]: Launching browser PhantomJS with unlimited concurrency
10 03 2017 16:50:38.990:INFO [launcher]: Starting browser PhantomJS
10 03 2017 16:50:40.233:INFO [PhantomJS 2.1.1 (Mac OS X 0.0.0)]: Connected on socket ell7vpbJqyKlJc-qAAAA with id 49878678
PhantomJS 2.1.1 (Mac OS X 0.0.0) ERROR
  SyntaxError: Use of reserved word 'class'
  at http://localhost:9877webpack:///~/alfresco-js-api/src/alfrescoApi.js:18:0 <- config/karma-test-shim.js:50360

Finished in 0.629 secs / 0 secs @ 16:50:40 GMT+0000 (GMT)

This is, of course, breaking all the other test in a third party ADF application. 
**Expected behavior:**

The test should run. 

**Steps to reproduce the issue:**

**Component name and version:** 

**Browser and version:** 

 
**Node version (for build issues):**  

**New feature request:**

DenisVuyka:
This is related to PhantomJS, we need migrating to Chrome launcher instead.

PR #1716, opened 2017-03-10
migrate from PhantomJS to Chrome

DenisVuyka:
refs #1715


eromano:
can you also add it to the travis matrix?

DenisVuyka:
@eromano I suggest we update Travis as a separate task, I'm afraid I can be stuck with the config file

eromano:
https://github.com/Alfresco/alfresco-ng2-components/issues/1722

PR #1836, opened 2017-04-21
[ADF-495] disable debug button by default

DenisVuyka:
- debug button is now disabled by default for the Form and Task Details
- debug mode now enabled by the demo shell explicitly


eromano:
@DenisVuyka add @mauriziovitale  suggested can you add also to the analytics component?

mauriziovitale:
@DenisVuyka the file should be this one 
https://github.com/Alfresco/alfresco-ng2-components/blob/development/ng2-components/ng2-activiti-analytics/src/components/analytics-report-parameters.component.ts#L65

mauriziovitale:
As we discuss, we will open a separate issue for the Analytics component.

PR #1846, opened 2017-04-26
[DW-46] - Process Service - Add a new method to get the application details by Id

infad-kachancheri:
Fix for Issue#DW-46: New service that returns application details for a given appId. This service should be part of ng2-activiti-processlist component.

https://issues.alfresco.com/jira/browse/DW-46

### DIFF
--- a/ng2-components/ng2-activiti-processlist/src/services/activiti-process.service.ts
+++ b/ng2-components/ng2-activiti-processlist/src/services/activiti-process.service.ts
@@ -43,6 +43,21 @@ export class ActivitiProcessService {
             .catch(err => this.handleError(err));
     }
 
+    /**
+     * Retrieve deployed apps details by id
+     * @returns {Observable<any>}
+     */
+    getApplicationDetailsById(appId?: string): Observable<any> {
+        return Observable.fromPromise(this.apiService.getInstance().activiti.appsApi.getAppDefinitions())
+            .map((response: any) => {
+                if (appId) {
+                    return response.data.find(p => p.id !== null && p.id.toString() === appId);
+                }
+                return response.data;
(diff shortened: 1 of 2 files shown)


infad-kachancheri:
removed if condition and null comparison for number.

ng2-components/ng2-activiti-processlist/src/services/activiti-process.service.ts
```diff
@@ -43,6 +43,19 @@ export class ActivitiProcessService {
             .catch(err => this.handleError(err));
     }
 
+    /**
+     * Retrieve deployed apps details by id
+     * @param appId - number - optional - The id of app
+     * @returns {Observable<any>}
+     */
+    getApplicationDetailsById(appId?: number): Observable<any> {
```
mauriziovitale: the appId should not be optional but required. Please remove the character  '?'

ng2-components/ng2-activiti-processlist/src/services/activiti-process.service.ts
```diff
@@ -43,6 +43,19 @@ export class ActivitiProcessService {
...
+    /**
+     * Retrieve deployed apps details by id
+     * @param appId - number - optional - The id of app
+     * @returns {Observable<any>}
+     */
+    getApplicationDetailsById(appId: number): Observable<any> {
+        return Observable.fromPromise(this.apiService.getInstance().activiti.appsApi.getAppDefinitions())
+            .map((response: any) => {
+                return response.data.find(p => p.id === appId);
+            })
```
eromano: Please we don't use one variable name, can you give a name that give the idea of the parameter thanks

PR #1847, opened 2017-04-26
[DW-47] - Analytics Service - Add a new method to get the report details by name

infad-kachancheri:
Fix for Issue#DW-47: Create a service that returns report details for a given report name as part of ng2-activiti-analytics component. There is no spec file for this service thats why skipped Unit Testcase.

https://issues.alfresco.com/jira/browse/DW-47

### DIFF
--- a/ng2-components/ng2-activiti-analytics/src/services/analytics.service.ts
+++ b/ng2-components/ng2-activiti-analytics/src/services/analytics.service.ts
@@ -55,6 +55,18 @@ export class AnalyticsService {
             }).catch(err => this.handleError(err));
     }
 
+    /**
+      * Retrive Report by name
+      * @param reportName - string - The name of report
+      * @returns {Observable<any>}
+      */
+     getReportByName(reportName: string): Observable<any> {
+         return Observable.fromPromise(this.apiService.getInstance().activiti.reportApi.getReportList())
+             .map((response: any) => {
+                 return response.find(report => report.name === reportName);
+             }).catch(err => this.handleError(err));
+     }
+
     private isReportValid(appId: string, report: ReportParametersModel) {
         let isValid: boolean = true;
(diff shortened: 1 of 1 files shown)


ng2-components/ng2-activiti-analytics/src/services/analytics.service.ts
```diff
@@ -55,6 +55,22 @@ export class AnalyticsService {
...
+    /**
+     * Retrive Report by name
+     * @param reportName - string - The name of report
+     * @returns {Observable<any>}
+     */
+    getReportByName(reportName: string): Observable<any> {
+        return Observable.fromPromise(this.apiService.getInstance().activiti.reportApi.getReportList())
+            .map((response: any) => {
+
+                if (reportName) {
```
mauriziovitale: We don't need this condition, the reportName is required as parameter

ng2-components/ng2-activiti-analytics/src/services/analytics.service.ts
```diff
@@ -55,6 +55,22 @@ export class AnalyticsService {
...
+                    return response.find(p => p.name === reportName);
+                }
+                return null;
```
eromano: never propagate null and provide only one exit point for method

ng2-components/ng2-activiti-analytics/src/services/analytics.service.ts
infad-kachancheri: these changes done.

ng2-components/ng2-activiti-analytics/src/services/analytics.service.ts
infad-kachancheri: these changes done.

infad-kachancheri:
re-opened

PR #1872, opened 2017-05-10
[ADF-573] ability to disable drop upload area

DenisVuyka:
- [disabled]=“expression” support for drop upload area
- extended underlying directive with disabled support
- readme updates


DenisVuyka:
@mauriziovitale added extra tests for the directive

PR #1996, opened 2017-06-20
[ADF-850] support binding [form] data directly

DenisVuyka:
Adds features to Form component as a prerequisite to implementing ADF-850.
Note: the ADF-850 is still in progress and will be based on the changes from this PR.

- ability to bind [form] data directly inside `<activiti-form>` component
- ability to parse forms with FormService
- demo of the custom form binding


ng2-components/ng2-activiti-form/src/components/activiti-form.component.ts
```diff
@@ -41,6 +41,9 @@ export class ActivitiForm implements OnInit, AfterViewChecked, OnChanges {
     static CUSTOM_OUTCOME_ID: string = '$custom';
 
     @Input()
+    form: FormModel;
```
eromano: Missing documentation in readme

demo-shell-ng2/app/components/form/form-demo.component.css
```diff
@@ -0,0 +1,3 @@
+.form-container {
```
eromano: can you add the adf-prefix?

ng2-components/ng2-activiti-form/src/services/form.service.ts
```diff
@@ -47,6 +47,23 @@ export class FormService {
                 private logService: LogService) {
     }
 
+    parseForm(json: any, data?: FormValues, readOnly: boolean = false): FormModel {
```
eromano: Can not retrun an empty form instead of null?I would like to prefer to not propagate null in the code

demo-shell-ng2/app/components/form/form-demo.component.css
DenisVuyka: this is an end application (demo shell), it does not matter whether it has prefix or not :)

ng2-components/ng2-activiti-form/src/services/form.service.ts
DenisVuyka: This will complicate entire flow, as it will be hard to tell the parsing didn't happen. Developers will need to do additional checks to see whether result value is what they expect or not, especially if binding in Angular template. I think 'null' is perfectly valid in this case. You can always do in your code `let model = parseForm(...) || new FormModel(...)`

ng2-components/ng2-activiti-form/src/components/activiti-form.component.ts
DenisVuyka: It is in the readme in the `Advanced Properties` section since the beginning, it was just missing the `@Input` decorator.

PR #2040, opened 2017-07-03
[ADF-967] evaluate "hostname" and "port" for string settings

DenisVuyka:
Provides ability for AppConfigService to evaluate "hostname" and "port" variables within string values


DenisVuyka:
@eromano updated readme

PR #2052, opened 2017-07-06
[ADF-1015] make "loading" state two-way bindable (document list)

DenisVuyka:
Add @Input decorator for the "loading" property in Document List component


DenisVuyka:
@eromano updated documentation

PR #2154, opened 2017-07-31
[ASF-1227] new adf-node-permission directive

DenisVuyka:
- New `NodePermissionDirective` (`adf-node-permission`)
- update readme with a link to directive docs
- update demo shell to demonstrate the directive in the toolbar (delete button)

### DIFF
--- a/ng2-components/ng2-alfresco-core/src/directives/node-permission.directive.ts
+++ b/ng2-components/ng2-alfresco-core/src/directives/node-permission.directive.ts
@@ -45,8 +45,13 @@ export class NodePermissionDirective implements OnChanges, AfterViewInit {
         }
     }
 
-    updateElement() {
-
+    /**
+     * Updates disabled state for the decorated elememtn
+     *
+     * @returns {boolean} True if decorated element got disabled, otherwise False
+     * @memberof NodePermissionDirective
+     */
+    updateElement(): boolean {
         let enable = true;
 
         if (this.nodes && this.nodes.length > 0) {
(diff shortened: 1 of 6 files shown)


ng2-components/ng2-alfresco-core/src/directives/node-permission.directive.ts
```diff
@@ -0,0 +1,95 @@
...
+    /**
+     * Updates disabled state for the decorated elememtn
+     *
+     * @returns {boolean} True if decorated element got disabled, otherwise False
+     * @memberof NodePermissionDirective
+     */
+    updateElement(): boolean {
+        let enable = true;
+
+        if (this.nodes && this.nodes.length > 0) {
```
mauriziovitale: Can you create a new method hasPermission(node, permission) that returns true/false
In this way, we can simplify the code like
updateElement() {
if (this.hasPermission(node, this.permission)){
     this.enableElement();
} else {
     this.disableElement();
}

DenisVuyka:
@mauriziovitale Updated the code according to the suggestion

ng2-components/ng2-alfresco-core/src/directives/node-permission.directive.ts
```diff
@@ -0,0 +1,101 @@
...
+    /**
+     * Disables decorated element
+     *
+     * @memberof NodePermissionDirective
+     */
+    disableElement() {
+        this.renderer.setAttribute(this.elementRef.nativeElement, 'disabled', 'true');
+    }
+
+    hasPermission(nodes: MinimalNodeEntity[], permission: string) {
```
chris-equis: Why not just:
```js
hasPermission(nodes: MinimalNodeEntity[], permission: string): boolean {
    return nodes.every(node => this.contentService.hasPermission(node.entry, permission));
}
```

DenisVuyka: Will update the code shortly.

PR #2183, opened 2017-08-07
breadcrumb style enhancements

DenisVuyka:
Improve breadcrumb styles


DenisVuyka:
@eromano Can you please suggest on how make colours support primary/accent/warn modes?

PR #2204, opened 2017-08-11
[ADF-1372] support for tooltips (Upload button)

DenisVuyka:
Allow setting custom tooltips via `tooltip` property


DenisVuyka:
@eromano It's a tooltip property, not sure transclusion will help here. But we could use transclusion in the future to set button text, i.e. `<adf-upload-button>Some text and icon</adf-upload-button>`

PR #2270, opened 2017-09-01
[ADF-1458] wrongly named input attribute - renamed disabled into disable

VitoAlbano:

> - [X] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

Upload button is not disabled when the attribute disabled is changed, this is due overlapping of the attributes with the standard html one mapped by angular.

Renamed disabled into disable to allow the correct attribute mapping by angular.

> - [ ] Yes
> - [ ] No

**Other information**:

### DIFF
--- a/demo-shell-ng2/app/components/files/files.component.html
+++ b/demo-shell-ng2/app/components/files/files.component.html
@@ -268,10 +268,10 @@
         </md-input-container>
     </section>
     <div *ngIf="!acceptedFilesTypeShow">
-        <alfresco-upload-button
+        <adf-upload-button
             #uploadButton
             tooltip="Custom tooltip"
-            [disabled]="!enableUpload"
+            [disable]="!enableUpload"
             data-automation-id="multiple-file-upload"
             [rootFolderId]="documentList.currentFolderId"
             [multipleFiles]="multipleFileUpload"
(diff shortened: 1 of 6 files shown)


demo-shell-ng2/app/components/files/files.component.html
```diff
@@ -268,10 +268,10 @@
         </md-input-container>
     </section>
     <div *ngIf="!acceptedFilesTypeShow">
-        <alfresco-upload-button
+        <adf-upload-button
             #uploadButton
             tooltip="Custom tooltip"
-            [disabled]="!enableUpload"
+            [disable]="!enableUpload"
```
DenisVuyka: 'disabled' is a well-known attribute name for all HTML elements. why do we need renaming it to "disable" instead?

DenisVuyka: Example: https://www.w3schools.com/tags/att_input_disabled.asp

VitoAlbano: because it's not a valid input attribute name, 'cause in this way the value in input won't change. But due to compliance of name we have found a better solution @Input('disabled').

PR #2481, opened 2017-10-17
[ADF-1745] Delete multiple nodes

pionnegru:

> - [ ] Bugfix
> - [x] Feature
> - [ ] Build related changes
> - [ ] Documentation

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/docs/node-delete.directive.md
+++ b/docs/node-delete.directive.md
@@ -1,0 +1,48 @@
+# Node Delete directive
+
+Delete multiple files and folders
+
+<!-- markdown-toc start - Don't edit this section.  npm run toc to generate it-->
+
+<!-- toc -->
+
+- [Basic Usage](#basic-usage)
+  * [Properties](#properties)
+  * [Events](#events)
+- [Details](#details)
+
+<!-- tocstop -->
+
+<!-- markdown-toc end -->
+
+## Basic Usage
+
+```html
+<adf-toolbar>
+    <button md-icon-button
+            (delete)="documentList.reload()"
+            [adf-delete]="documentList.selection">
+    </button>
+</adf-toolbar>
+
+<adf-document-list #documentList ...>
+ ...
+</adf-document-list>
+```
+
+### Properties
+
+| Name              | Type                | Default | Description                 |
+| ----------------- | ------------------- | ------- | --------------------------- |
+| adf-delete        | MinimalNodeEntity[] | []      | Nodes to delete             |
+| permanent         | boolean             | false   | Permanent delete, no undo action |
+
+### Events
+
+| Name                      | Description                                  |
+| ------------------------- | -------------------------------------------- |
+| delete                    | emitted when delete/undo process is done |
+
+## Details
+
+See **Demo Shell**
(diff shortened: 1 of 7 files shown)


ng2-components/ng2-alfresco-core/src/directives/node-delete.directive.ts
```diff
@@ -0,0 +1,227 @@
...
+
+    constructor(
+        private notification: NotificationService,
+        private alfrescoApiService: AlfrescoApiService,
+        private translation: TranslationService
+    ) {
+        this.nodesApi = this.alfrescoApiService.getInstance().nodes;
+    }
+
+    process(selection: any, isRestore = false) {
```
eromano: - avoid to use boolean in method firm . Two behaviour in the same class are bad pratice 
https://martinfowler.com/bliki/FlagArgument.html.
- Is possible to use a type instead of any?

ng2-components/ng2-alfresco-core/src/directives/node-delete.directive.ts
```diff
@@ -0,0 +1,227 @@
...
+
+    private getDeletedNodesBatch(selection) {
+        return selection.map((node) => this.performAction('delete', node));
+    }
+
+    private getRestoredNodesBatch(selection) {
+        return selection.map((node) => this.performAction('restore', node));
+    }
+
+    private performAction(action, node) {
```
eromano: can we add the return type in all class?

ng2-components/ng2-alfresco-core/src/directives/node-delete.directive.ts
```diff
@@ -0,0 +1,227 @@
...
+
+                this.notification.openSnackMessageAction(message, action)
+                    .onAction()
+                    .subscribe(() => {
+                        this.process(status.success, true);
+                    });
+            });
+    }
+
+    private getMessage(status, isRestore): Observable<string> {
```
eromano: avoid flag arguments

docs/node-delete.directive.md
```diff
@@ -0,0 +1,48 @@
...
+| Name              | Type                | Default | Description                 |
+| ----------------- | ------------------- | ------- | --------------------------- |
+| adf-delete        | MinimalNodeEntity[] | []      | Nodes to delete             |
+| permanent         | boolean             | false   | Permanent delete, no undo action |
+
+### Events
+
+| Name                      | Description                                  |
+| ------------------------- | -------------------------------------------- |
+| delete                    | emitted when delete/undo process is done |
```
eromano: I don't think should be in the same component the concern to delete and undo it looks two separate directive to me

pionnegru: You are right, but this could lead to satellite directives which will do nothing on their own Maybe we would need a service or Command Pattern undo/redo last action.

eromano: I would prefer to separate it also because I immagine people could decide to have two separate buttons for this action

eromano: A command pattern could make sense I agree

pionnegru:
@eromano , updated as discussed

PR #2603, opened 2017-11-03
[ADF-1586] Added boilerplate files to doc folder

therealandeeee:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [x] Documentation

Added "boilerplate" files that can be used as templates for starting new component and service docs. The indexing script doesn't list them and Compodoc will ignore them, so it seems like a good idea to put them in the docs folder with the other files. I'll add some separate instructions for writing docs in a wiki page and then link to that from the boilerplate page.

> - [ ] Yes
> - [x] No

### DIFF
--- a/docs/boilerplate.component.md
+++ b/docs/boilerplate.component.md
@@ -1,0 +1,63 @@
+# Boilerplate component
+
+Shows how to write a Markdown file for a component.
+
+![Screenshot goes here if necessary](docassets/images/adf-toolbar-01.png)
+
+<!-- Most doc files don't need a table of contents. Delete this part unless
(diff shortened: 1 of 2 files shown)


eromano:
could be better have it in the wiki?

therealandeeee:
I was thinking that these files should be in the docs folder because you are supposed to copy/paste them when you are creating a new doc file. But if you think they will be confusing then I'll put them somewhere else. We could have a folder for them in docassets, maybe? I think this would be more convenient than the wiki for copy/paste because you could just duplicate and rename the file if they are in a folder.

PR #2838, opened 2018-01-16
[ADF-1769] Added JSDocs and updated prop table script

therealandeeee:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [x] Documentation

Added JSDocs to sources. Also fixed two issues with the prop table tool: Markdown wasn't transferred properly from the JSDoc comments and get/set accessors weren't picked up correctly as properties.

> - [ ] Yes
> - [x] No

### DIFF
--- a/docs/accordion.component.md
+++ b/docs/accordion.component.md
@@ -29,15 +29,6 @@ export class MyComponent implements OnInit {
 }
 ```
 
-### Properties
-
-| Name | Type | Description |
-| ---- | ---- | ----------- |
-| heading | string | The header title. |
-| isSelected | boolean | Define if the accordion group is selected or not. |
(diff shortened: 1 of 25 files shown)


docs/accordion.component.md
```diff
@@ -29,15 +29,6 @@ export class MyComponent implements OnInit {
 }
 ```
```
eromano: Why this table has been removed?

therealandeeee: The properties are actually declared in Accordion Group, which has its own separate page and the table is OK there. Having the pages separate like this is useful for auto-generating the props tables (because the .ts file corresponds directly to the .md file for the same class). However, if you think this is unhelpful for the reader then I'll try to find a way around it.

PR #2868, opened 2018-01-23
[ADF-2176] [Destination picker] File Libraries are not displayed

suzanadirla:
update columns used for display to be compatible with the items to be displayed

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://issues.alfresco.com/jira/browse/ADF-2176

bug fixed

> - [ ] Yes
> - [x] No

**Other information**:


lib/content-services/document-list/components/document-list.component.ts
```diff
@@ -872,6 +868,47 @@ export class DocumentListComponent implements OnInit, OnChanges, OnDestroy, Afte
...
+    /**
+     * Updates the columns for display according to the type of nodes needed to be displayed and the custom layout defined
+     * @param nodeId The ID of the folder node to display or the reserved string alias for special sources
+     */
+    updateLayout(nodeId) {
+        if (!this.hasCustomLayout) {
+            this.setupDefaultColumns(nodeId);
+
+        } else {
+            const columns = this.columnList.columns.map(c => <DataColumn> c);
```
popovicsandras: I don't understand the point of this mapping.

lib/content-services/document-list/components/document-list.component.ts
```diff
@@ -872,6 +868,47 @@ export class DocumentListComponent implements OnInit, OnChanges, OnDestroy, Afte
...
+
+    canDisplayNameColumn(nodeId, columns: DataColumn[]): boolean {
+        if (nodeId) {
+            const nameKey = columns.filter(col => col.key === this.getNameColumnKey(nodeId));
+            return nameKey.length > 0;
+        }
+        return true;
+    }
+
+    getNameColumnKey(nodeId): string {
```
popovicsandras: I don't understand why we need to differentiate (here). In case of -mysites- and -sites- the API response contains only the title instead of name?

suzanadirla: We need to differentiate, indeed, because of the API response: in case of -sites- and -mysites- the response for getSites and getSiteMembership is of type Promise<SitePaging> and type Site does not have 'name' (please see [the index.d.ts from alfresco-js-api]( https://github.com/Alfresco/alfresco-js-api/blob/2.1.0/index.d.ts#L1673)  https://github.com/Alfresco/alfresco-js-api/blob/2.1.0/index.d.ts#L1673 )

lib/content-services/document-list/components/document-list.component.ts
suzanadirla: Because of ShareDataTableAdapter's  `setColumns` call from [here](https://github.com/Alfresco/alfresco-ng2-components/pull/2868/files/aa9fea80e94aa8c1bb9931e15fddc08c8d339e39#diff-80f7c6137233f6a2440b1e75cc0b9244R885) which requires that `columns` is of type `Array<DataColumn>`

popovicsandras: Hmm, correct me if I'm wrong but when it is transpiled, I think we get something like this: 
const columns = this.columnList.columns.map(c => c);
(since types are just exist in Typescript, and only important for compile time errors).

That is why I think iterating over an array (with Array.map) in runtime, just for bypassing the transpiler's type checking, is not correct.

Something like this, would it work?
const columns = <DataColumn[]>this.columnList.columns;

suzanadirla: I understand your concerns and that there must be a better way so as to avoid extra iterations. Although the line of code that you proposed is not working in this case, I have found [here](https://github.com/angular/angular/blob/5.1.1/packages/core/src/linker/query_list.ts#L39) that the `QueryList<DataColumnComponent>` (`columns`' type) has method `toArray` that I will use instead.

lib/content-services/document-list/components/document-list.component.ts
popovicsandras: So, the root cause of the problem is that the documentlist sometimes use the NodePaging (as I understand this was the interface originally) and sometimes the SitePaging (this was implemented later).

So instead of 1 clear interface, the documentlist uses 2 interfaces. And that is the main problem I think, the documentlist component should not know about how it is used. It should not care if it gets a Nodepaging or a SitePaging, and especially should not have a logic inside to handle both of them.

NodaPaging has a property called "name", SitePaging has the property "title". For some historical reason, they are different, but the documentlist originally wanted to access the "name" because the first interface it worked with was the NodePaging.

The best option would be to use one clear (common) interface defined by the documentlist, instead of the two, but maybe it is something more difficult to do it as a bugfix.

So for now I'd say the easisest option would be to add a getter method to the SitePaging like this:

get name() {
    return this.title
}

DenisVuyka: @popovicsandras SitePaging is an interface of the list, the type of the item is [Site](https://github.com/Alfresco/alfresco-js-api/blob/master/src/alfresco-core-rest-api/src/model/Site.js#L37) and it is automatically generated from the swagger definition. Do you suggest manually updating the file with the new property?

DenisVuyka: I would rather file a feature request to the REPO team to have consistent naming, and have the current code in ADF as a workaround. @eromano ?

popovicsandras: Hmm, sorry I didn't know it was automatically generated.
The best would be to have consistent naming, I agree. 
However ff it is not possible, we still have to find a way to somehow create a structure from SitePaging which implements the interface which is required by the document list component.

popovicsandras: In the document-list component, we could do this trick, until the REPO team fixes it. 
We have 2 different method for loading the sites (loadSites) and mysites (loadMemberSites). If we create a disposable helper function separately, which wraps the result in an object which implements the desired interface, I think we are good to go. 
Any ideas? @DenisVuyka @suzanadirla

suzanadirla: Indeed, the code would be much simplified if doing something like you mentioned above, Andras. I will make another commit with a simplified approach soon. Thanks for the feedback.

eromano:
@suzanadirla can you re add the missing test for this scenario?

PR #3037, opened 2018-03-06
[ADF-2358] splitted copy and move opening mechanism

VitoAlbano:

> - [X] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

Copy dialog checks if the current node has permission and this is not how CS works.

Copyis always be possible, move only if user has permissions

> - [ ] Yes
> - [ ] No

**Other information**:

### DIFF
--- a/lib/content-services/content-node-selector/content-node-dialog.service.ts
+++ b/lib/content-services/content-node-selector/content-node-dialog.service.ts
@@ -72,33 +72,40 @@ export class ContentNodeDialogService {
     /** @param action Name of the action (eg, "Copy" or "Move") to show in the title */
     /** @param contentEntry Item to be copied or moved */
     /** @param permission Permission for the operation */
-    openCopyMoveDialog(action: string, contentEntry: MinimalNodeEntryEntity, permission?: string): Observable<MinimalNodeEntryEntity[]> {
+    openMoveDialog(action: string, contentEntry: MinimalNodeEntryEntity, permission?: string): Observable<MinimalNodeEntryEntity[]> {
         if (this.contentService.hasPermission(contentEntry, permission)) {
-
-            const select = new Subject<MinimalNodeEntryEntity[]>();
-            select.subscribe({
-                complete: this.close.bind(this)
-            });
-
-            const title = this.getTitleTranslation(action, contentEntry.name);
-
-            const data: ContentNodeSelectorComponentData = {
-                title: title,
-                actionName: action,
-                currentFolderId: contentEntry.parentId,
-                imageResolver: this.imageResolver.bind(this),
-                rowFilter : this.rowFilter.bind(this, contentEntry.id),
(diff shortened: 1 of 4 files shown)


DenisVuyka:
Why then we don't have two dialogs? Putting a magic string like "copy" into the parameter seems like a workaround

lib/content-services/content-node-selector/content-node-dialog.service.ts
```diff
@@ -72,33 +72,40 @@ export class ContentNodeDialogService {
     /** @param action Name of the action (eg, "Copy" or "Move") to show in the title */
     /** @param contentEntry Item to be copied or moved */
     /** @param permission Permission for the operation */
-    openCopyMoveDialog(action: string, contentEntry: MinimalNodeEntryEntity, permission?: string): Observable<MinimalNodeEntryEntity[]> {
+    openMoveDialog(action: string, contentEntry: MinimalNodeEntryEntity, permission?: string): Observable<MinimalNodeEntryEntity[]> {
```
DenisVuyka: Should this be a breaking change? The public method gets renamed.

eromano: nope

eromano: the component input and output and name are our contracts

DenisVuyka: what is someone is using this api already?

DenisVuyka: But this is a service

suzanadirla:
On ACA side we do not have this problem because we have this method:
 ` private isActionAllowed(action: string, node: MinimalNodeEntryEntity, permission?: string): boolean {
        if (action === 'copy') {
            return true;
        }
        return this.contentService.hasPermission(node, permission);
    }`
Maybe it would be good to consider such an approach so that there are not too many code changes introduced. What do you think?

lib/content-services/content-node-selector/content-node-dialog.service.ts
suzanadirla: On ACA side we do not have this problem because we have this method:
 ` private isActionAllowed(action: string, node: MinimalNodeEntryEntity, permission?: string): boolean {
        if (action === 'copy') {
            return true;
        }
        return this.contentService.hasPermission(node, permission);
    }`
Maybe it would be good to consider such an approach so that there are not too many code changes introduced. What do you think?

VitoAlbano: I think it's ok, so i will close this one and create a new one adding the new enum and condition to the method :)

VitoAlbano:
This will be fixed with https://github.com/Alfresco/alfresco-ng2-components/pull/3044

PR #3050, opened 2018-03-08
[ADF-2361] added breadcrumb transform function to the breadcrumb comp…

VitoAlbano:
…onent

> - [ ] Bugfix
> - [X] Feature
> - [ ] Build related changes
> - [ ] Documentation

custom transformation funcion is something that could be only applied to the content-node-panel compoenent.

breadcrumb component has a new input parameter which allow you to apply a transformation function to the node given.

> - [ ] Yes
> - [ ] No

**Other information**:

### DIFF
--- a/docs/breadcrumb.component.md
+++ b/docs/breadcrumb.component.md
@@ -25,7 +25,7 @@ Indicates the current position within a navigation hierarchy.
 | root | `string` | `null` | (optional) Name of the root element of the breadcrumb. You can use this property to rename "Company Home" to "Personal Files" for example. You can use an i18n resource key for the property value. |
 | rootId | `string` | `null` | (optional) The id of the root element. You can use this property to set a custom element the breadcrumb should start with. |
 | target | `DocumentListComponent` |  | (optional) Document List component to operate with. The list will update when the breadcrumb is clicked. |
-| transform | (node) => any | null | Action to be performed to the chosen/folder node before building the breadcrumb UI. Can be useful in case a custom formatting is needed to the breadcrumb, so changing the node's path elements that help build the breadcrumb can be done through this function. |
+| transform | (node) => any | null | Action to be performed to the chosen/folder node before building the breadcrumb UI. Can be useful when custom formatting is needed to the breadcrumb, so changing the node's path elements that help build the breadcrumb can be done through this function. |
 
 ### Events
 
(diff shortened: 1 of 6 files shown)


docs/breadcrumb.component.md
```diff
@@ -25,6 +25,7 @@ Indicates the current position within a navigation hierarchy.
 | root | `string` | `null` | (optional) Name of the root element of the breadcrumb. You can use this property to rename "Company Home" to "Personal Files" for example. You can use an i18n resource key for the property value. |
 | rootId | `string` | `null` | (optional) The id of the root element. You can use this property to set a custom element the breadcrumb should start with. |
 | target | `DocumentListComponent` |  | (optional) Document List component to operate with. The list will update when the breadcrumb is clicked. |
+| transform | (node) => any | null | Action to be performed to the chosen/folder node before building the breadcrumb UI. Can be useful in case a custom formatting is needed to the breadcrumb, so changing the node's path elements that help build the breadcrumb can be done through this function. |
```
popovicsandras: in case *of*?

lib/content-services/breadcrumb/breadcrumb.component.spec.ts
```diff
@@ -205,4 +205,26 @@ describe('Breadcrumb', () => {
         expect(route[0].id).toBe('custom-id');
         expect(route[0].name).toBe('custom-name');
     });
+
+    it('should apply the transformation function when there is one', () => {
```
popovicsandras: The parameter of transform (node) method is not tested this way. In the other test file the parameter is not needed to be tested, but here, I think it would be good.

docs/breadcrumb.component.md
VitoAlbano: 👯‍♂️

lib/content-services/breadcrumb/breadcrumb.component.spec.ts
VitoAlbano: 🥇

docs/breadcrumb.component.md
popovicsandras: I meant instead of:
"Can be useful in case a custom formatting..."
should be this:
"Can be useful in case of a custom formatting..."

Since I speak the Queen's english, you can trust me. :trollface: 
By the way, I think the version with "of" is the correct one.

PR #3063, opened 2018-03-12
[ADF-2477] message bus in log service

eromano:

> - [ ] Bugfix
> - [x] Feature
> - [ ] Build related changes
> - [ ] Documentation

to allow third pat integrations I added this message bus. Meanwhile I was trying to write an article to integrate our log service with other external services I discovered that is really difficult to integrate. My idea is provide a bus message where you can subscribe and receive all the messages. Once you subscribe to this message bus you can from there call an external service

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/core/services/log.service.ts
+++ b/lib/core/services/log.service.ts
@@ -18,13 +18,22 @@
 import { Injectable } from '@angular/core';
 import { AppConfigService } from '../app-config/app-config.service';
 import { logLevels, LogLevelsEnum } from '../models/log-levels.model';
+import { Observable } from 'rxjs/Observable';
+import { BehaviorSubject } from 'rxjs/BehaviorSubject';
 
 @Injectable()
 export class LogService {
 
     currentLogLevel: LogLevelsEnum = LogLevelsEnum.TRACE;
 
+    private onMessageSubject: BehaviorSubject<any>;
+    onMessage: Observable<any>;
+
     constructor(appConfig: AppConfigService) {
+
+        this.onMessageSubject = new BehaviorSubject('');
+        this.onMessage = this.onMessageSubject.asObservable();
+
         if (appConfig) {
             let configLevel: string = appConfig.get<string>('logLevel');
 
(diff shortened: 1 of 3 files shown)


lib/core/services/log.service.ts
```diff
@@ -15,16 +15,27 @@
...
 export class LogService {
 
     currentLogLevel: LogLevelsEnum = LogLevelsEnum.TRACE;
 
+    private onMessageSubject: BehaviorSubject<any>;
+    onMessage: Observable<any>;
+
     constructor(appConfig: AppConfigService) {
+
+        this.onMessageSubject = new BehaviorSubject('');
```
DenisVuyka: If you use Subject<T> you will be able both raising and subscribing with the same object, without 2 different properties.

lib/core/services/log.service.ts
```diff
@@ -34,82 +45,79 @@ export class LogService {
...
     }
 
-    get error(): (message?: any, ...optionalParams: any[]) => any {
+    error(message?: any, ...optionalParams: any[]) {
         if (this.currentLogLevel >= LogLevelsEnum.ERROR) {
-            return console.error.bind(console);
+
+            this.messageBus(message, 'ERROR');
+
+            console.error(message, ...optionalParams);
```
DenisVuyka: The whole point and complexity of the original code was to preserve stack trace in the debugger, and point debugger to the code that raise the error, rather than log service itself.  The change pretty much kills the purpose.

lib/core/services/log.service.ts
```diff
@@ -15,16 +15,27 @@
...
+import { Observable } from 'rxjs/Observable';
+import { BehaviorSubject } from 'rxjs/BehaviorSubject';
 
 @Injectable()
 export class LogService {
 
     currentLogLevel: LogLevelsEnum = LogLevelsEnum.TRACE;
 
+    private onMessageSubject: BehaviorSubject<any>;
+    onMessage: Observable<any>;
```
DenisVuyka: Note that we risk introducing memory leaks if we do not unsubscribe from such events. This service might prevent garbage collection for subscribers.

lib/core/services/log.service.ts
eromano: It was calling the log as well before binding it . Anyway I didn't find a better solution to put between the bind and the call the message bus

eromano:
I added the subject good idea

lib/core/services/log.service.ts
DenisVuyka: That is fine, was just explaining the previous implementation and why it’s ugly :)

PR #3087, opened 2018-03-16
[ADF-2444] supportedPageSize is used only via input parameter

VitoAlbano:

> - [X] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

supporte page size was taken from the target and always the default value was taken.

The supported page size given in input to the pagination component is the only and trusted one.

> - [ ] Yes
> - [ ] No

**Other information**:

### DIFF
--- a/demo-shell/src/app/components/files/files.component.ts
+++ b/demo-shell/src/app/components/files/files.component.ts
@@ -188,7 +188,6 @@ export class FilesComponent implements OnInit, OnChanges, OnDestroy {
     }
 
     ngOnInit() {
-
         if (!this.pagination) {
             this.pagination = <Pagination>{
                 maxItems: this.preference.paginationSize,
--- a/demo-shell/src/app/components/files/files.component.ts
+++ b/demo-shell/src/app/components/files/files.component.ts
@@ -33,7 +33,7 @@ import {
     AuthenticationService, ContentService, TranslationService,
     FileUploadEvent, FolderCreatedEvent, LogService, NotificationService,
     UploadService, DataColumn, DataRow, UserPreferencesService,
-    PaginationComponent, FormValues, DisplayMode
+    PaginationComponent, FormValues, DisplayMode, UserPreferenceValues
 } from '@alfresco/adf-core';
 
 import { DocumentListComponent, PermissionStyleModel } from '@alfresco/adf-content-services';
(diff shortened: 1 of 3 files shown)


demo-shell/src/app/components/files/files.component.ts
```diff
@@ -159,6 +159,10 @@ export class FilesComponent implements OnInit, OnChanges, OnDestroy {
                 private preference: UserPreferencesService,
                 @Optional() private route: ActivatedRoute,
                 public authenticationService: AuthenticationService) {
+        this.preference.select(UserPreferenceValues.SupportedPageSizes)
+            .subscribe((pages) => {
+                this.supportedPages = pages;
```
suzanadirla: @Vito, wouldn’t be better to change the getDefaultPageSizes from the UserPreferenceValues service to return what it did before: the value from the app.config.json.  Something like “this.onChangeSubject.getValue()[UserPreferenceValues.SupportedPageSizes];”? This way there would be no need to make changes to set the supported page sizes on each of the document-lists.

VitoAlbano: What we have done before is : change the default object into the user preference that doesn't sound the best solution. That's why the document list was working fine . Also i think that the container component should  be in charge of handling the "talking" between the pagination component and the document list, otherwise we will end up to put again the pagination handling into the documentlist.

suzanadirla: I suppose there is no ‘right’ way to fix this, it all depends on what it is that you want to achieve.. But, from a component user perspective, I thought that pagination was supposed to be integrated with the document-list. @DenisVuyka, I think you might be more appropriate than me to review these changes, pls. thank you.

DenisVuyka: I guess we will be setting the pagination values at ACA side when declaring the pagination component.

VitoAlbano: Yep that's how it supposed to work. The default object inside the user preference is not something intended to be changed ( maybe we could 'const' it ) . What i expect is that the 'orchestrator' between the two components handle this, also not so far in time we removed the pagination from document list 'cause we wanted it more generic, and it lacks its purpose if we start to put everything in again.

eromano:
ca n you add the deprecated tag on the interface?

PR #3094, opened 2018-03-19
[ADF-2128] facet container component

DenisVuyka:

> - [ ] Bugfix
> - [x] Feature
> - [ ] Build related changes
> - [ ] Documentation

- Facet Container implementation
- Demo configuration for faceted search (`app.config.json`)
- Simple integration with the demo shell
- Several demo facet components (real implementations are pending as separate tasks)
- Chip list component for the selected facets
- Unit tests for the Query Builder

Notes:

- no tests for demo facets (real components to be introduced separately)
- docs will be written in the scope of separate task (ADF-2525)

search-04

search-05

> - [ ] Yes
> - [x] No

**Other information**:


DenisVuyka:
@eromano unit tests can be found at:
- https://github.com/Alfresco/alfresco-ng2-components/pull/3094/files#diff-7f4055ced1f6a90a349777d14e65a972
- https://github.com/Alfresco/alfresco-ng2-components/pull/3094/files#diff-e17d0f29e54edce07e7094237c44b4a0

the are also tests that are missing after rebase and force push of the repository, not related to the current PR.

demo-shell/src/app.config.json
```diff
@@ -51,6 +51,118 @@
             "label": "Simplified Chinese"
         }
     ],
+    "search": {
```
eromano: Is missing the definition of the search field in the app.config schema

demo-shell/src/app/components/search/search-result.component.html
```diff
@@ -1,19 +1,24 @@
-<adf-search [searchTerm]="searchedWord"
-            [maxResults]="maxItems"
-            [skipResults]="skipCount"
-            (resultLoaded)="onSearchResultLoaded($event)"
-            #search>
-</adf-search>
+<div class="adf-search-results__facets">
```
eromano: if we add the new custom search here we don't test anymore the adf-search component in the demo shell could it be better have two separate test

DenisVuyka: Do we really need two search implementations? Given that first one has no design and requirements? I though we will have only latest one

DenisVuyka: We still have auto-completion and templating for the search input, the only thing that is replaced is "search results" page. Correct me if I'm wrong, but this is the "real" search results as expected from us, from what I understand, and previous implementation can be removed.

ohej: Denys is correct here, as long as we still have the "live search" (with autocomplete) with the customisable template this is good enough. The full search results page should be the new one with everything.

eromano: If we remove it is a BC can notbe removed

eromano: We still need to test it so is implementation need to be in the demo shell otherwise the Way team can not keep to check those functionality

demo-shell/src/app.config.json
eromano: Ok

demo-shell/src/app/components/search/search-result.component.html
eromano: The first implementation had requirements and I think this one should keep to work with it I will set up a meeting monday

DenisVuyka: Ok, I will restore old demo and have faceted one as a separate page with separate input

DenisVuyka: @eromano @ohej I will have two pages to have both demos up and running in the demo shell

eromano: @DenisVuyka I don't think honestly is just material of two pages . I think is more mater of make the new search a lint extension of the previous code and I think there is room to do it

DenisVuyka: Guys, can we have an architecture design for the entire epic before I spend 18 days fleshing out the container? It's it's not something ADF wants, then I need to know what exactly ADF wants

docs/content-services/search-filter.component.md
```diff
@@ -3,24 +3,21 @@ Added: v2.3.0
...
 
-![Search Settings overview](../docassets/images/search-settings-overview.png)
-
 ## Usage example
 
 ```html
-<adf-search-settings #settings
-    [queryBuilder]="queryBuilder">
+<adf-search-filter #settings>
 </adf-search-settings>
```
eromano: wrong close tag

PR #3136, opened 2018-03-29
[ADF-2500] fix trashcan bug plus refactoring documentlist 

eromano:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

fix trashcan bug plus refactoring documentlist 
https://issues.alfresco.com/jira/browse/ADF-2500

> - [ ] Yes
> - [ ] No

**Other information**:

### DIFF
--- a/lib/core/pagination/pagination-component.interface.ts
+++ b/lib/core/pagination/pagination-component.interface.ts
@@ -15,15 +15,10 @@
  * limitations under the License.
  */
 
+import { PaginatedComponent } from './paginated-component.interface';
 import { Pagination } from 'alfresco-js-api';
-import { BehaviorSubject } from 'rxjs/BehaviorSubject';
 
-export interface PaginatedComponent {
-    pagination: BehaviorSubject<Pagination>;
-    /**
-     * @deprecated 2.3.0 : the supported page size should be retrieved via the user preferences
-     * and given to the pagination component, and not retrieved by the paginated object
-     */
-    supportedPageSizes: number[];
-    updatePagination(params: Pagination);
+export interface PaginationComponentInterface {
+    target: PaginatedComponent;
+    pagination: Pagination;
 }
--- a/lib/core/services/search.service.ts
+++ b/lib/core/services/search.service.ts
@@ -17,11 +17,11 @@
 
 import { Injectable } from '@angular/core';
 import { NodePaging, QueryBody } from 'alfresco-js-api';
-import 'rxjs/add/observable/throw';
-import { Subject } from 'rxjs/Subject';
-
+import { Observable } from 'rxjs/Observable';
 import { AlfrescoApiService } from './alfresco-api.service';
+import 'rxjs/add/observable/throw';
 import { SearchConfigurationService } from './search-configuration.service';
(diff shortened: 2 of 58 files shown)


lib/core/pagination/pagination-component.interface.ts
```diff
@@ -15,17 +15,10 @@
...
- *
- * @TODO Contribute this to AlfrescoJS API
- */
+import { PaginatedComponent } from './paginated-component.interface';
+import { Pagination } from 'alfresco-js-api';
 
-export interface PaginationQueryParams {
-    skipCount: number;
-    maxItems: number;
+export interface PaginationComponentInterface {
```
DenisVuyka: I think adding "Interface" to the name of the interface is against all common practices. Also violates the https://angular.io/guide/styleguide#interfaces style guide. Otherwise we have to add "Class" and "Enum" to all classes and enums.

lib/core/services/search.service.ts
```diff
@@ -32,26 +32,45 @@ export class SearchService {
                 private searchConfigurationService: SearchConfigurationService) {
     }
 
-    async getNodeQueryResults(term: string, options?: SearchOptions): Promise<NodePaging> {
-        const data = await this.apiService.getInstance().core.queriesApi.findNodes(term, options);
+    getNodeQueryResults(term: string, options?: SearchOptions): Observable<NodePaging> {
```
DenisVuyka: Why it needs to be Observable? The method never streams anything, it's a 1-time operation here.

lib/core/services/search.service.ts
```diff
@@ -32,26 +32,45 @@ export class SearchService {
...
+        const promise = this.apiService.getInstance().core.queriesApi.findNodes(term, options);
```
DenisVuyka: Why you remove "await" here?

lib/core/services/search.service.ts
```diff
@@ -32,26 +32,45 @@ export class SearchService {
...
+
+        return Observable
+            .fromPromise(promise)
+            .catch(err => this.handleError(err));
     }
 
-    async search(searchTerm: string, maxResults: number, skipCount: number): Promise<NodePaging> {
-        const searchQuery = this.searchConfigurationService.generateQueryBody(searchTerm, maxResults, skipCount);
-        const data = await this.apiService.searchApi.search(searchQuery);
+    search(searchTerm: string, maxResults: number, skipCount: number): Observable<NodePaging> {
```
DenisVuyka: Same as above? What is the reason removing "async" here and switching back to Observable?

lib/core/services/search.service.ts
eromano: Is a braking change change it in promise we need to ripristinate it as observable plus all the other services are observable so is a consistent choose

lib/core/pagination/pagination-component.interface.ts
eromano: Ok

lib/core/services/search.service.ts
eromano: Same as above

PR #3141, opened 2018-04-04
[ADF-2562] Download a previous version

suzanadirla:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation
Further improvement on file versioning. Please see https://issues.alfresco.com/jira/browse/ADF-2562

> - [ ] Yes
> - [x] No

**Other information**:


eromano:
Why do we need the name parameter?

DenisVuyka:
@eromano I believe to be able to set the custom name to download all versions, including the cases when/if names are different or missing.

eromano:
Can we do it just automatically like name_ver_NUMBERVERSION? This name parameter sounds error prone. And even from it's description is not really linked to an override of the download name

DenisVuyka:
@eromano I don't think it is a good idea. Can we let end application developers decide on the final download name without any automatic naming? otherwise next bug request will be to provide custom naming support. We have similar feature in the Viewer (custom file name) and that's not auto-generated.

DenisVuyka:
Viewer allows setting custom download names: https://github.com/Alfresco/alfresco-ng2-components/blob/development/lib/core/viewer/components/viewer.component.ts#L520

eromano:
The name as you say will be the same.for any version I don't see the advantages . This scenario seems to be quite different from the viewer where you download only the last version

DenisVuyka:
@eromano in the PR, the name is taken from the "node", and propagated to the download method that generates the link based on id and uses node name as resulting file name. Can you give more details on what your concern is with that?

eromano:
Let's say I have 5 version uploaded. You use the filename given from input or from the API. All 5 versions will result with the same name when you download it. So I will no be able to identify which file is which version. If we do as I say. We will have :
Filename_ver_1
Filename_ver_2
Filename_ver_3
Filename_ver_4
Filename_ver_5

All this will happen automatically. Otherwise  you will have for all 5 versions:

Filename , filename

DenisVuyka:
So you suggest adding a version, not removing the "name"?

DenisVuyka:
Ok, got it.

DenisVuyka:
@eromano @suzanadirla 

Okay, this is what I found. It is not possible to change the name of the download file when using the "attachment link" (with content disposition headers from the server). Whatever value we provide - the name will always be the same. However browsers will append "(01)", "(02)", etc. 

We can disable "attachment: true" when generating the link, but we are then creating a security issue that we have already been addressing in Activiti, and the main reason for having "content-disposition" headers.

What are the thoughts?

eromano:
Seems that we need a feature from the BE, right?

DenisVuyka:
If we want a secure way, yes.

eromano:
I guess yes :)

DenisVuyka:
@eromano I can create a separate ticket for changing names, so that we can close this one. Changing names with BE might require a lot of time.

PR #3165, opened 2018-04-10
[ADF-2607] Task Form - Number Widget placeholder no longer displayed

camorra-skk:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

* https://issues.alfresco.com/jira/browse/ADF-2607

* Removed css to display placeholder after focus on input.

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/core/form/components/widgets/text/text.widget.spec.ts
+++ b/lib/core/form/components/widgets/text/text.widget.spec.ts
@@ -150,13 +150,16 @@ describe('TextWidgetComponent', () => {
                 expect(inputElement.placeholder).toBe('simple palceholder');
             });
 
-            it('should show the field placeholder when clicked', () => {
+            it('should show the field placeholder when clicked', async(() => {
                 inputElement.click();
                 fixture.detectChanges();
-                expect(inputElement).toBeDefined();
-                expect(inputElement).not.toBeNull();
-                expect(inputElement.placeholder).toBe('simple palceholder');
-            });
+                fixture.whenStable().then(() => {
+                    fixture.detectChanges();
+                    expect(inputElement).toBeDefined();
+                    expect(inputElement).not.toBeNull();
+                    expect(inputElement.placeholder).toBe('simple palceholder');
+                });
+            }));
 
             it('should prevent text to be written if is not allowed by the mask on keyUp event', async(() => {
                 expect(element.querySelector('#text-id')).not.toBeNull();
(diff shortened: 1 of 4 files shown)


lib/core/form/components/widgets/text/text.widget.spec.ts
```diff
@@ -269,10 +279,18 @@ describe('TextWidgetComponent', () => {
...
             });
 
             it('should show the input mask placeholder', () => {
-                const inputElement: HTMLInputElement = <HTMLInputElement> element.querySelector('#text-id');
+                expect(inputElement).toBeDefined();
+                expect(inputElement).not.toBeNull();
+                expect(inputElement.placeholder).toBe('Phone : (__) ___-___');
+            });
+
+            it('should show the input mask placeholder when clicked', () => {
```
VitoAlbano: this tests is related to the input mask placeholder that is different from the normal placeholder. Also i don't think the placeholder should be visibile when clicking on the widget.

lib/core/form/components/widgets/text/text.widget.ts
```diff
@@ -39,10 +39,12 @@ export class TextWidgetComponent extends WidgetComponent implements OnInit {
     }
 
     ngOnInit() {
-        if (this.field.params && this.field.params['inputMask']) {
-            this.mask = this.field.params['inputMask'];
+        if (this.field.params) {
             this.placeholder = this.field.params['inputMaskPlaceholder'] ? this.field.params['inputMaskPlaceholder'] : this.field.placeholder;
             this.isMaskReversed = this.field.params['inputMaskReversed'] ? this.field.params['inputMaskReversed'] : false;
+            if (this.field.params['inputMask']) {
```
VitoAlbano: do we need this if? will it work without?

VitoAlbano:
This will fix the problem only for the text widget, what about the number widget?

lib/core/form/components/widgets/text/text.widget.spec.ts
```diff
@@ -150,6 +150,14 @@ describe('TextWidgetComponent', () => {
                 expect(inputElement.placeholder).toBe('simple palceholder');
             });
 
+            it('should show the field placeholder when clicked', () => {
```
mauriziovitale: Can you use the async and whenStable method, please?

lib/core/form/components/widgets/text/text.widget.spec.ts
camorra-skk: The similar behaviour is in APS also. When we click on the text area the placeholder value is visible. It disappears only when value is entered in the text area.

lib/core/form/components/widgets/text/text.widget.spec.ts
camorra-skk: The test case is only to check the value of placeholder when clicked. There is no asynchronous call made. So I didnt add async and whenStable methods.

PR #3190, opened 2018-04-13
[ADF-2692] - refresh must be done on tab changes otherwise ViewChild obj refrence are null

VitoAlbano:

> - [X] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

On demoshell at the moment when we click on the active task from the process and then we complete the task, the processlist is not reloaded cause it is unreferenced due the tab.

When we click on tab we will reload the relative list that we are going to see so we will always have the correct results.

> - [ ] Yes
> - [ ] No

**Other information**:

### DIFF
--- a/demo-shell/src/app/components/process-service/process-service.component.ts
+++ b/demo-shell/src/app/components/process-service/process-service.component.ts
@@ -79,7 +79,7 @@ export class ProcessServiceComponent implements AfterViewInit, OnDestroy, OnInit
     @ViewChild('activitifilter')
     activitifilter: TaskFiltersComponent;
 
-    @ViewChild('banano')
+    @ViewChild('processListPagination')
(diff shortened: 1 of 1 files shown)


demo-shell/src/app/components/process-service/process-service.component.ts
```diff
@@ -78,7 +78,7 @@ export class ProcessServiceComponent implements AfterViewInit, OnDestroy, OnInit
     @ViewChild('activitifilter')
     activitifilter: TaskFiltersComponent;
 
-    @ViewChild('processListPagination')
+    @ViewChild('banano')
```
eromano: I guess this name is not the right name

VitoAlbano: no it's not

PR #3278, opened 2018-05-08
[ADF-2626] added preserved state functionality for sidenav component

georgiana-roman:

> - [x ] The commit message follows our [guidelines](https://github.com/Alfresco/alfresco-ng2-components/wiki/Commit-format)
> - [x ] Docs have been added / updated (for bug fixes / features)

> - [ ] Bugfix
> - [x ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://issues.alfresco.com/jira/browse/ADF-2626

The sideNav collapsed/expanded state is preserved between sessions

> - [ ] Yes
> - [ ] No

**Other information**:


demo-shell/src/app/components/app-layout/app-layout.component.ts
```diff
@@ -48,6 +50,18 @@ export class AppLayoutComponent {
         { href: '/about', icon: 'info_outline', title: 'APP_LAYOUT.ABOUT' }
     ];
 
-    constructor() {
+    @Input() expandedSidenav = false;
+
+    constructor(private storage: StorageService, private config: AppConfigService) {
+        const preserved = this.storage.getItem('openedSidenav');
```
popovicsandras: Can you please move this logic into ngOnInit? constructors are supposed to be as "flat" as possible.

And I'm not sure this logic is correct. 
First we would need to check the config, if we need to read from localstorage. If so we can read from localstorage, and set the value if it is already persisted in the local storage, otherwise use the config's openedSidenav value (if it is set!!!) at last but not least, the default value of the component's.

lib/core/sidenav-layout/components/sidenav-layout/sidenav-layout.component.ts
```diff
@@ -56,7 +59,9 @@ export class SidenavLayoutComponent implements OnInit, AfterViewInit, OnDestroy
         isMenuMinimized: () => this.isMenuMinimized
     };
 
-    constructor(private mediaMatcher: MediaMatcher) {
+    constructor(private mediaMatcher: MediaMatcher,  private config: AppConfigService, private storage: StorageService) {
```
popovicsandras: The preservation should not be part of this component. The component should communicate the collapsed/opened state with an event bubbling up, and in the application, where we use this component we can do the saving to the local storage. For example in our case in AMA, everything would go through the store, and as a side effect we could  save to the localstorage.

demo-shell/src/app/components/app-layout/app-layout.component.ts
```diff
@@ -48,6 +50,25 @@ export class AppLayoutComponent {
         { href: '/about', icon: 'info_outline', title: 'APP_LAYOUT.ABOUT' }
     ];
 
-    constructor() {
+    expandedSidenav: Boolean = false;
```
popovicsandras: We should avoid using capitalised types, see this:
https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html

And when we have an initial value set, type declarations are not needed, type will be inferred.

demo-shell/src/app/components/app-layout/app-layout.component.ts
```diff
@@ -48,6 +50,25 @@ export class AppLayoutComponent {
...
+
+
+    ngOnInit() {
+        const expand = this.config.get<boolean>('sideNav.expandedSidenav');
+        const preserveState = this.config.get('sideNav.preserveState');
+
+        if (preserveState && expand)  {
+            this.expandedSidenav = (this.userpreference.get('expandedSidenav', String(expand)) === 'true');
```
popovicsandras: I'm confused about this String expression here.
according to how the "expand" is retrieved I presume, the expand is already a boolean (this.config.get&lt;boolean&gt;). Why don't we just compare it to a boolean value, why is the casting needed?
expand === true

demo-shell/src/app/components/app-layout/app-layout.component.html
```diff
@@ -1,4 +1,4 @@
-<adf-sidenav-layout [sidenavMin]="70" [sidenavMax]="220" [stepOver]="780" [hideSidenav]="false" [expandedSidenav]="false">
+<adf-sidenav-layout [sidenavMin]="70" [sidenavMax]="220" [stepOver]="780" [hideSidenav]="false"[expandedSidenav]= "expandedSidenav" (notify)='setState($event)'>
```
popovicsandras: In template we should use double quotes instead on single ones.

lib/core/sidenav-layout/components/sidenav-layout/sidenav-layout.component.ts
```diff
@@ -38,6 +38,8 @@ export class SidenavLayoutComponent implements OnInit, AfterViewInit, OnDestroy
     @Input() hideSidenav = false;
     @Input() expandedSidenav = true;
 
+    @Output() notify = new EventEmitter();
```
popovicsandras: Generic type of the EventEmitter is missing. I suppose it would be EventEmitter<boolean>. Instead of this general "notify" name, what about using a more meaningful name, like "expanded".

lib/core/sidenav-layout/components/sidenav-layout/sidenav-layout.component.ts
```diff
@@ -81,14 +84,15 @@ export class SidenavLayoutComponent implements OnInit, AfterViewInit, OnDestroy
         this.mediaQueryList.removeListener(this.onMediaQueryChange);
     }
 
-    toggleMenu() {
+    toggleMenu(expandedSidenav) {
```
popovicsandras: Where is the argument expandedSidenav used in this method? Do we need it?

lib/core/sidenav-layout/components/sidenav-layout/sidenav-layout.component.ts
```diff
@@ -81,14 +84,15 @@ export class SidenavLayoutComponent implements OnInit, AfterViewInit, OnDestroy
...
         if (!this.mediaQueryList.matches) {
             this.isMenuMinimized = !this.isMenuMinimized;
         } else {
             this.isMenuMinimized = false;
         }
 
         this.container.toggleMenu();
+        this.notify.emit((!this.isMenuMinimized).toString());
```
popovicsandras: Please emit a boolean. How we want to store this value should not have any effect on what is emitted.

demo-shell/src/app/components/app-layout/app-layout.component.ts
georgiana-roman: The get method from user preference service require a string param, and also returns a string, that's why I did the casting.

lib/core/pipes/app-config.pipe.ts
```diff
@@ -0,0 +1,31 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { Pipe, PipeTransform } from '@angular/core';
+import { AppConfigService } from './../app-config/app-config.service';
+
+@Pipe({
+    name: 'appConfig',
+    pure: true
```
eromano: pipe are always pure by default https://angular.io/guide/pipes#pure-and-impure-pipes

eromano: where do we use this new pipe?

eromano: do we need it pure or impure?

DenisVuyka: It is used in ACA right now, for example: 
```html
<adf-pagination
                [supportedPageSizes]="'pagination.supportedPageSizes' | appConfig"
                [target]="documentList"
                (changePageSize)="onChangePageSize($event)">
            </adf-pagination>
```

DenisVuyka: I think `pure` (default) is okay. We usually need evaluating config values only once

georgiana-roman: That is indeed the pipe Denis did for ACA. I initially planned to use it here, but I've ended not using it. I'll remove it, sorry.

eromano:
can we add a documentation for this pipe?

PR #3279, opened 2018-05-08
[ADF-2916] number range form validator added

bbcodrin:

> - [X] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

If FROM value is bigger than TO value, we still make an backend call, even if we know that list will be empty.
https://issues.alfresco.com/jira/browse/ADF-2916

That behaviour is deactivated now. If FROM value is bigger than TO value, APPLY button will be disabled.

> - [ ] Yes
> - [X] No

**Other information**:

### DIFF
--- a/lib/content-services/search/components/search-number-range/search-number-range.component.ts
+++ b/lib/content-services/search/components/search-number-range/search-number-range.component.ts
@@ -62,7 +62,11 @@ export class SearchNumberRangeComponent implements SearchWidget, OnInit {
         this.form = new FormGroup({
             from: this.from,
             to: this.to
-        });
+        }, this.formValidator);
+    }
+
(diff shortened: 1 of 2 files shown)


lib/content-services/search/components/search-number-range/search-number-range.component.ts
```diff
@@ -53,7 +53,11 @@ export class SearchNumberRangeComponent implements SearchWidget, OnInit {
         this.form = new FormGroup({
             from: this.from,
             to: this.to
-        });
+        }, this.formValidator);
+    }
+
+    formValidator(g: FormGroup) {
```
eromano: please do not use one character variable name are really bad for readability

PR #3323, opened 2018-05-15
[ADF-2984] Show date invalid message on search date range picker

suzanadirla:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation
Improve validation messages on date range picker.

Please see Jira https://issues.alfresco.com/jira/browse/ADF-2984

> - [ ] Yes
> - [x] No

**Other information**:


eromano:
In the ticket is not clear how to test it

eromano:
Two problems:

1. I guess you need some margin in the bottom, the error message is to close to the clear button:

screen shot 2018-05-17 at 10 11 56

2. All the time that i try to insert a different date with a wrong format it is automatically changed. I think the right behaviour should just raise an error not  guess the date:
may-17-2018 10-14-23

suzanadirla:
I've added a change for the problem no. 1.
As for problem no. 2, this is the default behavior of the datepicker when using the recommend MomentDateAdapter. If we do need a custom adapter, it would be better to have a separate story for this as there might be other behaviors we might want to be changed.

eromano:
We do use the moment data adapter also in other places without having this problem. Can you give a look to the start task component? This behaviour has to be the same in all ADF

suzanadirla:
Eugenio, I was talking about the `MomentDateAdapter` from `@angular/material-moment-adapter`. The start task uses a custom date adapter with the same `MomentDateAdapter` name. You mean I should make changes to use that custom adapter in here?

eromano:
can you try?

suzanadirla:
Sure. I'll create a task and separate PR for that - to make the Search Date Range Widget use the custom `MomentDateAdapter` from `@alfresco/adf-core`.

PR #3451, opened 2018-06-07
[ADF-3147] AppsList Enanchement

sivakumar414ram:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://issues.alfresco.com/jira/browse/ADF-3147

> - [ ] Yes
> - [x] No

**Other information**:


eromano:
can we have just a toggle in the current applist to change the view instead to have another page?

lib/process-services/app-list/apps-list.component.ts
```diff
@@ -50,10 +52,29 @@ export class AppsListComponent implements OnInit, AfterContentInit {
     @Input()
     filtersAppId: any[];
 
+    /* Toggles multiple row selection, renders checkboxes at the beginning of each row */
+    @Input()
+    multiselect: boolean = false;
```
eromano: The problem that I see with this solution is that we try to merge two things togheter that are actually really different. So when you seee the apps as tails most of those property are not possible to use. It will be difficult explain it.
I guess we need to split this different view it in two different component. Or make available the same options even for the tiles view

PR #3537, opened 2018-06-27
[ADF-3177] added general header component

georgiana-roman:

> - [ ] Bugfix
> - [x] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://issues.alfresco.com/jira/browse/ADF-3177

 general reusable header component for all of the Alfresco applications was added

> - [ ] Yes
> - [x] No

**Other information**:


lib/core/core.module.ts
```diff
@@ -39,7 +39,7 @@ import { ToolbarModule } from './toolbar/toolbar.module';
 import { UserInfoModule } from './userinfo/userinfo.module';
 import { ViewerModule } from './viewer/viewer.module';
 import { FormModule } from './form/form.module';
-import { SidenavLayoutModule } from './sidenav-layout/sidenav-layout.module';
+import { LayoutModule as SidenavLayoutModule } from './sidenav-layout/layout.module';
```
popovicsandras: Can you please rename the module's folder also from sidenav-layout to layout (from this point on, this folder is supposed to containe every layout related general elements).

lib/core/sidenav-layout/components/header/header.component.ts
```diff
@@ -0,0 +1,37 @@
...
+@Component({
+    selector: 'adf-layout-header',
+    templateUrl: './header.component.html',
+    styleUrls: ['./header.component.scss']
+})
+
+export class HeaderLayoutComponent {
+
+    @Input() title: string;
+    @Input() logo?: string;
```
popovicsandras: Sorry, it was wrongly inside the description. We would like to have the Alfresco flower as logo by default, which can be overridden by a custom one.

lib/core/sidenav-layout/components/header/header.component.html
```diff
@@ -0,0 +1,11 @@
+<mat-toolbar color="{{color}}">
+    <button class="mat-icon-button" mat-icon-button="" (click)="toggleMenu()">
+        <mat-icon class="mat-icon material-icons" role="img" aria-hidden="true">menu</mat-icon>
+    </button>
+
+    <img *ngIf="logo" src="{{logo}}" class="adf-app-logo">
```
popovicsandras: As I wrote before this was wrong in the description. We would like to have always a logo, by default the Alfresco flower, and it can be overridden by input.

lib/core/sidenav-layout/components/header/header.component.scss
```diff
@@ -0,0 +1,43 @@
+@mixin  adf-header-layout-theme($theme) {
+    $primary: map-get($theme, primary);
+    $accent: map-get($theme, accent);
+    $background: map-get($theme, background);
+
+    adf-layout-header .mat-toolbar-single-row {
```
popovicsandras: Please check again the guideline, I think the are some problems with the padding around  and between the buttons (hamburger menu, left side, right side)

lib/core/sidenav-layout/components/header/header.component.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+
+    it('should display img element with the expected src if a logo path is set', () => {
+        component.logo = 'logo.png';
+        fixture.detectChanges();
+
+        const logo = fixture.nativeElement.querySelector('.adf-app-logo');
+        const src = logo.getAttribute('src');
+        expect(logo === null).toBeFalsy;
+        expect(src).toEqual('logo.png');
+    });
```
popovicsandras: can you please add a test for the projected content (transclusion)?

lib/core/layout/components/header/header.component.ts
```diff
@@ -0,0 +1,37 @@
...
+    styleUrls: ['./header.component.scss']
+})
+
+export class HeaderLayoutComponent {
+
+    @Input() title: string;
+    @Input() logo: string = '../assets/logo.png';
+    @Input() color: string;
+    @Output() toggled = new EventEmitter<boolean>();
+    expanded = false;
```
popovicsandras: Why do we need this? The expanded value is never modified. I think instead of "toggled" output parameter, we can name it "clicked" and it can be an EventEmitter<void> since, we are not using this value

lib/core/layout/components/header/header.component.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+        it('should display the img element with the expected src if a logo path is set', () => {
+            component.logo = 'logo.png';
+            fixture.detectChanges();
+
+            const logo = fixture.nativeElement.querySelector('.adf-app-logo');
+            const src = logo.getAttribute('src');
+            expect(logo === null).toBeFalsy();
+            expect(src).toEqual('logo.png');
+        });
+    });
```
popovicsandras: toggled/clicked is not tested.

lib/core/layout/components/header/header.component.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+});
+
+describe('Template tranclusion', () => {
```
popovicsandras: I'm not sure ComponentFactoryResolver is needed for test this behaviour. Let's do it together on Monday.

lib/core/layout/components/header/header.component.ts
```diff
@@ -0,0 +1,37 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { Component, Input, Output, EventEmitter } from '@angular/core';
+
+@Component({
+    selector: 'adf-layout-header',
+    templateUrl: './header.component.html',
+    styleUrls: ['./header.component.scss']
```
DenisVuyka: Please add `ViewEncapsulation.None` and set the host class name to the `adf-layout-header` like we have for other components.

lib/core/layout/components/header/header.component.html
```diff
@@ -0,0 +1,11 @@
+<mat-toolbar color="{{color}}">
+    <button class="mat-icon-button adf-menu-icon" mat-icon-button="" (click)="toggleMenu()">
```
mauriziovitale: What if I don't want to have a button? ( Think about APW, when we show the apps we don't have any sidenav)

DenisVuyka: I agree, maybe we have `showSidenavToggle` property or something like that?

lib/core/core.module.ts
```diff
@@ -39,7 +39,7 @@ import { ToolbarModule } from './toolbar/toolbar.module';
 import { UserInfoModule } from './userinfo/userinfo.module';
 import { ViewerModule } from './viewer/viewer.module';
 import { FormModule } from './form/form.module';
-import { SidenavLayoutModule } from './sidenav-layout/sidenav-layout.module';
+import { LayoutModule as SidenavLayoutModule } from './layout/layout.module';
```
DenisVuyka: We should not be renaming the components internally I think.

lib/core/layout/components/header/header.component.html
```diff
@@ -0,0 +1,11 @@
+<mat-toolbar color="{{color}}">
+    <button *ngIf="showSidenavToggle" class="mat-icon-button adf-menu-icon" mat-icon-button (click)="toggleMenu()">
```
DenisVuyka: This property is missing in the documentation

lib/core/core.module.ts
popovicsandras: After discussed it on Slack with @DenisVuyka , until we export the module with the old (@deprecated) name, it is ok.

PR #3551, opened 2018-07-02
[ADF-3274] people widget instead of the people list

bbcodrin:

> - [ ] Bugfix
> - [X] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://issues.alfresco.com/jira/browse/ADF-3274
Right now, the start form component is using the people list to render the assignee. 

We should use the people widget to achieve that as we do for the adf form.

> - [ ] Yes
> - [X] No

**Other information**:


lib/process-services/task-list/components/start-task.component.html
```diff
@@ -65,7 +65,7 @@
                 </div>
 
                 <div class="adf-grid-column adf-grid-half-width">
-                    <adf-people-selector [(peopleId)]="assigneeId" id="assignee_id"></adf-people-selector>
+                    <people-widget [(peopleId)]="assigneeId" [field]="field"></people-widget>
```
mauriziovitale: is the people widget able to show the person?
I think we are only using the output and not the two way binding
If so we can change this code

 ```@Output()
+    peopleIdChange: EventEmitter<number>;```
to
 ```@Output()
+    peopleSelected: EventEmitter<number>;```

lib/process-services/task-list/components/start-task.component.ts
```diff
@@ -81,6 +85,7 @@ export class StartTaskComponent implements OnInit {
     }
 
     ngOnInit() {
+        this.field = new FormFieldModel(new FormModel(), {id: 'fakeId', value: 'fakeValue', placeholder: 'Assignee'});
```
mauriziovitale: would be better remove the fakeId and use assigneeId 
same for value you can use assigneeId I guess

mauriziovitale: I don't see a unit test about that. can you add one?

mauriziovitale: Even the documentation is missing

VitoAlbano: I think this highlights a problem of trying to use this widget here . If this is needed like that wouldn't it be better try to create a common one that could be extended by the widget(which has the form) and the one you need?

mauriziovitale:
opened a different issue

PR #3563, opened 2018-07-05
[ADF_NOISSUE] Header image fix

popovicsandras:

> - [ ] Bugfix
> - [ ] Feature
> - [x] Build related changes
> - [ ] Documentation

> - [ ] Yes
> - [x] No

**Other information**:


lib/core/styles/_index.scss
```diff
@@ -1,35 +1,35 @@
-@import './theme-colors';
-@import './default-class';
-@import './theming';
+@import "./theme-colors";
```
eromano: it should be single please remove the auto save option from vs code 👎

popovicsandras: We have rules for ts files, but we have no style linter at the moment. If we had, there would be no problem for VS Code Prettier plugin. :P
Or am I wrong about the existence of style linnter? 🤔

eromano: we need to add one for the scss you are right

PR #3592, opened 2018-07-13
[ADF-3362][ADF-3361] Search on document picker with custom site list - fix

suzanadirla:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

Please see https://issues.alfresco.com/jira/browse/ADF-3362 and https://issues.alfresco.com/jira/browse/ADF-3361

Fixes issues https://issues.alfresco.com/jira/browse/ADF-3362 and https://issues.alfresco.com/jira/browse/ADF-3361

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/content-services/content-node-selector/content-node-selector-panel.component.ts
+++ b/lib/content-services/content-node-selector/content-node-selector-panel.component.ts
@@ -33,6 +33,7 @@ import { CustomResourcesService } from '../document-list/services/custom-resourc
 export type ValidationFunction = (entry: MinimalNodeEntryEntity) => boolean;
 
 const defaultValidation = () => true;
+const ALL_ITEMS_PAGE_SIZE = 999;
 
 @Component({
     selector: 'adf-content-node-selector-panel',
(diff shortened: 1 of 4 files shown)


eromano:
Two things:
1. can you reactor this method to have only one return?
2. Where is used this method without passing the nodeId?

suzanadirla:
Ok, @eromano, I’m working on a refactoring at the moment.

lib/content-services/content-node-selector/content-node-selector-panel.component.ts
```diff
@@ -255,8 +257,8 @@ export class ContentNodeSelectorPanelComponent implements OnInit, PaginatedCompo
     private querySearch(): void {
         this.loadingSearchResults = true;
 
-        if (this.dropdownSiteList) {
-            this.documentList.getCorrespondingNodeIds(this.siteId)
+        if (this.customResourcesService.hasCorrespondingNodeIds(this.siteId)) {
+            this.customResourcesService.getCorrespondingNodeIds(this.siteId, { maxItems: 999 })
```
eromano: magic number 999 we should never add this kind of fix number in the code

eromano:
without any test is not clear to me which problem we are trying to address

suzanadirla:
I tried to remove the getCorrespondingNodeIds method and use the loadFolderByNodeId method instead... but the problem is that the IDs for custom sources are in different properties: for example for `sharedLinks` it is the `nodeId`, for the sites is the `guid`, for the favorites it is the `targetGuid` :neutral_face: so I need to handle these cases and would end up with another method for this.
In this case, I don’t think it is worth the effort to change the getCorrespondingNodeIds so that it uses the loadFolderByNodeId inside...

PR #3915, opened 2018-10-24
[ADF-3696] Create an automated test to cover the search filter buckets scenario

jdosti:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation
Automated test to cover the search filter buckets scenario

> - [ ] Yes
> - [ ] No

**Other information**:
https://issues.alfresco.com/jira/browse/ADF-3696

### DIFF
--- a/e2e/content-services/search/search_filters.e2e.ts
+++ b/e2e/content-services/search/search_filters.e2e.ts
@@ -57,8 +57,6 @@ describe('Search Filters', () => {
 
     let fileUploaded, fileTypePng;
 
-    // let filterType = 'TYPE-PNG Image';
-
     let filter = {
         type: 'TYPE-PNG Image',
         name: 'PNG Image ('
--- a/e2e/content-services/search/search_filters.e2e.ts
+++ b/e2e/content-services/search/search_filters.e2e.ts
@@ -56,10 +56,7 @@ describe('Search Filters', () => {
 
     let fileUploaded, fileTypePng;
 
-    let filter = {
-        type: 'TYPE-PNG Image',
-        name: 'PNG Image ('
-    };
+    let filter = {type: 'TYPE-PNG Image'};
 
     beforeAll(async (done) => {
 
@@ -138,7 +135,7 @@ describe('Search Filters', () => {
 
         searchFiltersPage.clickPngImageType();
 
-        let bucketNumberForFilter = searchFiltersPage.getBucketNumberOfFilterType(filter.type, filter.name.length);
+        let bucketNumberForFilter = searchFiltersPage.getBucketNumberOfFilterType(filter.type);
 
         let resultFileNames = contentList.getAllRowsNameColumn();
 
--- a/e2e/content-services/search/search_filters.e2e.ts
+++ b/e2e/content-services/search/search_filters.e2e.ts
@@ -18,12 +18,16 @@
 import { LoginPage } from '../../pages/adf/loginPage';
 import SearchDialog = require('../../pages/adf/dialog/searchDialog');
 import { SearchFiltersPage } from '../../pages/adf/searchFiltersPage';
+import PaginationPage = require('../../pages/adf/paginationPage');
+import ContentList = require('../../pages/adf/dialog/contentList');
+
 
 import AcsUserModel = require('../../models/ACS/acsUserModel');
 import FileModel = require('../../models/ACS/fileModel');
 
 import TestConfig = require('../../test.config');
 import Util = require('../../util/util');
+import resources = require('../../util/resources');
 
 import AlfrescoApi = require('alfresco-js-api-node');
 import { UploadActions } from '../../actions/ACS/upload.actions';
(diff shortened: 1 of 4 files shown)


e2e/content-services/search/search_filters.e2e.ts
```diff
@@ -100,4 +119,24 @@ describe('Search Filters', () => {
...
+
+        let bucketNumberForFilter = searchFiltersPage.getBucketNumberOfFilterType(filter.type, filter.name.length);
+
+        let resultFileNames = contentList.getAllRowsNameColumn();
+
+        expect(bucketNumberForFilter).not.toEqual('0');
+
+        expect(paginationPage.getTotalNumberOfFiles()).toEqual(bucketNumberForFilter);
+
+        resultFileNames.then((fileNames) => {
```
cristinaj: I think in this case this line is not necessary. I think it will work like this: 
resultFileNames.map((nameOfResultFiles) => {
                expect(nameOfResultFiles).toContain('.png');
            });

e2e/pages/adf/dialog/contentList.js
```diff
@@ -28,7 +28,7 @@ var ContentList = function () {
     var actionMenu = element(by.css("div[role='menu']"));
     var optionButton = by.css("button[data-automation-id*='action_menu_']");
     var rowByRowName = by.xpath("ancestor::div[contains(@class, 'adf-datatable-row')]");
-    var nameColumn = by.css("div[id*='document-list-container'] div[class*='adf-datatable-row'] div[class*='--text full-width'] span");
+    var nameColumn = by.css('div[class*="adf-data-table-cell"][title="Display name"]');
```
cristinaj: This locator will retrieve the name of the column also('Display name' in our case) that may make other tests fail. I suggest to change the locator to this one: by.css('div[class*="datatable-body"] div[class*="adf-data-table-cell"][title="Display name"]')

e2e/pages/adf/paginationPage.js
```diff
@@ -130,6 +131,16 @@ var PaginationPage = function () {
         Util.waitUntilElementIsNotOnPage(previousButtonDisabled);
     };
 
+    this.getTotalNumberOfFiles = function () {
+        Util.waitUntilElementIsVisible(totalFiles);
+        var numberOfFiles = totalFiles.getText().then(function (totalNumber) {
+          var totalNumberOfFiles = totalNumber.substring(16, totalNumber.length);
```
cristinaj: Can you use, instead of '16', the character that is at that position?

e2e/pages/adf/searchFiltersPage.ts
```diff
@@ -153,4 +154,20 @@ export class SearchFiltersPage {
...
+    clickPngImageType() {
+        Util.waitUntilElementIsVisible(this.pngImageFileType);
+        return this.pngImageFileType.click();
+    }
+
+    getBucketNumberOfFilterType(fileType, filterLength) {
+        let fileTypeFilter = element(by.css('mat-checkbox[data-automation-id="checkbox-SEARCH.FACET_FIELDS.' + fileType + '"] span'));
+        Util.waitUntilElementIsVisible(fileTypeFilter);
+        let bucketNumber = fileTypeFilter.getText().then((valueOfBucket) => {
+            let numberOfBucket = valueOfBucket.substring(filterLength, valueOfBucket.length - 1);
```
cristinaj: Cn you use instead of filterLength the last character of filter.name (in this case '(') ?

e2e/content-services/search/search_filters.e2e.ts
jdosti: I tried this method but it's not working like that.

e2e/pages/adf/paginationPage.js
jdosti: This method takes the total number of the text "Showing 1-20 of 26415" and in this case the starting point will always remain 16. I can assign the number to a variable, but it still will be 16.

e2e/pages/adf/searchFiltersPage.ts
jdosti: That is not working either, it requires the length of the total characters that are not needed (in this case "PNG Image (" ), not the starting point of the characters that are needed.

e2e/content-services/search/search_filters.e2e.ts
cristinaj: Ok, then we can leave it as it is for now. Probably it's not working cause we do a defer fulfil in contentList.getAllRowsNameColumn().

e2e/pages/adf/paginationPage.js
cristinaj: If you will be on the second page when you will want to use this method, i think 16 will not work anymore, cause the text will be "Showing 21-40 of 26415". I think it's better to use a character/string instead of the position. In this case, you could split by 'of '.

PR #4084, opened 2018-12-14
[APM-419] Adding data input to dynamic component

popovicsandras:

> - [ ] Bugfix
> - [x] Feature
> - [ ] Build related changes
> - [ ] Documentation

We lack the ability to pass through data for dynamically loaded components in the adf-extensions package.

Now we have it.

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/extensions/src/lib/components/dynamic-component/dynamic.component.ts
+++ b/lib/extensions/src/lib/components/dynamic-component/dynamic.component.ts
@@ -29,6 +29,7 @@ import {
 import { ExtensionService } from '../../services/extension.service';
 import { ExtensionComponent } from '../../services/component-register.service';
 
+// cSpell:words lifecycle
 @Component({
     selector: 'adf-dynamic-component',
     template: `<div #content></div>`
--- a/lib/extensions/src/lib/components/dynamic-component/dynamic.component.ts
+++ b/lib/extensions/src/lib/components/dynamic-component/dynamic.component.ts
@@ -44,20 +44,7 @@ export class DynamicExtensionComponent implements OnChanges, OnDestroy {
     private componentRef: ComponentRef<ExtensionComponent>;
     private loaded: boolean = false;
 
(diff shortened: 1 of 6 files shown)


lib/extensions/src/lib/components/dynamic-component/dynamic.component.ts
```diff
@@ -19,49 +19,98 @@ import {
...
         }
     }
+
+    private updateInstance() {
+        if (this.componentCreated()) {
+            this.componentRef.instance.data = this.data;
+        }
+    }
+
+    private proxy(lifecycleMethod, ...args) {
```
DenysVuika: When created dynamically, child component will already have own lifecycle methods, like onInit, of onDestroy. You may be calling them 2 times now (might cause errors with double destroy or double init)

DenysVuika: I suggest checking manually at runtime, by putting console.log in the component (to avoid components not being ready during test run). Angular should wire those with all life-cycle stuff properly. The only issue I had is propagating custom data changes down to the child ([example](https://github.com/Alfresco/alfresco-content-app/blob/master/src/app/components/common/dynamic-column/dynamic-column.component.ts#L67)) to have like `<dynamic-component [context]="node">` and then `node` value being passed to constructed child each time it changes.

popovicsandras: Ok, thanks, I'm checking it now, and try to cover it with better tests.

popovicsandras: Tested it in both the testing environment and browser. You are right. Testing it in browser, every lifecycle method of the dynamically created component was invoked properly without the need of patching it. However the testing environment for some reason can't do the same [which is very misleading]. So I remove those patches and tests.

I left the patch only for the ngOnChanges as you suggested, and the ngOnDestroy will be called because of the componentRef.destroy method.

PR #4154, opened 2019-01-14
[ADF-3906] extend SSO docs

DenysVuika:
Update SSO docs.


DenysVuika:
I used fancy VS code plugin to create PR, and it misses the standard issue template, sorry.

PR #4158, opened 2019-01-15
[ADF-3854] destination picker - exclude site content after rowFilter reset

suzanadirla:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

If the rowFilter parameter is changed, then the excludeSiteContent is ignored

Exclude the specified site content even after rowFilter param changes.
please check comment on https://issues.alfresco.com/jira/browse/ADF-3854

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/content-services/content-node-selector/content-node-selector-panel.component.ts
+++ b/lib/content-services/content-node-selector/content-node-selector-panel.component.ts
@@ -165,7 +165,9 @@ export class ContentNodeSelectorPanelComponent implements OnInit {
         this.breadcrumbTransform = this.breadcrumbTransform ? this.breadcrumbTransform : null;
         this.isSelectionValid = this.isSelectionValid ? this.isSelectionValid : defaultValidation;
         this.excludeSiteContent = this.excludeSiteContent ? this.excludeSiteContent : [];
-        this.rowFilter = this.getRowFilter(this.rowFilter);
+        if (!this.rowFilter) {
+            this.rowFilter = this.getRowFilter(this.rowFilter);
+        }
     }
 
     ngOnChanges(changes: SimpleChanges) {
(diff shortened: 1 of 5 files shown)


lib/content-services/content-node-selector/content-node-selector-panel.component.ts
```diff
@@ -167,7 +176,15 @@ export class ContentNodeSelectorPanelComponent implements OnInit, PaginatedCompo
         this.breadcrumbTransform = this.breadcrumbTransform ? this.breadcrumbTransform : null;
         this.isSelectionValid = this.isSelectionValid ? this.isSelectionValid : defaultValidation;
         this.excludeSiteContent = this.excludeSiteContent ? this.excludeSiteContent : [];
-        this.rowFilter = this.getRowFilter(this.rowFilter);
+        if (!this.rowFilter) {
```
DenysVuika: I think you don't need this one, as ngOnChanges will be setting the value anyway. But I suggest to remove and check with a unit test that it still works as expected.

suzanadirla:
> ```
>     this.rowFilter = this.getRowFilter(this.rowFilter); is also in the oninit like that is goint to be executed twice the first time
> ```

I must specify that there is a case when the call inside the `ngOnInit` is needed:  when the `ContentNodeSelectorPanel` has no value set for the `rowFilter`, but it does have one for the `excludeSiteContent`.

If I remove the call completely from the `ngOnInit`, then the filtering out based on the `excludeSiteContent` is not done (because on the first `ngOnChanges` call, the `changes.rowFilter` is undefined... so the `getRowFilter` would never be called )
Such broken filtering could be observed here: http://adfdev.envalfresco.com/node-selector if I remove the [rowFilter] property from the ‘Exclude Site Content’ panel and also that call from ngOnInit
Of course, the unit tests don't pass without that code - even when trying to adapt them.

I hope this clarifies better this case.

PR #4164, opened 2019-01-15
Revert "[ADF-3911] Converted GIF image URLs to absolute"

therealandeeee:
Reverts Alfresco/alfresco-ng2-components#4159


therealandeeee:
Apparently, they don't need this now.

PR #4174, opened 2019-01-17
[ADF-3923] Favorites - remove favorite node doesn't work

pionnegru:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://issues.alfresco.com/jira/browse/ADF-3923

> - [ ] Yes
> - [ ] No

**Other information**:

### DIFF
--- a/lib/core/directives/node-favorite.directive.ts
+++ b/lib/core/directives/node-favorite.directive.ts
@@ -69,8 +69,7 @@ export class NodeFavoriteDirective implements OnChanges {
             const batch = this.favorites.map((selected) => {
                 // shared files have nodeId
                 const id = selected.entry.nodeId || selected.entry.id;
-
-                return from(this.alfrescoApiService.favoritesApi.removeFavoriteSite('-me-', id));
+                return from(this.alfrescoApiService.favoritesApi.favoritesApi.deleteFavorite('-me-', id));
             });
 
             forkJoin(batch).subscribe(
--- a/lib/core/directives/node-favorite.directive.spec.ts
+++ b/lib/core/directives/node-favorite.directive.spec.ts
@@ -175,7 +175,7 @@ describe('NodeFavoriteDirective', () => {
         let addFavoriteSpy;
 
         beforeEach(() => {
-            removeFavoriteSpy = spyOn(alfrescoApiService.getInstance().core.favoritesApi, 'removeFavoriteSite').and.callThrough();
+            removeFavoriteSpy = spyOn(alfrescoApiService.getInstance().core.favoritesApi.favoritesApi, 'deleteFavorite').and.callThrough();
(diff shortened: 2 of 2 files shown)


lib/core/directives/node-favorite.directive.ts
```diff
@@ -69,8 +69,7 @@ export class NodeFavoriteDirective implements OnChanges {
             const batch = this.favorites.map((selected) => {
                 // shared files have nodeId
                 const id = selected.entry.nodeId || selected.entry.id;
-
-                return from(this.alfrescoApiService.favoritesApi.removeFavoriteSite('-me-', id));
+                return from(this.alfrescoApiService.favoritesApi.favoritesApi.deleteFavorite('-me-', id));
```
DenysVuika: why there's 2 favoritesApi properties? is that a typo or problem at the js-api level?

DenysVuika: I mean `alfrescoApiService.favoritesApi.favoritesApi`

eromano: One is the get of the alfrescoApiService. that is quite ugly to see.
```
   get favoritesApi(): Core.FavoritesApi {
        return this.getInstance().core.favoritesApi;
    }
```
would be better remove this get from the alfrescoApiService and do so just evrywhere
```
        return from(this.alfrescoApiService.core.favoritesApi.deleteFavorite('-me-', id));
```

pionnegru: the problem is that 'deleteFavorite' si not a member of core. favoritesApi

pionnegru: it seems I need to access the new api methods via legacy api 
https://github.com/Alfresco/alfresco-js-api/blob/development/src/api-legacy/content-rest-api/src/api/favoritesApi.ts#L29

eromano: this is the comaptibility layer id you call 
```
        return from(this.alfrescoApiService.favoritesApi.removeFavoriteSite('-me-', id));
```

removeFavoriteSite is the old name that proxy the call to the new method in the new method deleteSiteFavorite. I will put private on all those proxy class in the JS-API .

adinapitul: @eromano , please re-visit ADF-3471. There should be a single method deleteFavorite, since the exact same endpoint works for files, folders and sites.

pionnegru: I guess https://github.com/Alfresco/alfresco-js-api/blob/development/src/api-legacy/content-rest-api/src/api/favoritesApi.ts#L80 should proxy https://github.com/Alfresco/alfresco-js-api/blob/development/src/api/content-rest-api/api/favorites.api.ts#L273

eromano:
closed and solved in the js-api it was a wrong proxy in the JS-API legacy layer thanks

PR #4214, opened 2019-01-27
[ADF-4125] simplify extension load in extension module

eromano:

> - [ ] Bugfix
> - [ ] Feature
> - [x] Build related changes
> - [ ] Documentation

Simplify extension load in extension module. 
As we do in the core module where we initialize the app.config.json I think will be usfull load the extension config in the module declaration. The advantages of this approach are:

1. A common way to load extension
2. No need of extra code in application
3. A simpler way to add the extension capability in an app
4. All the apps can contribute in update or fix bugs in this necessary piece of code

> - [ ] Yes
> - [x] No

**Other information**:


lib/extensions/src/lib/extensions.module.ts
```diff
@@ -15,10 +15,15 @@
...
  */
 
-import { NgModule } from '@angular/core';
 import { DynamicExtensionComponent } from './components/dynamic-component/dynamic.component';
 import { DynamicTabComponent } from './components/dynamic-tab/dynamic-tab.component';
 import { DynamicColumnComponent } from './components/dynamic-column/dynamic-column.component';
+import { NgModule, ModuleWithProviders, APP_INITIALIZER } from '@angular/core';
+import { AppExtensionService } from './services/app-extension.service';
+
+export function setupExtensions(service: AppExtensionService): Function {
```
DenysVuika: this will be a bad thing, as it prevents applications controlling the load process. Here's example what will break if you merge this PR: https://github.com/Alfresco/alfresco-content-app/blob/development/src/app/extensions/extension.service.ts#L130

lib/extensions/src/lib/services/app-extension.service.ts
```diff
@@ -31,22 +29,22 @@ export class AppExtensionService {
 
     references$: Observable<ExtensionRef[]>;
 
-    constructor(private extensions: ExtensionService) {
+    constructor(private extensionService: ExtensionService, private logService: LogService) {
```
DenysVuika: The idea was that Core will reference Extensions once we start moving stuff, not the Extensions referencing Core. Otherwise, we have to carry the huge Core dependency on extension projects.

DenysVuika:
Can we consider this PR for 3.1, we don't have time reworking ACA for this refactoring :(

lib/extensions/src/lib/extensions.module.ts
eromano: Why the application should not control the load process. extension.service.ts what I see in this class of ACA. I guess need to be moved in any single component that exposes this extensibility.
In fact in this moment this service has inside too many and different responsibility.

eromano:
> Can we consider this PR for 3.1, we don't have time reworking ACA for this refactoring :(

At the moment is really difficult to start a project that uses extension because of that classes and modules that are the same all the time. We have to make easy the use of extensions. If it requires more time we can talk about it in our usual standup

DenysVuika:
Do you plan having a fix for the ACA, because .forRoot will break the code. All guides and documentation examples using that ExtensionModule will need to be changed now.

lib/extensions/src/lib/extensions.module.ts
```diff
@@ -15,10 +15,15 @@
  * limitations under the License.
  */
 
-import { NgModule } from '@angular/core';
 import { DynamicExtensionComponent } from './components/dynamic-component/dynamic.component';
 import { DynamicTabComponent } from './components/dynamic-tab/dynamic-tab.component';
 import { DynamicColumnComponent } from './components/dynamic-column/dynamic-column.component';
+import { NgModule, ModuleWithProviders, APP_INITIALIZER } from '@angular/core';
+import { AppExtensionService } from './services/app-extension.service';
```
DenysVuika: Renaming this service will break many stuff partners build for us, as they import and use it as ExtensionService, the App prefix was for application level and demo shell only.

PR #4299, opened 2019-02-11
[ADF-4067] Process Service Cloud change applicationName input

SilviuCPopa:

> - [ ] Bugfix
> - [x] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://issues.alfresco.com/jira/browse/ADF-4067

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/demo-shell/src/app/components/app-layout/cloud/processes-cloud-demo.component.html
+++ b/demo-shell/src/app/components/app-layout/cloud/processes-cloud-demo.component.html
@@ -1,6 +1,6 @@
 <div fxLayout="column" fxFill fxLayoutGap="2px">
     <adf-cloud-edit-process-filter
-        [appName]="applicationName"
+        [appName]="appName"
         [id]="filterId"
         [filterProperties]="processFilterProperties"
         (filterChange)="onFilterChange($event)"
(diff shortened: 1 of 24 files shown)


demo-shell/src/app/components/app-layout/cloud/processes-cloud-demo.component.html
```diff
@@ -8,7 +8,7 @@
     </adf-cloud-edit-process-filter>
     <div fxLayout="column" fxFlex fxLayoutAlign="space-between" *ngIf="editedFilter">
         <adf-cloud-process-list fxFlex class="adf-cloud-layout-overflow"
-                                [applicationName]="editedFilter.appName"
+                                [appName]="editedFilter.appName"
```
davidcanonieto: Could we replace the local applicationName variable to appName so it matches the input name? This way if it's called the same in every single place it'll make it easier to spot and understand what it's supposed to be. Take a look at the appId which is also the pk for APS1 components. It's the same in app.routes and everywhere else (expect for to places that should change as well). This way we can avoid things like  [appName]="applicationName" and be more consistent in our code. Thank you.

PR #4302, opened 2019-02-12
[ADF-4069] Added updated license list tool

therealandeeee:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [x] Documentation

Added a new license list script that puts the output from license-checker through an EJS template to generate the Markdown page. The script takes a command line parameter which is just the version string for the filename and title, etc.

Issues to think about:

- Do we want just links to the licence pages or is it better to have a copy of each licence text in the same Markdown file with the list?
- There is a lookup to replace non-standard licence strings. Should we aim to make pull requests to libraries we use to fix non-standard strings? If there is any doubt about which licence is used then should we link to the project's LICENSE.md page or just not have a link at all?

> - [ ] Yes
> - [x] No

### DIFF
--- a/docs/abn-tree.yml
+++ b/docs/abn-tree.yml
@@ -24,3 +24,4 @@
 - versionIndex.md: 'Version index'
 - compatibility.md: 'Version compatibility'
 - release-notes: 'Release notes'
+- license-info: 'License info'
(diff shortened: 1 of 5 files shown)


fcorti:
Looks good.
How the license-info-v3.0.0.md file is visible into the Builder Network?

therealandeeee:
> Looks good.
> How the license-info-v3.0.0.md file is visible into the Builder Network?

If we decide to go with this format then we can add it to the abn-tree.yml file and it will come up as an item in the sidebar.

fcorti:
Sounds perfect to me.
Thanks.

PR #4322, opened 2019-02-15
[ADF-4083] search facetFields - empty spaced labels support

suzanadirla:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation
There was a API limitation so that the response was not correct for the facet fields that had labels with spaces inside them. This problem is now fixed.
The filter component is now changed by this PR so to support such labels for facet fields.

Facet fields cannot have spaces inside labels, because the current response parser cannot fetch the response https://issues.alfresco.com/jira/browse/ADF-4083

Facet fields can now have labels with spaces. In order for the API to return the correct response, such labels need to be wrapped inside quotes.

> - [ ] Yes
> - [x] No

**Other information**:


docs/content-services/search-filter.component.md
```diff
@@ -66,7 +66,7 @@ A typical configuration is shown below:
           "expanded": true,
           "fields": [
             { "field": "content.mimetype", "mincount": 1, "label": "Type" },
-            { "field": "content.size", "mincount": 1, "label": "Size" },
+            { "field": "content.size", "mincount": 1, "label": "\"File Size\"" },
```
DenysVuika: This example is inconsistent with the one below that has single quotes.

docs/content-services/search-filter.component.md
```diff
@@ -277,7 +280,7 @@ the next block of results.
 
 | Name | Type | Default | Description |
 | ---- | ---- | ------- | ----------- |
-| field | string |  | Specifies the facet field. |
+| field | string |  | Specifies the facet field. If the field string contains spaces, it must be escaped by wrapping it inside quotes - in order to get the correct response from the server.|
```
DenysVuika: Can we do this in the code when we build query? checking for spaces and wrapping the code so that developers don't need to remember this thing?

lib/content-services/search/search-query-builder.service.ts
```diff
@@ -384,4 +386,12 @@ export class SearchQueryBuilderService {
 
         return null;
     }
+
+    getSupportedLabel(item: string) {
```
eromano: missing return type, is the item the right name for this input parameter?

suzanadirla: changed to a more suggestive name

lib/content-services/search/components/search-filter/search-filter.component.ts
```diff
@@ -192,7 +192,7 @@ export class SearchFilterComponent implements OnInit, OnDestroy {
     private parseFacetItems(context: ResultSetContext, configFacetFields, itemType): FacetField[] {
         return configFacetFields.map((field) => {
```
eromano: field doesn't look the right name. after you do field.field. Maybe currentConfigFacetField could be a better name to avoid to have field.field

PR #4426, opened 2019-03-12
[AD-4222] e2e compilation fixes

gmandakini:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

Fixing the compilation errors, so that the basic compile is run and in place, before hte e2e tests are run on Travis.

> - [ ] Yes
> - [ ] No

**Other information**:
https://issues.alfresco.com/jira/browse/ADF-4222


e2e/pages/adf/dialog/shareDialog.ts
```diff
@@ -111,7 +111,7 @@ export class ShareDialog {
     }
 
     calendarTodayDayIsDisabled() {
-        const today = this.dayPicker.element(by.css('.mat-datetimepicker-calendar-body-today')).getText();
+        const today = this.dayPicker.element(by.css('.mat-datetimepicker-calendar-body-today')).getText().toString();
```
adinapitul: I think element(...).getText() returns promise.Promise<string>. toString() is incorrectly used in this case.

e2e/process-services/form-widgets-component.e2e.ts
```diff
@@ -100,63 +100,63 @@ describe('Form widgets', () => {
                     return formInstance;
                 })
                 .then(() => {
-                    expect(taskPage.formFields().getFieldLabel(appFields.text_id))
+                    expect(taskPage.formFields().getFieldLabel(appFields.text_id, undefined))
```
adinapitul: wouldn't it be more elegant to have an optional parameter instead of using "undefined" ?

e2e/core/user-info-component-cloud.e2e.ts
```diff
@@ -29,7 +29,7 @@ describe('User Info - SSO', () => {
     const loginSSOPage = new LoginSSOPage();
     const navigationBarPage = new NavigationBarPage();
     const userInfoPage = new UserInfoPage();
-    const identityService: Identity = new Identity();
+    const identityService: Identity = new Identity(undefined);
```
cristinaj: Shouldn't we have new Identity() instead of new Identity(undefined)?

gmandakini: undefined is added for the tests to compile. as the constructor expects a parameter here.

e2e/process-services/form-widgets-component.e2e.ts
gmandakini: I googled and came across this:
Optional and Default Parameters. In TypeScript, every parameter is assumed to be required by the function. This doesn't mean that it can't be given null or undefined , but rather, when the function is called the compiler will check that the user has provided a value for each parameter.
is it possible to add an optional parameter in typescript?

adinapitul: Assuming you got this info from here: https://www.typescriptlang.org/docs/handbook/functions.html
The answer to your question is right below the fragment you copied:
`We can get this functionality in TypeScript by adding a ? to the end of parameters we want to be optional`

gmandakini: great, will try that.

e2e/core/user-info-component-cloud.e2e.ts
cristinaj: Maybe we can have also a constructor without params?

e2e/pages/adf/dialog/shareDialog.ts
gmandakini: what would you suggest, in terms of usage here? can you please elaborate.

e2e/core/user-info-component-cloud.e2e.ts
gmandakini: will check.

e2e/pages/adf/dialog/shareDialog.ts
adinapitul: why did you add .toString() ? what was the problem you were trying to solve?

e2e/pages/adf/filePreviewPage.ts
```diff
@@ -286,15 +285,14 @@ export class FilePreviewPage {
...
+        this.checkCanvasWidth().then((width) => {
             zoomedOutWidth = width;
             if (actualWidth && zoomedOutWidth) {
                 expect(zoomedOutWidth).toBeLessThan(actualWidth);
             }
         });
 
-        this.checkCanvasHeight().then(function (height) {
-            zoomedInHeight = height;
+        this.checkCanvasHeight().then(() => {
```
suzanadirla: I'm pretty sure the fix in this case (`'zoomedInHeight' is declared but its value is never read.`) should have been just to replace the `zoomedInHeight = height;` with `zoomedOutHeight = height;`

PR #4483, opened 2019-03-22
Exporting base login page

VitoAlbano:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

Exporting the base login page to allow other projects to import it from adf-testing package

> - [ ] Yes
> - [ ] No

**Other information**:


lib/testing/src/lib/core/pages/login.page.ts
```diff
@@ -15,32 +15,47 @@
...
+import { SettingsPage } from './settings.page';
 
 export class LoginPage {
-
     formControllersPage = new FormControllersPage();
     txtUsername = element(by.css('input[id="username"]'));
     txtPassword = element(by.css('input[id="password"]'));
     logoImg = element(by.css('img[id="adf-login-img-logo"]'));
-    successRouteTxt = element(by.css('input[data-automation-id="adf-success-route"]'));
+    successRouteTxt = element(
```
adinapitul: demo-shell specific, should not be in the generic page object

lib/testing/src/lib/core/pages/login.page.ts
```diff
@@ -15,32 +15,47 @@
...
+import { FormControllersPage } from './form-controller.page';
 import { browser, by, element, protractor } from 'protractor';
-import TestConfig = require('../../test.config');
-import { BrowserVisibility } from '@alfresco/adf-testing';
+import { BrowserVisibility } from '../browser-visibility';
+import { SettingsPage } from './settings.page';
 
 export class LoginPage {
-
     formControllersPage = new FormControllersPage();
```
adinapitul: is an entire separate class really needed only for clicking on a toggle?

lib/testing/src/lib/core/pages/login.page.ts
```diff
@@ -15,32 +15,47 @@
...
     txtUsername = element(by.css('input[id="username"]'));
     txtPassword = element(by.css('input[id="password"]'));
     logoImg = element(by.css('img[id="adf-login-img-logo"]'));
-    successRouteTxt = element(by.css('input[data-automation-id="adf-success-route"]'));
+    successRouteTxt = element(
+        by.css('input[data-automation-id="adf-success-route"]')
+    );
     logoTxt = element(by.css('input[data-automation-id="adf-url-logo"]'));
```
adinapitul: demo-shell specific, should not be in the generic page object

lib/testing/src/lib/core/pages/login.page.ts
```diff
@@ -50,7 +65,12 @@ export class LoginPage {
     logoSwitch = element(by.id('adf-toggle-logo'));
```
adinapitul: demo-shell specific, should not be in the generic page object

lib/testing/src/lib/core/pages/login.page.ts
```diff
@@ -50,7 +65,12 @@ export class LoginPage {
...
     header = element(by.id('adf-header'));
```
adinapitul: demo-shell specific, should not be in the generic page object

lib/testing/src/lib/core/pages/login.page.ts
```diff
@@ -50,7 +65,12 @@ export class LoginPage {
...
     settingsPage = new SettingsPage();
-    settingsIcon = element(by.cssContainingText('a[data-automation-id="settings"] mat-icon', 'settings'));
+    settingsIcon = element(
```
adinapitul: demo-shell specific, should not be in the generic page object

lib/testing/src/lib/core/pages/login.page.ts
```diff
@@ -163,7 +183,7 @@ export class LoginPage {
 
     goToLoginPage() {
```
adinapitul: browser.driver.get is to be used when the web page doesn't have Angular components, but doesn't really make sense after calling browser.waitForAngularEnabled(true)

lib/testing/src/lib/core/pages/login.page.ts
adinapitul: besides the other ones I added comments to, the following are also demo-shell specific and should not be in the generic page object:
footerSwitch
rememberMeSwitch
successRouteSwitch

lib/testing/src/lib/core/pages/login.page.ts
```diff
@@ -50,7 +65,12 @@ export class LoginPage {
...
     settingsPage = new SettingsPage();
-    settingsIcon = element(by.cssContainingText('a[data-automation-id="settings"] mat-icon', 'settings'));
+    settingsIcon = element(
+        by.cssContainingText(
+            'a[data-automation-id="settings"] mat-icon',
+            'settings'
+        )
+    );
 
     waitForElements() {
```
adinapitul: can this be changed to return the proper promise ?

lib/testing/src/lib/core/pages/login.page.ts
adinapitul: methods should be changed to return the proper promise instead of returning "this"
for example clearUsername

lib/testing/src/lib/core/pages/login.page.ts
```diff
@@ -183,7 +183,7 @@ export class LoginPage {
 
     goToLoginPage() {
```
adinapitul: checkRegisterDisplayed() is a method that in case of failure gives a timeout

I would suggest refactoring such methods into something like isRegisterDisplayed() returning a boolean promise. This can then be used in the tests inside expects, and in case of failure this will avoid the timeout

adinapitul: enableFooter, disableFooter and others are methods specific to demo-shell, not needed in the generic page object

lib/testing/src/lib/core/pages/login.page.ts
```diff
@@ -163,7 +183,7 @@ export class LoginPage {
...
         browser.waitForAngularEnabled(true);
-        browser.driver.get(TestConfig.adf.url + TestConfig.adf.port + '/login');
+        browser.driver.get(browser.baseUrl + '/login');
```
adinapitul: in order for this to be generic, this should also consider the case when # is used (like in the case of ACA)

eromano:
@adinapitul most of them seem to e good suggestion please let's create a different task to refactor this page. This task, for now, was focus only on the move. @VitoAlbano already was thinking on refactoring it will be better to create a task on the board for it. 
Plus I will create a meeting to gather all the idea around this new package tomorrow

PR #4539, opened 2019-04-02
[ADF-3887] Different local storages for different ADF apps

davidcanonieto:

> - [ ] Bugfix
> - [x] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://issues.alfresco.com/jira/browse/ADF-3887
Different ADF apps were sharing the same local storage so the changes in those app settings would affect all of them

Each app has its own local storage so changes are only applied to the specific app.
After login user preferences are stored in this local storage.
Clean up Alfresco Api Service

> - [ ] Yes
> - [x] No

**Other information**:
https://issues.alfresco.com/jira/browse/ADF-3887

### DIFF
--- a/lib/core/services/storage.service.spec.ts
+++ b/lib/core/services/storage.service.spec.ts
@@ -26,6 +26,8 @@ describe('StorageService', () => {
 
     let storage: StorageService;
     let appConfig: AppConfigServiceMock;
+    const key = 'test_key';
+    const value = 'test_value';
 
     setupTestBed({
         imports: [CoreTestingModule],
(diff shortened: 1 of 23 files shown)


lib/core/services/storage.service.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+    });
+
+    it('should get the prefix for the storage from app config', (done) => {
+        appConfig.load().then(() => {
+            expect(storage.storagePrefix).toBe('ADF_APP_');
+            done();
+        });
+    });
+
+    it('should set an empty prefix when the it is not defined in the app config', (done) => {
```
eromano: can you try to set and after get and element with and without the prefix? and can you please also rease te PR?

PR #4546, opened 2019-04-03
[ADF-4323] Add style fixes from ACA

suzanadirla:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation
Style improvements brought from ACA style-override files. https://issues.alfresco.com/jira/browse/ADF-4323

The UI fixes that were done on ACA side will be available on ADF components by default, allowing clean-up of code on ACA side and possibly on other projects as well.
Here is a list of the components that got the style fixes from ACA:
- File uploading dialog - set the z-index
- File uploading list row
- Info drawer layout component
- Pagination - on handsets
- Search Filter
- Search Sorting Picker
- Sidenav Layout
- Toolbar
- Upload drag area component
- Version manager - fixes long comments display issue

> - [ ] Yes
> - [x] No

**Other information**:


lib/core/layout/components/sidenav-layout/sidenav-layout.component.ts
```diff
@@ -25,7 +37,8 @@ import { BehaviorSubject, Observable } from 'rxjs';
 @Component({
     selector: 'adf-sidenav-layout',
     templateUrl: './sidenav-layout.component.html',
-    styleUrls: ['./sidenav-layout.component.scss']
+    styleUrls: ['./sidenav-layout.component.scss'],
+    encapsulation: ViewEncapsulation.None
```
DenysVuika: Can we please also add host element class like all Material components have?

```ts
host: { class: 'adf-sidenav-layout' }
```

Will be much eaiser to style from outside apps.

DenysVuika:
There's a css lint issue,

```
lib/core/toolbar/toolbar.component.scss
 34:2  ✖  Expected class selector ".inline" to match specified pattern   selector-class-pattern
```

I think the class needs to be renamed to something like `adf-toolbar-inline`

suzanadirla:
Thanks, Denys, I'll make the requested changes

DenysVuika:
@suzanadirla in the PR details, please provide a list of ADF components to be verified after this PR is merged.

demo-shell/src/app/components/files/files.component.scss
```diff
@@ -76,28 +76,10 @@
 
     .adf-manage-versions-sidebar {
         width: 360px !important;
-        color: rgba(0, 0, 0, 0.87);
```
eromano: Why do we need to remove this part?

lib/core/pagination/pagination.component.scss
```diff
@@ -96,6 +77,10 @@
             max-height: 250px !important;
         }
 
+        &.adf-pagination__empty {
```
eromano: do we need it? hasItems in the template already hide it

eromano:
can you also please rebase your PR i see one test in PS that is not failing in development at the moment

demo-shell/src/app/components/files/files.component.scss
suzanadirla: :) I tested how it looks after I remove it and there is no difference with or without the 'color: rgba(0, 0, 0, 0.87);' line.
Maybe it would be good to have a task to clean-up the demo-shell scss files and also move styles from there to components...

lib/core/pagination/pagination.component.scss
suzanadirla: That's what I also thought in the beginning, but... if we do not set 'display:none;', then the pagination is still displayed with its border-top: 1px solid rgba(0, 0, 0, 0.07) having height: 48px;
On demo-shell is not noticeable, but on ACA it looks really ugly.

demo-shell/src/app/components/files/files.component.scss
DenysVuika: @eromano it's a demo shell file, always nice to clean it a bit as for me :)

lib/core/pagination/pagination.component.scss
DenysVuika: @eromano @suzanadirla maybe we take a chance fixing this in the ADF right now, and have proper pagination hiding on empty?

suzanadirla: I tried to find a better way, but I do not see much options: 
As solution no 2, we could add instead a wrapping div that would have the styles applied to it, which would be hidden along with any border, height, spacings applied to it - and the component tag would remain completly unstyled.

What ever is the chosen solution, we should make sure that all components that have all content inside `<ng-container *ngIf="` element would get the empty state fixed - so no spacing or borders/background would still be displayed when they should not.

suzanadirla: After talking with Denys, I've changed the styling on the empty pagination.
I also added 2 variables so the styling for the empty state can be changed (on ACA side, for example).
@eromano, please check. Hope it's better now.

PR #4608, opened 2019-04-15
[ADF-4406] - Confirm Dialog doesn't support a third extra button option to be customised

dhrn:

> - [ ] Bugfix
> - [x] Feature
> - [ ] Build related changes
> - [ ] Documentation

yes-all

> - [ ] Yes
> - [x] No

**Other information**:


lib/content-services/dialogs/confirm.dialog.ts
```diff
@@ -31,19 +32,28 @@ export class ConfirmDialogComponent {
     title: string;
     message: string;
     yesLabel: string;
+    thirdOption: string;
```
eromano: yesLabel: string;
thirdOption: string;
noLabel: string;	    noLabel: string;

I see we are using a kind of convention label as a suffix. Maybe is better to keep it?
ThirdButtonLabel, ThirdOptionLabel. The other question that I think we should at this point ask to ourself do we need a more generic way to add buttons?
Otherwise, next time we need a 4th buttons we are going to add another property and so on

eromano:
yes thanks

PR #4615, opened 2019-04-16
[ADF-4401] Fix DL issues because of nested 'adf-datatable-cell' items

suzanadirla:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

Please check https://issues.alfresco.com/jira/browse/ADF-4401

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/core/datatable/components/datatable/datatable.component.scss
+++ b/lib/core/datatable/components/datatable/datatable.component.scss
@@ -409,7 +409,7 @@
             white-space: nowrap;
 
             &.adf-datatable-cell-header,
-            .adf-content-cell {
+            .adf-datatable-content-cell {
                 max-width: calc(100% - 0.1px);
                 overflow: hidden;
                 text-overflow: ellipsis;
@@ -419,7 +419,7 @@
                     text-overflow: ellipsis;
                 }
             }
-            .adf-content-cell {
+            .adf-datatable-content-cell {
                 position: absolute;
             }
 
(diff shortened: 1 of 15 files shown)


lib/core/datatable/components/datatable/datatable.component.scss
```diff
@@ -402,23 +408,20 @@
...
                     overflow: hidden;
                     text-overflow: ellipsis;
-                    white-space: nowrap;
-                    display: block;
-                    width: calc(100% - 2em);
-                    position: absolute;
-                    // margin-top: -10px;
                 }
             }
+            .adf-content-cell {
```
eromano: maybe we should to keep to use the same prefix adf-datatable

lib/core/datatable/components/datatable/datatable.component.scss
```diff
@@ -402,23 +408,20 @@
             text-overflow: ellipsis;
             white-space: nowrap;
 
-            .adf-datatable-cell, .adf-datatable-cell-header {
+            &.adf-datatable-cell-header,
```
davidcanonieto: Can we double check that these changes keep the usability and the accessibility for the cells.
In this PR the cell is not clickable on its entire area, only the texts. The borders need to be applied to the whole area of the cell and not only the text.

suzanadirla: I've noticed that behavior too, but on development, that is why I thought that was expected...

PR #4622, opened 2019-04-18
[ADF-4424] Pdf Viewer - password dialog does not close when pressing Escape

pionnegru:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

[ADF-4424](https://issues.alfresco.com/jira/browse/ADF-4424)

- closing the dialog without providing a password will return to the location from where the file was opened

> - [ ] Yes
> - [x] No

**Other information**:


lib/core/viewer/components/pdfViewer.component.ts
```diff
@@ -436,13 +438,14 @@ export class PdfViewerComponent implements OnChanges, OnDestroy {
...
                 data: { reason }
             })
             .afterClosed().subscribe((password) => {
-            if (password) {
-                callback(password);
-            }
+                if (password) {
+                    callback(password);
+                } else {
+                    this.location.back();
```
eromano: @davidcanonieto has introduced a new service to go back in a better way can we use it https://github.com/Alfresco/alfresco-ng2-components/blob/development/lib/core/viewer/components/viewer.component.ts   ?  private previousRouteService: PreviousRouteService)

pionnegru: doesn't seem to work.

eromano: @davidcanonieto can you please check with @pionnegru if is possible to reuse your service or it needs some improvement?

lib/core/services/previous-route.service.ts
```diff
@@ -39,4 +41,16 @@ export class PreviousRouteService {
...
+    public goBackToPreviousPage(): void {
+        if (this.previousUrl && this.previousUrl.includes('login') || window.history.length <= 2) {
+            this.goBackToHome();
+        } else {
+            this.location.back();
+        }
+    }
+
+    public goBackToHome(): void {
+        this.router.navigate([{outlets: {overlay: null, primary: ['home']}}]);
```
pionnegru: what is 'home' from the perspective of another app?

pionnegru: also 'login' route might not be also

DenysVuika: @davidcanonieto yeah, for other apps outside demo shell there is no "home" route :)

pionnegru: as suggested previously, navigation commands can give some details regarding location/overlay
```typescript
        const tree = this.router.parseUrl(this.router.url);
        const primaryGroup = tree.root.children[PRIMARY_OUTLET];
        const overlaySegmentGroup = tree.root.children['overlay'];

        console.log(overlaySegmentGroup.toString()); overlay -> /files/something/view
        console.log(primaryGroup.toString()); original location -> // files
```

eromano: I agree need some love to make it generic can you please collaborate with David in order to improve it?

pionnegru: I think the same problem can be found here 
https://github.com/Alfresco/alfresco-ng2-components/blob/development/lib/core/viewer/components/viewer.component.ts#L491

eromano: yes is the same. for the login route check redirect we should use  :

```
 protected getLoginRoute(): string {
        return (
            this.appConfigService &&
            this.appConfigService.get<string>(
                AppConfigValues.LOGIN_ROUTE,
                'login'
            )
        );
    }
```

For home, we could define a similar strategy in the appconfig or we could just redirect the user to the '/'

DenysVuika: Can we just pass the url as the route.data ? @pionnegru @eromano

pionnegru: From my perspective, we are just trying to close the overlay and we can avoid checking url for login or do other logic if we are just navigate to primary outlet of the overlay.

DenysVuika: I think this PR is turning into a mess. All we need is to close a password dialog upon pressing ESC.

eromano: @DenysVuika why is a mess? I removed the routing from inside do you think is a bad idea?

DenysVuika: @eromano no, that's a good idea :)

pionnegru: with the latest commits the overlay is still there and you cannot navigate documentlist. are there any commits to this?

DenysVuika: @pionnegru you might be looking at the outdated changes? I checked the code and there's no router there.

pionnegru:
I test with lates commits and when the overlay viewer is closed in url you can still see /files(overlay:files/.../view) and you cannot interact with anything on the page

eromano:
@pionnegru , Yep i see but should be it handle by the app, maybe listening on some close event?

pionnegru:
Ok, I see now. We now only deal when its overlay/in document view but not as a secondary outlet

DenysVuika:
What is the status of the PR?

pionnegru:
the only tests failing are  process-services and process-services-cloud. do we expect this?

PR #4668, opened 2019-04-29
[ADF-4232] Fix filePreviewPage methods to reach expectations

davidcanonieto:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://issues.alfresco.com/jira/browse/ADF-4232
The methods in the filePreviewPage never reach the expect conditions

The filePreviewPage has been removed as it mostly had duplicated code that already existed on viewerPage. The few parts that were different were added to the viewerPage and the expects were moved inside the viewer-content-services-component.e2e.ts

> - [ ] Yes
> - [x] No

**Other information**:
https://issues.alfresco.com/jira/browse/ADF-4232


e2e/pages/adf/filePreviewPage.ts
```diff
@@ -257,10 +256,9 @@ export class FilePreviewPage {
         });
     }
 
-    /*
     zoomOut() {
```
suzanadirla: I cannot find any place where this method is being called... The expect inside it is not reached because of that. The same case is for `zoomIn` and `actualSize` methods.

e2e/pages/adf/filePreviewPage.ts
```diff
@@ -89,16 +89,10 @@ export class FilePreviewPage {
 
         this.checkCanvasWidth().then((width) => {
```
cristinaj: checkCanvasWidth() method should be renamed to getCanvasWidth(), since it's not checking anything, just returning a value. Same for checkCanvasHeight().

e2e/pages/adf/filePreviewPage.ts
```diff
@@ -89,16 +89,10 @@ export class FilePreviewPage {
...
             actualWidth = width;
```
cristinaj: If the expect was removed, i don't think actualWidth is still useful. From my point of view, this zoomIn() method has no sense. I would instead do these steps in test. Add an expect in the test related to width and height in case it's necessary, then clickZoomIn(), then add other expects in test.

suzanadirla: the 'actualWidth' is still needed, but I do understand your point that the 'expect' should better be inside the 'it(...)' from the actual test file...

e2e/core/viewer/viewer-content-services-component.e2e.ts
```diff
@@ -28,8 +26,10 @@ import { AcsUserModel } from '../../models/ACS/acsUserModel';
 
 import { AlfrescoApiCompatibility as AlfrescoApi } from '@alfresco/js-api';
 import { UploadActions } from '../../actions/ACS/upload.actions';
+import { browser } from 'protractor';
 
 describe('Content Services Viewer', () => {
+    jasmine.DEFAULT_TIMEOUT_INTERVAL = 50000;
```
adinapitul: AFAIK, this will change the global jasmine timeout.
if these tests require q different timeout, either bigger or smaller, the recommended solution is to same the original timeout, change the value, run the tests, then in afterAll change the timeout to the original value

PR #4701, opened 2019-05-07
[ADF-4498] Disable Sticky header on Datatable if header is not visible

davidcanonieto:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://issues.alfresco.com/jira/browse/ADF-4498
Sticky header is enabled even when there's no header in templates such as no content, no permissions and loading.

Sticky header should be enable only when the header is visible. Otherwise the component should not implement this class in its layout.

> - [ ] Yes
> - [x] No

**Other information**:
https://issues.alfresco.com/jira/browse/ADF-4498


lib/core/datatable/components/datatable/datatable.component.html
```diff
@@ -3,7 +3,7 @@
     *ngIf="data" class="adf-full-width"
     [class.adf-datatable-card]="display === 'gallery'"
     [class.adf-datatable-list]="display === 'list'"
-    [class.adf-sticky-header]="stickyHeader"
+    [class.adf-sticky-header]="stickyHeader && isHeaderVisible()"
```
DenysVuika: maybe makes more sense adding the "stickyHeader" check inside the "isHeaderVisible()" ? to have only one condition check?

PR #4712, opened 2019-05-09
[ADF-4457] StorageService should be independent of AppConfigService

davidcanonieto:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://issues.alfresco.com/jira/browse/ADF-4457
StorageService depends on AppConfigService to work

StorageService does not depend on the AppConfigService anymore and the storage
prefix is set by the app initializer.
IdentityService now makes use of the StorageService

> - [ ] Yes
> - [x] No

https://issues.alfresco.com/jira/browse/ADF-4457

**Other information**:

### DIFF
--- a/lib/core/services/alfresco-api.service.ts
+++ b/lib/core/services/alfresco-api.service.ts
@@ -27,6 +27,7 @@ import { AlfrescoApiCompatibility, AlfrescoApiConfig } from '@alfresco/js-api';
 import { AppConfigService, AppConfigValues } from '../app-config/app-config.service';
 import { Subject } from 'rxjs';
 import { OauthConfigModel } from '../models/oauth-config.model';
+import { StorageService } from './storage.service';
 
 /* tslint:disable:adf-file-name */
 
@@ -95,7 +96,9 @@ export class AlfrescoApiService {
         return this.getInstance().core.groupsApi;
     }
 
-    constructor(protected appConfig: AppConfigService) {
+    constructor(
+        protected appConfig: AppConfigService,
+        protected storageService: StorageService) {
     }
 
     async load() {
@@ -115,6 +118,8 @@ export class AlfrescoApiService {
             oauth.redirectUriLogout = window.location.origin + (oauth.redirectUriLogout || '/');
         }
 
+        this.storageService.storagePrefix = this.appConfig.get<string>(AppConfigValues.STORAGE_PREFIX, '');
+
         const config = new AlfrescoApiConfig({
             provider: this.appConfig.get<string>(AppConfigValues.PROVIDERS),
             hostEcm: this.appConfig.get<string>(AppConfigValues.ECMHOST),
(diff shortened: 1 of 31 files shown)


lib/core/services/storage.service.ts
```diff
@@ -25,11 +24,18 @@ export class StorageService {
 
     private memoryStore: { [key: string]: any } = {};
     private useLocalStorage: boolean = false;
-    storagePrefix: string;
+    private _storagePrefix: string = '';
```
popovicsandras: This is the storage service. I think simply _prefix would be enough, since having `storageWhatever` in storage service is redundant, we could simply say `whatever`.

lib/core/services/storage.service.ts
```diff
@@ -25,11 +24,18 @@ export class StorageService {
...
 
-    constructor(private appConfigService: AppConfigService) {
+    get storagePrefix() {
+        return this._storagePrefix;
+    }
+
+    set storagePrefix(prefix: string) {
```
popovicsandras: The same way: `set prefix(...)` ?

lib/core/services/storage.service.ts
```diff
@@ -25,11 +24,18 @@ export class StorageService {
 
     private memoryStore: { [key: string]: any } = {};
     private useLocalStorage: boolean = false;
-    storagePrefix: string;
+    private _storagePrefix: string = '';
 
-    constructor(private appConfigService: AppConfigService) {
+    get storagePrefix() {
```
popovicsandras: `get prefix()` ?

lib/core/services/user-preferences.service.ts
```diff
@@ -67,6 +68,10 @@ export class UserPreferencesService {
         }
     }
 
+    private initStoragePrefix() {
+        this.storage.storagePrefix = this.appConfig.get<string>('application.storagePrefix', '');
```
popovicsandras: This is something I don't really understand. We've set it already in the AlfrescoApiService. Why do we need to set it again? I assume, that the point here would be to set in only once, inside the initial bootstrapping.

lib/core/services/alfresco-api.service.ts
```diff
@@ -115,6 +118,8 @@ export class AlfrescoApiService {
             oauth.redirectUriLogout = window.location.origin + (oauth.redirectUriLogout || '/');
         }
 
+        this.storageService.storagePrefix = this.appConfig.get<string>(AppConfigValues.STORAGE_PREFIX, '');
```
popovicsandras: I think this would be better outside of this function. The method's name says initAlfrescoApi, which it does indeed. However now this method has a sideeffect, it initialises the Storage Service also. What do you think about putting this initialisation into the `load` method, just before initialising the alfresco-api?

lib/core/services/user-preferences.service.ts
popovicsandras: This is something I don't really understand. We've set it already in the AlfrescoApiService. Why do we need to set it again? I assume, that the point here would be to set in only once, inside the initial bootstrapping.

PR #4716, opened 2019-05-09
[ADF-4227] Material Dialog - change direction option based on textOrientation

pionnegru:

> - [ ] Bugfix
> - [x] Feature
> - [ ] Build related changes
> - [ ] Documentation

[ADF-4227](https://issues.alfresco.com/jira/browse/ADF-4227)

previewortl

> - [ ] Yes
> - [x] No

**Other information**:


lib/core/services/dialog-config-factory.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export function dialogConfigFactory(
+    userPreferencesService: UserPreferencesService,
+    dialogConfigService: DialogConfigService
+): any {
+    return () => {
+        return userPreferencesService
+            .select('textOrientation')
+            .subscribe((direction: Direction) => {
+                dialogConfigService.changeDirection(direction);
```
DenysVuika: Can the service listen to those events without this factory?

pionnegru: and how would it be initialized?

DenysVuika: In the service constructor?

pionnegru: I mean, how would the service be initiliezed. the constructor will work if we declare it for example as a dependency of a component. at this stage it won't

DenysVuika: ok, makes sense.

PR #4748, opened 2019-05-20
Fix concurrently command use

davidcanonieto:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

concurrently command not used correctly as it uses the global package

Command fixed by using `npx` so it does not run the global package.

> - [ ] Yes
> - [x] No

**Other information**:


package.json
```diff
@@ -6,7 +6,7 @@
   "main": "./index.js",
   "scripts": {
     "ng": "ng",
-    "concurrently": "concurrently",
+    "concurrently": "npx concurrently",
```
DenysVuika: you don't need that here, `npx` makes sense when calling from shell scripts

PR #4786, opened 2019-05-28
[ADF-4546] Visibility conditions for Form variables don't work on Multiline Widget.

monsterflaw:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

* In cloud-form visiblity fails when the visibility conditions are configured based on form variables
* The visibility conditon fails when the conditon is based form variables of date type

https://issues.alfresco.com/jira/browse/ADF-4546

* In old cloud form the variable defintion does not have id and the visibity conditions used variable name. The cloud-form visiblity condtions uss the new id property of the variables. So the variable value reading has been modified to find variables using id as well.

* If the form variable is of date type then the value was not converted to the required format and the visibility condtion based date form variable always failed. This behaviour has been fixed by converting the date form value to the appropriate type.

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/core/form/services/widget-visibility.service.ts
+++ b/lib/core/form/services/widget-visibility.service.ts
@@ -75,7 +75,7 @@ export class WidgetVisibilityService {
         }
     }
 
-    getLeftValue(form: FormModel, visibilityObj: WidgetVisibilityModel) {
+    getLeftValue(form: FormModel, visibilityObj: WidgetVisibilityModel): string {
         let leftValue = '';
         if (visibilityObj.leftRestResponseId && visibilityObj.leftRestResponseId !== 'null') {
             leftValue = this.getVariableValue(form, visibilityObj.leftRestResponseId, this.processVarList);
@@ -86,7 +86,7 @@ export class WidgetVisibilityService {
         return leftValue;
     }
 
-    getRightValue(form: FormModel, visibilityObj: WidgetVisibilityModel) {
+    getRightValue(form: FormModel, visibilityObj: WidgetVisibilityModel): string {
         let valueFound = '';
         if (visibilityObj.rightRestResponseId) {
             valueFound = this.getVariableValue(form, visibilityObj.rightRestResponseId, this.processVarList);
@@ -102,7 +102,7 @@ export class WidgetVisibilityService {
         return valueFound;
     }
 
-    getFormValue(form: FormModel, fieldId: string) {
+    getFormValue(form: FormModel, fieldId: string): any {
         let value = this.getFieldValue(form.values, fieldId);
 
         if (!value) {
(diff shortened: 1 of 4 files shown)


lib/core/form/services/widget-visibility.service.ts
```diff
@@ -186,10 +186,19 @@ export class WidgetVisibilityService {
             this.getProcessVariableValue(name, processVarList);
     }
 
-    private getFormVariableValue(form: FormModel, name: string) {
-        if (form.json.variables) {
-            const formVariable = form.json.variables.find((formVar) => formVar.name === name);
-            return formVariable ? formVariable.value : formVariable;
+    private getFormVariableValue(form: FormModel, identifier: string) {
```
eromano: can you add the return type?

lib/core/form/services/widget-visibility.service.ts
```diff
@@ -186,10 +186,19 @@ export class WidgetVisibilityService {
...
+        const variables = form.json.variables || (form.json.formRepresentation && form.json.formRepresentation.formDefinition.variables);
```
eromano: can you move the condition in a method and use it directly in the if?

lib/core/form/services/widget-visibility.service.ts
```diff
@@ -186,10 +186,19 @@ export class WidgetVisibilityService {
...
+        if (variables) {
+            const formVariable = variables.find((formVar) => {
+                return formVar.name === identifier || formVar.id === identifier;
+            });
+
+            let value = formVariable ? formVariable.value : formVariable;
```
eromano: you can check if is the type date from the model imnstead to use this control with moment

lib/core/form/services/widget-visibility.service.ts
```diff
@@ -112,7 +112,7 @@ export class WidgetVisibilityService {
         return value;
     }
 
-    getFieldValue(valueList: any, fieldId: string) {
+    getFieldValue(valueList: any, fieldId: string): any {
```
mauriziovitale: why any ?

lib/core/form/services/widget-visibility.service.ts
```diff
@@ -181,22 +181,47 @@ export class WidgetVisibilityService {
...
+        const variables = this.getFormVariables(form);
+        if (variables) {
+            const formVariable = variables.find((formVar) => {
+                return formVar.name === identifier || formVar.id === identifier;
+            });
+
+            let value;
+            if (formVariable) {
+                value = formVariable.value;
+                if (formVariable.type === 'date') {
```
mauriziovitale: Instead of doing like that, why we don’t compare only the day without the time?

lib/core/form/services/widget-visibility.service.ts
monsterflaw: Just added missing return type according to the current code

PR #4812, opened 2019-06-05
[ADF-4567]Add a test for 'Should have all the fields on the task header cloud r…

cristinaj:
…eadonly when the task is unclaimed.'

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

> - [ ] Yes
> - [ ] No

**Other information**:
https://issues.alfresco.com/jira/browse/ADF-4567

### DIFF
--- a/e2e/process-services-cloud/task-header-cloud.e2e.ts
+++ b/e2e/process-services-cloud/task-header-cloud.e2e.ts
@@ -23,6 +23,7 @@ import { browser } from 'protractor';
 import { NavigationBarPage } from '../pages/adf/navigationBarPage';
 import { LoginSSOPage, AppListCloudPage, TaskHeaderCloudPage, TasksService } from '@alfresco/adf-testing';
 import { TasksCloudDemoPage } from '../pages/adf/demo-shell/process-services/tasksCloudDemoPage';
+import { TaskDetailsCloudDemoPage } from '../pages/adf/demo-shell/process-services/taskDetailsCloudDemoPage';
 import resources = require('../util/resources');
 
 describe('Task Header cloud component', () => {
@@ -33,6 +34,7 @@ describe('Task Header cloud component', () => {
     let completedTask;
     let completedCreatedDate;
     let subTask;
+    let unclaimedTask;
     let subTaskCreatedDate;
     let completedEndDate;
     const simpleApp = resources.ACTIVITI7_APPS.SIMPLE_APP.name;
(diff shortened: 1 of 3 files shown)


e2e/process-services-cloud/task-header-cloud.e2e.ts
```diff
@@ -149,4 +154,19 @@ describe('Task Header cloud component', () => {
         expect(taskHeaderCloudPage.getParentTaskId())
             .toEqual(subTask.entry.parentTaskId === null ? '' : subTask.entry.parentTaskId);
     });
+
+    it('[C309699] SShould have all the fields on the task header cloud readonly when the task is unclaimed', () => {
```
gmandakini: Should we extend this test to check for all the other fields on the task header cloud as well? I realise that the bug was for these 2 fields, but will be nice to cover all the fields in the test.

lib/testing/src/lib/process-services-cloud/pages/task-header-cloud-component.page.ts
```diff
@@ -32,6 +33,8 @@ export class TaskHeaderCloudPage {
     idField = element.all(by.css('span[data-automation-id*="id"] span')).first();
     descriptionField = element(by.css('span[data-automation-id*="description"] span'));
     taskPropertyList = element(by.css('adf-cloud-task-header adf-card-view div[class="adf-property-list"]'));
+    descriptionEditIcon = element(by.css('mat-icon[data-automation-id*="edit-icon-description"]'));
```
gmandakini: I suggest we declare just 1 locator, and take the task header cloud element name dynamically. This way we can check the assertions for the all the elements on the Task Header Cloud.

lib/testing/src/lib/process-services-cloud/pages/task-header-cloud-component.page.ts
```diff
@@ -77,4 +80,20 @@ export class TaskHeaderCloudPage {
         return BrowserActions.getText(this.dueDateField);
     }
 
+    checkDescriptionEditIconIsDisplayed() {
```
gmandakini: We can just keep 1 method for isDisplayed, and 1 method for isNotDisplayed, and send the element name dynamically for assertions.

e2e/process-services-cloud/task-header-cloud.e2e.ts
cristinaj: I think it's a very good idea, i will add all fields.

lib/testing/src/lib/process-services-cloud/pages/task-header-cloud-component.page.ts
cristinaj: From my point of view, it's better to have methods for each property, because it's much more clear for devs that will create new tests which are the properties of task header component. Otherwise we will have to provide the element name from the tests, which is not ideal. In case something changes, all tests will be affected.

mauriziovitale:
Reopen it in the case is needed

PR #4823, opened 2019-06-07
[ADF-4637] logout cross tabs event

eromano:

> - [ ] Bugfix
> - [x] Feature
> - [ ] Build related changes
> - [ ] Documentation

How to reproduce:

Behavior 1

UserA logs in Tab1
UserB logs in Tab2
Current behavior the user in Tab 1 will not be able to perform any action and will be redirected to the login page or will get some error

Behavior 2

UserA logs in Tab1
UserA open another Tab2 with an ADF app
User a Tab 2 press the logout button
Current behavior the user in Tab 2 will not be able to perform any action and will be redirected to the login page only after he tries to perform some actions

Expected behavior:

Behavior 1

UserA Tab 1 get refresh and the UserB appear as current User also in the Tab 1
Behavior 2

UserA gets logged out int All the tabs too and redirects to the login page without the necessity to perform any action

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/core/services/auth-guard.service.ts
+++ b/lib/core/services/auth-guard.service.ts
@@ -35,7 +35,7 @@ export class AuthGuard extends AuthGuardBase {
         super(authenticationService, router, appConfigService);
         this.ticketChangeBind = this.ticketChange.bind(this);
 
-        window.addEventListener('storage', this.ticketChange.bind(this));
+        window.addEventListener('storage', this.ticketChangeBind);
     }
 
     ticketChange(event: StorageEvent) {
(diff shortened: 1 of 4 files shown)


lib/core/services/auth-guard.service.ts
```diff
@@ -30,6 +30,31 @@ export class AuthGuard extends AuthGuardBase {
...
+    }
+
+    private ticketChangeRedirect(event: StorageEvent, provider: string) {
+        if (!event.newValue) {
+            this.redirectToUrl(provider, this.router.url);
+        } else {
+            window.location.reload();
+        }
+
+        window.removeEventListener('storage', this.ticketChange);
```
popovicsandras: I don't think it would work properly. You added the listener: `this.ticketChange.bind(this)`, which created a new function, so if you want to remove the `this.ticketChange` it won't remove anything, since it wasn't the thing which was added. If I understand it properly.

eromano: I think you are right

PR #4870, opened 2019-06-20
[ADF-4696] Display name on metadata card

suzanadirla:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

There are some properties for which the display is wrong. Please check issue: https://issues.alfresco.com/jira/browse/ADF-4696

For the properties that have displayName data, show that data to the user.

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/core/card-view/models/card-view-baseitem.model.ts
+++ b/lib/core/card-view/models/card-view-baseitem.model.ts
@@ -30,7 +30,7 @@ export abstract class CardViewBaseItemModel {
 
     constructor(cardViewItemProperties: CardViewItemProperties) {
         this.label = cardViewItemProperties.label || '';
-        this.value = cardViewItemProperties.value;
+        this.value = cardViewItemProperties.value && cardViewItemProperties.value.displayName || cardViewItemProperties.value;
         this.key = cardViewItemProperties.key;
         this.default = cardViewItemProperties.default;
         this.editable = !!cardViewItemProperties.editable;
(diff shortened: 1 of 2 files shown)


lib/core/card-view/models/card-view-baseitem.model.ts
```diff
@@ -30,7 +30,7 @@ export abstract class CardViewBaseItemModel {
 
     constructor(cardViewItemProperties: CardViewItemProperties) {
         this.label = cardViewItemProperties.label || '';
-        this.value = cardViewItemProperties.value;
+        this.value = cardViewItemProperties.value && cardViewItemProperties.value.displayName || cardViewItemProperties.value;
```
pionnegru: if displayName is a falsy value we have the same scenario that raised this bug in the first place 
`... || cardViewItemProperties.value;`

suzanadirla: true, indeed it is not the perfect solution, but it is an improvement that fixes the mentioned bug.

pionnegru: I don't consider it an improvement since it's too specific and defaults to the same bug

suzanadirla: let's take it one step at a time: fix this specific case and I'll open bug on ADF so to find the fix of the general case.

suzanadirla: https://issues.alfresco.com/jira/browse/ADF-4700 - created.

PR #4875, opened 2019-06-21
[ADF-4471]Reload task header after claim/unclaim

cristinaj:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

> - [ ] Yes
> - [ ] No

**Other information**:
https://issues.alfresco.com/jira/browse/ADF-4471

### DIFF
--- a/demo-shell/src/app/components/cloud/task-details-cloud-demo.component.html
+++ b/demo-shell/src/app/components/cloud/task-details-cloud-demo.component.html
@@ -8,10 +8,12 @@
                 [taskId]="taskId"
                 (cancelClick)="goBack()"
                 (taskCompleted)="onTaskCompleted()"
+                (taskClaimed)="onTaskClaimed()"
+                (taskUnclaimed)="onTaskUnclaimed()"
                 (formSaved)="onFormSaved()">
             </adf-cloud-task-form>
         </div>
-        <adf-cloud-task-header fxFlex
+        <adf-cloud-task-header #taskHeader fxFlex
             [appName]="appName"
             [taskId]="taskId">
         </adf-cloud-task-header>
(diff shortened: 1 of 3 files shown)


demo-shell/src/app/components/cloud/task-details-cloud-demo.component.ts
```diff
@@ -54,6 +58,14 @@ export class TaskDetailsCloudDemoComponent {
         this.goBack();
     }
 
+    onTaskClaimed() {
+        this.taskHeader.ngOnInit();
```
eromano: You should never call Angular lifecycle method directly

demo-shell/src/app/components/cloud/task-details-cloud-demo.component.html
```diff
@@ -8,10 +8,12 @@ <h4 data-automation-id="task-details-header">Simple page to show the taskId: {{
                 [taskId]="taskId"
                 (cancelClick)="goBack()"
                 (taskCompleted)="onTaskCompleted()"
+                (taskClaimed)="onTaskClaimed()"
```
DenysVuika: Instead of introducing a method and using ViewChild, you can call that directly from the template, i.e. 
```
(taskClaimed)="taskHeader.ngOnInit()"
```

demo-shell/src/app/components/cloud/task-details-cloud-demo.component.ts
DenysVuika: If all you have to all is a single method, you can get rid of these methods and call that method directly from the html template. 

@eromano the only thing I have a feeling it's not very correct when one component calls Angular lifecycle methods of another component (there can be anything unexpected inside that OnInit). Maybe having a special public method for that? Like "taskHeader.reload()" ?

eromano: maybe is better yes

PR #4877, opened 2019-06-21
[ADF-4701] Upgrade angular libs

suzanadirla:

> - [x] Bugfix
> - [x] Feature
> - [ ] Build related changes
> - [ ] Documentation

Upgraded various angular libs to get the bug fixes and features available for version 7. 
https://issues.alfresco.com/jira/browse/ADF-4701

> - [ ] Yes
> - [x] No

**Other information**:


DenysVuika:
While v8 might be a complex update (requires also Components, Translate, DateTime Picker, etc. updated), we can have the latest v7 branch I think.

suzanadirla:
Because the upgrade to v8 implies multiple changes - some of them even breaking-changes - it would be good to first get all the fixes and features for v7 first - without any breaking-changes - and then decide when it is best to do the upgrade to v8.

Currently, on this PR, there is a unit test failing that I'm trying to fix and besides that, there should be no major impact on the code.

eromano:
There are no braking change between V7 and V8 that affect ADF. Can I know with this update which fix we are going to have?

lib/process-services-cloud/src/lib/process/start-process/components/start-process-cloud.component.spec.ts
```diff
@@ -387,7 +387,7 @@ describe('StartProcessCloudComponent', () => {
             expect(component.processDefinitionList).toBe(fakeProcessDefinitions);
         });
 
-        it('should filter processes in the select list if input is empty', fakeAsync(() => {
+        xit('should NOT filter processes in the select list if input is empty', fakeAsync(() => {
```
DenysVuika: the test is excluded?

suzanadirla: yeah, because I've planned to fix it with a separate PR for https://issues.alfresco.com/jira/browse/ADF-4704 ... but if you prefer I can do it in this PR

suzanadirla:
@eromano, this upgrade to V7 is needed for this ACA issue https://issues.alfresco.com/jira/browse/ACA-2111
I see other apps using ADF have upgraded to a more recent V7 and they would also benefit from this upgrade. 
This upgrade has no breaking changes, so it can be decided later what/when the upgrade to V8 or V9 will be done - so all the apps would know when to update to that major version.

PR #4942, opened 2019-07-17
Add rightclick method to browseractions

gmandakini:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

> - [ ] Yes
> - [ ] No

**Other information**:

### DIFF
--- a/lib/testing/src/lib/core/pages/data-table-component.page.ts
+++ b/lib/testing/src/lib/core/pages/data-table-component.page.ts
@@ -266,7 +266,7 @@ export class DataTableComponentPage {
     }
 
     getRowByIndex(index: number) {
-        const row = this.tableBody.element(by.xpath(`//div[contains(@class,'adf-datatable-row')][${index}]`));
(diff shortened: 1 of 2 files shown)


lib/testing/src/lib/core/pages/data-table-component.page.ts
```diff
@@ -166,6 +166,13 @@ export class DataTableComponentPage {
         BrowserVisibility.waitUntilElementIsVisible(element(by.id('adf-context-menu-content')));
     }
 
+    rightClickOnRowByIndex(index: number) {
+        BrowserActions.closeMenuAndDialogs();
```
adinapitul: this should be handled in the actual tests, not here.
there could be scenarios where we would need to right click on an item and then right click on another (without closing the context menu)

lib/testing/src/lib/core/utils/browser-actions.ts
```diff
@@ -60,6 +60,11 @@ export class BrowserActions {
         return this;
     }
 
+    static async rightClick(elementFinder: ElementFinder) {
```
adinapitul: why "async" if you're not await-ing any promise inside this method?

PR #4973, opened 2019-08-02
[ADF-4793] The filters don't work on community-edition [ Use local storage for community version ]

sivakumar414ram:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

> - [ ] Yes
> - [x] No

**Other information**:


demo-shell/src/app/app.module.ts
```diff
@@ -189,24 +187,17 @@ registerLocaleData(localeSv);
...
-        CommunityProcessDetailsCloudDemoComponent,
-        CommunityTaskDetailsCloudDemoComponent
+        ConfirmDialogExampleComponent
     ],
     providers: [
         {
             provide: HTTP_INTERCEPTORS, useClass:
             AuthBearerInterceptor, multi: true
         },
+        { provide: PROCESS_FILTERS_SERVICE_TOKEN, useClass: UserPreferenceCloudService },
```
DenysVuika: Why having tokens instead of replacing the service?

lib/process-services-cloud/src/lib/process-services-cloud.module.ts
```diff
@@ -46,7 +51,10 @@ import { UserPreferenceCloudService } from './services/user-preference.cloud.ser
             }
         },
         BaseCloudService,
-        UserPreferenceCloudService
+        UserPreferenceCloudService,
+        LocalPreferenceCloudService,
+        { provide: PROCESS_FILTERS_SERVICE_TOKEN, useClass: LocalPreferenceCloudService },
```
DenysVuika: Looks complicated and may require additional docs and explanations to the developers

mauriziovitale: Well by default the module will inject a LocalPreferenceCloudService. 
On the documentation, we can explain how to override that to use the UserPreferenceCloudService

demo-shell/src/app/app.module.ts
popovicsandras: I think tokens are better than using actual services to be rewritten. Why would you need to import the whole service, when you just need to have an identifier for the IoC. Unfortunately Angular's DI can't do this with interfaces, that is why I think tokens are inevitable.
I don't think either, it is that complex, that developers couldn't understand. Almost all of the Angular related components use tokens.

mauriziovitale: I agree it is always better to have an interface in the code instead of a concrete class that can be overridden from the outside. The idea is that by default the Cloud module will inject the localstorage implementation and in case of enterprise we will inject the UserPreferenceService

lib/process-services-cloud/src/lib/process-services-cloud.module.ts
DenysVuika: Why not using it like the following?
```
{ provide: UserPreferenceCloudService, useClass: LocalPreferenceCloudService }
```

popovicsandras: I think the reason is the same as above. Maybe we should agree on a rule here, either we want to use tokens or not, but whatever the decision is, adhere to that.

PR #4996, opened 2019-08-19
 [ADF-4823] added download and show preview for attach cloud widget

VitoAlbano:

> - [ ] Bugfix
> - [X] Feature
> - [ ] Build related changes
> - [ ] Documentation

At the moment is only possible remove a file attached but not download it or see it.

We have added some extra options so it will be possible consult and download the attached files for cloud

> - [ ] Yes
> - [ ] No

**Other information**:

### DIFF
--- a/lib/process-services-cloud/src/lib/form/services/process-cloud-content.service.ts
+++ b/lib/process-services-cloud/src/lib/form/services/process-cloud-content.service.ts
@@ -18,7 +18,7 @@
 import { Injectable } from '@angular/core';
 import { throwError, Observable, from } from 'rxjs';
 import { catchError, map } from 'rxjs/operators';
-import { AlfrescoApiService, LogService } from '@alfresco/adf-core';
+import { AlfrescoApiService, LogService, ContentService } from '@alfresco/adf-core';
 
 @Injectable({
   providedIn: 'root'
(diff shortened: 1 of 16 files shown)


lib/process-services-cloud/src/lib/form/services/process-cloud-content.service.ts
```diff
@@ -45,6 +45,18 @@ export class ProcessCloudContentService {
       );
   }
 
+  getRawContentNode(nodeId: string, contentHost: string): Observable<any> {
+    const changedConfig = this.apiService.lastConfig;
+    changedConfig.provider = 'ALL';
```
DenysVuika: does it modify the config forever or just for this call?

gmandakini:
I have looked at the failure => e2e/process-services-cloud/start-task-form-cloud.e2e.ts
it failing and is specific to this feature branch, as it can't find the folder path
Screenshot 2019-08-20 at 11 55 02

lib/process-services-cloud/src/lib/form/services/process-cloud-content.service.ts
mauriziovitale: it is forever but for the cloud should not be a problem. What is your concern?

DenysVuika: The concern is that the service modifies globally the provider for the entire application. That might not be an expected behavior when ECM is not configured but it will require the login?

VitoAlbano: As per design now the cloud is working with keyclock as auth provider, so at design time you will decide which content you want to have attached to your application. This means that the idea of changing the app config to try to connect to a different content loose it's validity into the cloud side. What we do here is (even though it's not the best and beautyfull way) to basically connect to the content the user configured at design time to fetch the information needed

VitoAlbano:
@gmandakini it looks like a problem CORS related, please double check locally with that disabled also the option is now changed so you will see a menu now

PR #5059, opened 2019-09-05
[ACTIVITI-3746] - fix form completion with tab fields

SilviuCPopa:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://issues.alfresco.com/jira/browse/ACTIVITI-3746

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/core/form/components/widgets/core/form-field.model.ts
+++ b/lib/core/form/components/widgets/core/form-field.model.ts
@@ -26,6 +26,7 @@ import { FormFieldTypes } from './form-field-types';
 import { NumberFieldValidator } from './form-field-validator';
 import { FormWidgetModel } from './form-widget.model';
 import { FormModel } from './form.model';
+import { TabModel } from '@alfresco/adf-core';
 
 // Maps to FormFieldRepresentation
 export class FormFieldModel extends FormWidgetModel {
@@ -123,7 +124,7 @@ export class FormFieldModel extends FormWidgetModel {
         if (!this.readOnly) {
             const validators = this.form.fieldValidators || [];
             for (const validator of validators) {
-                if (!validator.validate(this)) {
+                if (!validator.validate(this) && this.isTabVisible()) {
                     this._isValid = false;
                     return this._isValid;
                 }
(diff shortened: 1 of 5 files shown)


lib/core/form/components/widgets/core/form-field.model.ts
```diff
@@ -134,6 +135,20 @@ export class FormFieldModel extends FormWidgetModel {
         return this._isValid;
     }
 
+    isTabVisible() {
```
eromano: The visibility doesn't look like a concern of this model what di you think @DenysVuika @mauriziovitale ?

DenysVuika: Misses the return value typing. Besides that, the form model already parses tabs and tab fields on init, so I think we are already bound with the notion of tabs. I would suggest leaving here and maybe thinking over splitting in the future

lib/core/form/components/widgets/core/form-field.model.ts
```diff
@@ -209,6 +224,19 @@ export class FormFieldModel extends FormWidgetModel {
         this.updateForm();
     }
 
+    private findFieldByCurrentId(value: any) {
```
DenysVuika: misses return type

lib/core/form/components/widgets/core/form-field.model.ts
eromano: but not visibility

eromano: @DenysVuika should it be concern of visibility service?

DenysVuika: @eromano probably yes. If that api moves to the visibility service, it should also take `findFieldByCurrentId` function there too

mauriziovitale: Yep, I agree it should not be there.

mauriziovitale: I see that we already do some checks on tab visibility 
```
  public refreshVisibility(form: FormModel) {
        if (form && form.tabs && form.tabs.length > 0) {
            form.tabs.map((tabModel) => this.refreshEntityVisibility(tabModel));
        }
```

SilviuCPopa: The issue it's related to form completion. 
Right now we cannot complete a task that has an invisible tab with an invalid form inside.
If we have just an invisible invalid field we can complete it.
In this PR I added a check to make the field invisible if the tab that belongs to is also invisible.

PR #5060, opened 2019-09-06
[ADF-4787] Invalid value of number field validates to true in visibility condition

mcchrys:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://issues.alfresco.com/jira/browse/ADF-4787

Fractions are not supported unless the enableFractions flag is set.

> - [ ] Yes
> - [x] No

**Other information**:


lib/core/form/services/widget-visibility.service.ts
```diff
@@ -91,6 +94,19 @@ export class WidgetVisibilityService {
 
     }
 
+    private isValidFieldValue(form: FormModel, visibilityObj: WidgetVisibilityModel): boolean {
+        let isValid = true;
+        if (visibilityObj.leftType === WidgetTypeEnum.field) {
+            form.getFormFields().forEach((formField: FormFieldModel) => {
+                if (this.isSearchedField(formField, visibilityObj.leftFormFieldId)) {
```
davidcanonieto: It's not a good practice to interrupt a forEach this way. I guess that you're trying to find the field among all the fields. 
Can we use ```form.getFormFields().find((formField: FormFieldModel) => your stuff)``` instead?

mcchrys: Updated.

lib/core/form/services/widget-visibility.service.ts
```diff
@@ -67,6 +67,9 @@ export class WidgetVisibilityService {
         const rightValue = this.getRightValue(form, visibilityObj);
         const actualResult = this.evaluateCondition(leftValue, rightValue, visibilityObj.operator);
 
+        if (!this.isFieldValueValid(form, visibilityObj)) {
+            return false;
```
SilviuCPopa: Returning a value here will block the chaining condition operator. 
I think we need to find another way to fix this problem

lib/core/form/services/widget-visibility.service.spec.ts
```diff
@@ -916,10 +906,10 @@ describe('WidgetVisibilityService', () => {
         });
 
         it('should not set null value when the field is not visibile', () => {
-            visibilityObjTest.leftFormFieldId = 'test_4';
+            visibilityObjTest.leftFormFieldId = 'FIELD_TEST';
             visibilityObjTest.operator = '!=';
-            visibilityObjTest.rightFormFieldId = 'dropdown';
-            const fakeFormField: FormFieldModel = new FormFieldModel(formTest, jsonFieldFake);
+            visibilityObjTest.rightFormFieldId = 'RIGHT_FORM_FIELD_ID';
+            const fakeFormField: FormFieldModel = new FormFieldModel(fakeFormWithField, jsonFieldFake);
```
SilviuCPopa: Replacing the mock here and in any other places will not fix the unit tests.

lib/core/form/services/widget-visibility.service.ts
mcchrys: The expected behaviour is that the visibility condition should fail, if any one field in the condition has an invalid value. In that case, it is unnecessary to proceed with evaluating the condition

lib/core/form/services/widget-visibility.service.spec.ts
mcchrys: The existing formModel (formTest) hold only the _values_ property, and does not contain the  _fields_ property in the model. In this update, we check the _isValid_ property of the fields, which will not be available in the currently used formModel (formTest), which causes the tests to fail. Hence the changes.

DenysVuika:
@mcchrys what is the state of this PR? can you please update it

mcchrys:
> @mcchrys what is the state of this PR? can you please update it

Some changes have been suggested by Eugenio, as to where the value validity check should be done. I am working on it.

lib/core/form/services/widget-visibility.service.ts
```diff
@@ -91,6 +94,17 @@ export class WidgetVisibilityService {
 
     }
 
+    private isValidFieldValue(formField: FormFieldModel): boolean {
+        let isValid = true;
+        isValid = formField && formField.isValid ? true : false;
+        return isValid;
+    }
+
+    private checkVisibilityValue(visibilityObj, leftValue, rightValue) {
```
eromano: missign return type

eromano:
@mcchrys which is the status of this PR?

mcchrys:
> @mcchrys which is the status of this PR?

This issue fix is in need of further discussion since there are some concerns about using the form values from _formField_ object to check validity.

mcchrys:
Solution provided in this PR: https://github.com/Alfresco/alfresco-ng2-components/pull/5255

PR #5061, opened 2019-09-06
 [DW-1478] E2e - Move 'filter' component from Admin Application

rdiacenco:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation
> - [x] Other... Please describe: E2e framework updates

> - [ ] Yes
> - [x] No

**Other information**:


lib/testing/src/lib/core/utils/browser-actions.ts
```diff
@@ -80,6 +89,16 @@ export class BrowserActions {
         await elementFinder.sendKeys(text);
     }
 
+    static async sendKeysIfVisible(elementFinder: ElementFinder, text: string) {
```
eromano: All seems ok to be the only thing I am not sure if its a good idea to have methods like that or not. @gmandakini @adinapitul feelings about it?

lib/testing/src/lib/process-services-cloud/pages/edit-process-filter-cloud-component.page.ts
```diff
@@ -27,14 +27,19 @@ export class EditProcessFilterCloudComponentPage {
...
+    filter: ElementFinder = element(by.css(`adf-cloud-edit-process-filter mat-expansion-panel-header`));
 
     editProcessFilterDialogPage = new EditProcessFilterDialogPage();
 
     editProcessFilterDialog(): EditProcessFilterDialogPage {
         return this.editProcessFilterDialogPage;
     }
 
-    async clickCustomiseFilterHeader(): Promise<void> {
+    async isFilterDisplayed(): Promise<Boolean> {
```
adinapitul: I'm a bit concerned about this type of usage.
method set to return Promise<Boolean>
but inside return waitUntilElementIsPresent, which returns Promise<any>, and which will never ever return false.

lib/testing/src/lib/core/utils/browser-actions.ts
adinapitul: TBH, I don't understand the need for these 2 sendKeysIf.. methods. What problem are they solving?

what is your concern about these, @eromano ?

gmandakini: what is the difference? Having the same kind of method for visible and present...

lib/testing/src/lib/process-services-cloud/pages/edit-task-filter-cloud-component.page.ts
```diff
@@ -38,14 +38,19 @@ export class EditTaskFilterCloudComponentPage {
...
+    filter: ElementFinder = element(by.css(`adf-cloud-edit-task-filter mat-expansion-panel-header`));
 
     editTaskFilterDialogPage = new EditTaskFilterDialogPage();
 
     editTaskFilterDialog(): EditTaskFilterDialogPage {
         return this.editTaskFilterDialogPage;
     }
 
-    async clickCustomiseFilterHeader(): Promise<void> {
+    async isFilterDisplayed(): Promise<Boolean> {
```
adinapitul: a filter is displayed when it is "present" ? shouldn't it be visible?

lib/testing/src/lib/core/utils/browser-actions.ts
rdiacenco: These 2 methods add an extra check for the element to be visible or present, before sending the keys.
The difference between the 2 methods is that the element can be present in the DOM, but not visible, still we need to be able to sendkeys to it.

lib/testing/src/lib/process-services-cloud/pages/edit-task-filter-cloud-component.page.ts
rdiacenco: yes

lib/testing/src/lib/core/utils/browser-actions.ts
adinapitul: can _clearSendKeys_ be used instead of sendKeysIfVisible ?

in any case, naming wise I think I would use "when" instead of "if" . English wise, _when_ implies waiting. "if" makes me wonder what if it's not? :)

lib/testing/src/lib/process-services-cloud/pages/edit-process-filter-cloud-component.page.ts
rdiacenco: Could you be more specific?
Async functions return promises, what I think you meant is not to wait for the promise to be resolved:
return BrowserVisibility.waitUntilElementIsPresent(this.filter);

lib/testing/src/lib/core/utils/browser-actions.ts
rdiacenco: Since the 2 methods generated to much discussion, I will remove them.

lib/testing/src/lib/process-services-cloud/pages/edit-process-filter-cloud-component.page.ts
adinapitul: sorry, formatting removed parts of my message.
Edited now, makes a bit of sense?

adinapitul: IMO, a promise of boolean method should return either _true_ or _false_.
While this method would return true in the happy case, and would timeout otherwise.

I think I saw in another PR of yours a different approach that I liked. Something like - writing from memory:
```
try
{
 await BrowserVisibility.waitUntilElementIsVisible(...);
 return true;
} catch {
 return false;
}
```

eromano: do we need really to return true or false in this kind of cases?

adinapitul: Depends on the usage. If it's to be used in an expect, then returning true or false makes sense.

Alternative would be to have promise of void as return type and remove "return"

I'd vote for boolean though, to be able to correctly use those types of methods in expects

lib/testing/src/lib/core/utils/browser-actions.ts
rdiacenco: Like more of the methods from this file, the 2 I added are using a default Webdriver method (sendkeys) with an additional explicit wait for the element to be visible or present.
It would be great to at least understand what is causing so much confusion/discussions around them.
I'm removing both just because they are not in scope of this PR, but I've used them in my previous tests and I found them useful.

lib/testing/src/lib/process-services-cloud/pages/edit-process-filter-cloud-component.page.ts
rdiacenco: In this particular case, the method is used in an assertion.
If the element is visible, then it will return true, otherwise it might return false (falseIfMissing()) or time out.
If false is returned, then the assertion fails, otherwise the test will fail with timeout.
The artifice with try catch, was just for thoroughness, but I don't see it as a necessary need.

PR #5064, opened 2019-09-09
Refactor, Split ADF About component

arditdomi:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://issues.alfresco.com/jira/browse/ADF-4878

The ADF about component has been splitted into 3 in order to be able to use every piece that is needed across each application (AMA,AAA etc.)

> - [ ] Yes
> - [ ] No

**Other information**:


docs/core/components/about-application-configuration.md
```diff
@@ -0,0 +1,40 @@
...
+With default input values
+
+```html
+<adf-about-application-configuration></adf-about-application-configuration>
+```
+
+With custom input values:
+
+```html
+<adf-about-application-configuration
```
eromano: can we call it adf-about-githubb-link?

docs/core/components/about-github-link.md
```diff
@@ -0,0 +1,40 @@
...
+
+```html
+<adf-about-github-link></adf-about-github-link>
+```
+
+With custom input values:
+
+```html
+<adf-about-github-link
+    githubUrlCommitAlpha="YourUrl"
```
DenysVuika: the component tag already says "github-link", so we can remove that from the property name. Also I would suggest just calling the property "url" as we may want to display something else than "alpha" in the future. What do you think @eromano ?

lib/core/about/about-application-modules/about-application-modules.component.ts
```diff
@@ -0,0 +1,67 @@
...
+    /** Current version of the app running */
+    @Input() dependencies: any;
+
+    constructor(appExtensions: AppExtensionService) {
+        this.extensions$ = appExtensions.references$;
+    }
+
+    ngOnInit() {
+        const alfrescoPackages = Object.keys(this.dependencies).filter((val) => {
+            return new RegExp(this.regexp).test(val);
```
DenysVuika: is this responsibility of the component? I believe it should just show what the "dependencies" provide, and application can filter out based on different criteria. What if customers want to show more libraries there?

lib/core/about/about-github-link/about-github-link.component.ts
```diff
@@ -0,0 +1,44 @@
...
+@Component({
+    selector: 'adf-about-github-link',
+    templateUrl: './about-github-link.component.html',
+    encapsulation: ViewEncapsulation.None
+})
+export class AboutGithubLinkComponent {
+
+    /** Commit corresponding to the version of ADF to be used. */
+    @Input()
+    githubUrlCommitAlpha = 'https://github.com/Alfresco/alfresco-ng2-components/commits/';
```
DenysVuika: Can this be just `url` ?

lib/core/about/about-github-link/about-github-link.component.ts
```diff
@@ -0,0 +1,44 @@
...
+
+    /** Current version of the app running */
+    @Input() version: string;
+
+    ecmHost = '';
+    bpmHost = '';
+    application: string;
+
+    constructor(private appConfig: AppConfigService) {
+        this.ecmHost = this.appConfig.get<string>(AppConfigValues.ECMHOST);
```
DenysVuika: can you move this code to `ngOnInit` instead (good practice)

docs/core/components/about-github-link.md
eromano: good idea

lib/core/about/about-application-modules/about-application-modules.component.ts
eromano: maybe we can imporve this ting later what do you think @DenysVuika ?

lib/core/about/about-github-link/about-github-link.component.ts
eromano: agree

lib/core/about/about-application-modules/about-application-modules.component.ts
DenysVuika: Yeah, I'm fine with improving later

PR #5070, opened 2019-09-11
use custom chrome version with puppeteer

DenysVuika:
use custom Chrome that comes with puppeteer instead of the Travis version

For more details, see:

- https://github.com/karma-runner/karma-chrome-launcher#headless-chromium-with-puppeteer
- https://medium.com/@rogeriopvl/automating-karma-and-headless-chrome-with-puppetteer-51ce8f6a78b0


protractor.conf.ts
```diff
@@ -9,6 +9,8 @@ const { beforeAllRewrite, afterAllRewrite, beforeEachAllRewrite, afterEachAllRew
 const { uploadScreenshot, saveReport, cleanReportFolder } = require('./e2e/protractor/save-remote');
 const argv = require('yargs').argv;
 
+process.env.CHROME_BIN = require('puppeteer').executablePath();
```
adinapitul: is this used somewhere?

DenysVuika:
yeah, I will cleanup tests once I get the green e2e one

PR #5296, opened 2019-11-28
[AAE-1022] Fix components to work on Community environments

davidcanonieto:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://issues.alfresco.com/jira/browse/AAE-1022

This PR fixes the components so there is no need to pass a non-existing appName and it also fixes the service that makes the community work and write its filters in the local storage.

> - [ ] Yes
> - [ ] No

**Other information**:

### DIFF
--- a/demo-shell/src/app/app.module.ts
+++ b/demo-shell/src/app/app.module.ts
@@ -84,10 +84,7 @@ import { environment } from '../environments/environment';
 import { AppCloudSharedModule } from './components/cloud/shared/cloud.shared.module';
 import { DemoErrorComponent } from './components/error/demo-error.component';
 import {
-    ProcessServicesCloudModule,
-    UserPreferenceCloudService,
-    PROCESS_FILTERS_SERVICE_TOKEN,
-    TASK_FILTERS_SERVICE_TOKEN
+    ProcessServicesCloudModule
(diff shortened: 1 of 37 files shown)


demo-shell/src/app/app.module.ts
```diff
@@ -199,8 +196,6 @@ registerLocaleData(localeSv);
             provide: HTTP_INTERCEPTORS, useClass:
             AuthBearerInterceptor, multi: true
         },
-        { provide: PROCESS_FILTERS_SERVICE_TOKEN, useClass: UserPreferenceCloudService },
```
sivakumar414ram: The reason behind this line was :
For enterprise: We should call actual preference services to get task/process filters.
For the community: We should use local storage to get task/process filters.

Is it not required anymore?
Can you please check once?

davidcanonieto: When adding providers, specially with token injection, it's worth knowing that the parent provider overwrites a child provider. Because of this, the app.module.ts will overwrite all tokens of that kind making the community part also use the UserPreferenceService.

As it stands with this PR, it always makes use of the UserPreferenceService except for when it's overwritten in the community module.

PR #5312, opened 2019-12-04
[ADF-5009] close all the dialogs after false canActivate event

eromano:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

All the open modal dialogs needs handling and closing when the user logs out.

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/core/services/auth-guard-base.ts
+++ b/lib/core/services/auth-guard-base.ts
@@ -29,6 +29,7 @@ import {
     AppConfigValues
 } from '../app-config/app-config.service';
 import { OauthConfigModel } from '../models/oauth-config.model';
+import { MatDialog } from '@angular/material';
 
 export abstract class AuthGuardBase implements CanActivate, CanActivateChild {
     abstract checkLogin(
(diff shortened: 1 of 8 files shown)


pionnegru:
shouldn't this be the responsibility of the app and not auth guard? AlfrescoApiService has ```.on('error')``` that can do the trick

eromano:
> shouldn't this be the responsibility of the app and not auth guard? AlfrescoApiService has `.on('error')` that can do the trick

@pionnegru  I think the problem is that you can be redirected to the login also without error. 
There is another scenario when another tab logs out.
I anyway agree as concern doesn't really look the better place. Do you have another idea at the light of this other possible scenario?

pionnegru:
I guess the app could listen for NavigationEnd event to be the login route

eromano:
I tryed other way but they weren't working or were making the code even more complicate . I am still open to other suggestion otherwise I suggest to merge this

PR #5360, opened 2020-01-08
[ACA-2755] [SSO] Unable to access a private url after a public url was loaded in the same browser tab

mmuller88:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

See bug description. After loading public urls like /settings you couldn't change to other urls properly in the SSO config

Now it is forwarding you to login.

> - [ ] Yes
> - [ ] No

**Other information**:

### DIFF
--- a/lib/core/services/auth-guard-base.ts
+++ b/lib/core/services/auth-guard-base.ts
@@ -54,8 +54,7 @@ export abstract class AuthGuardBase implements CanActivate, CanActivateChild {
         protected appConfigService: AppConfigService,
         protected dialog: MatDialog
     ) {
-        if(this.authenticationService.isOauth())
-        {
+        if (this.authenticationService.isOauth()) {
             const apiService = new AlfrescoApi(this.appConfigService.config);
             this.oauth2Auth = new Oauth2Auth(this.appConfigService.config, apiService);
         }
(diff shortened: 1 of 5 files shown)


mmuller88:
@eromano thx for the feedback. I will try to use the existing js api method as it is just a copy of that. Regarding unit test I will see how much efforts that would be. But you know best it is SSO + silentlogin + Authguard + public / private url :-/

mmuller88:
@eromano Requested changes 1) using js-api 2) unit test are done pls review

lib/core/services/auth-guard-base.ts
```diff
@@ -70,6 +70,10 @@ export abstract class AuthGuardBase implements CanActivate, CanActivateChild {
             this.dialog.closeAll();
         }
 
+        if (this.isSilentLogin() && this.oauth2Auth && !this.oauth2Auth.isPublicUrl()) {
```
eromano: this method has only one exit point so to make it more readable can we keep it?

mmuller88: Sure but I doubt that it is more readable. I still assume that some of the SSO stuff shouldn't live in alfresco-js-api at all :-/

lib/core/services/auth-guard-base.ts
```diff
@@ -61,7 +62,7 @@ export abstract class AuthGuardBase implements CanActivate, CanActivateChild {
             this.dialog.closeAll();
         }
 
-        return checkLogin;
+        return true;
```
pionnegru: correct me if Im wrong, but this means that any other guards extending this base service, with its own 'checkLogin' implementation, [example this one](https://github.com/Alfresco/alfresco-ng2-components/blob/development/lib/core/services/auth-guard-ecm.service.ts#L40), event if it returns `false` wi discard it and return true?

PR #5381, opened 2020-01-21
[ADA-2474] adding instructions to the calendar for use

downzer0:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation
> - [X] Other... accessibility enhancement

While the Material Design calendar/datepicker is keyboard accessible, the elements aren't announced as functional. This creates a block of text providing instructions for screen reader users on how to interact with the control.

The text instructions will be announced when the user controls the datepicker.

> - [ ] Yes
> - [X] No

**Other information**:

### DIFF
--- a/lib/core/card-view/components/card-view-dateitem/card-view-dateitem.component.html
+++ b/lib/core/card-view/components/card-view-dateitem/card-view-dateitem.component.html
@@ -1,5 +1,6 @@
 <div [attr.data-automation-id]="'card-dateitem-label-' + property.key" class="adf-property-label" *ngIf="showProperty() || isEditable()">{{ property.label | translate }}</div>
 <div class="adf-property-value">
+<div class="adf-sr-only" id="{{uniqueId}}">{{'CORE.METADATA.ACCESSIBILITY.DATEPICKER' | translate}}</div>
     <span *ngIf="!isEditable()" [attr.data-automation-id]="'card-' + property.type + '-value-' + property.key">
         <span [attr.data-automation-id]="'card-dateitem-' + property.key">
             <span *ngIf="showProperty()">{{ property.displayValue }}</span>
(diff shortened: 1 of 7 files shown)


lib/core/styles/_a11y.scss
```diff
@@ -0,0 +1,8 @@
+.adf-sr-only {
```
eromano: better to add in card-view-dateitem.component.scss

lib/core/card-view/components/card-view-dateitem/card-view-dateitem.component.html
```diff
@@ -1,5 +1,6 @@
 <div [attr.data-automation-id]="'card-dateitem-label-' + property.key" class="adf-property-label" *ngIf="showProperty() || isEditable()">{{ property.label | translate }}</div>
 <div class="adf-property-value">
+<div class="adf-sr-only" id="{{uniqueId}}">{{'CORE.METADATA.ACCESSIBILITY.DATEPICKER' | translate}}</div>
```
eromano: can we use  [attr.data-automation-id]="'card-' + property.type + '-value-' + property.key" as uniqueId?

lib/core/styles/_a11y.scss
downzer0: this can be used globally, not just in cards. anything you want to make available only to screenreaders can use this class. is this still an appropriate location?

lib/core/card-view/components/card-view-dateitem/card-view-dateitem.component.html
downzer0: this unique id needs to be shared with the aria-describedby shortly below. is it possible to use the same id string in both places?

PR #5399, opened 2020-01-24
[ACA-2789] - edit button a11y fixes

tayrichie:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

Button label of "Edit" is not read on Screen Reader

Button is read as "Edit Button" by Screen Reader

> - [ ] Yes
> - [x] No

**Other information**:
Universally, a `title` attribute can affect how the screen reader announces the data of the focused element; all that is needed in this case is an `aria-label`.

### DIFF
--- a/lib/content-services/src/lib/content-metadata/components/content-metadata-card/content-metadata-card.component.html
+++ b/lib/content-services/src/lib/content-metadata/components/content-metadata-card/content-metadata-card.component.html
@@ -16,7 +16,6 @@
             <button *ngIf="!readOnly && hasAllowableOperations()"
                 mat-icon-button
                 (click)="toggleEdit()"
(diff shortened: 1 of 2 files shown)


eromano:
Removing this attribute make some e2e failing maybe they are wrongly relying on this attribute. Something to fix in the e2e page

tayrichie:
@eromano would it be sufficient to remove any tests checking for the `title` property? And am I correct in looking at `e2e/content-services/metadata`? Thanks!

eromano:
C270952 is the only failing I think you need to replace the selector for it

DenysVuika:
As for the failing test, the button loses the tooltip on mouse hover with this PR, so the test fails correctly. While supporting the screen readers, we should not break the desktop features.

PR #5416, opened 2020-01-28
[ACA-2597] search chip list accessibility

downzer0:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

the chip list and chips had improper aria parent/child use as well as funky tabindexes.

i had to use some post-DOM load tweaks to set and unset various attributes. the way material nests the parent/child structure is incorrect, so i had to set some elements to "presentation" roles to get past this.

> - [ ] Yes
> - [X] No

**Other information**:


lib/content-services/src/lib/search/components/search-chip-list/directives/search-chip-list.directive.ts
```diff
@@ -0,0 +1,32 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { Directive, ElementRef, Renderer2 } from '@angular/core';
+
+@Directive({
+  selector: '[adf-a11y-search-chip-list]'
```
pionnegru: I see that there are common issues we want to fix throughout the application components, mainly `role` and `tabindex`.  Why not have a configurable generic directive that can do that? what this directive manipulate could not be specific only to search-chip-list or search-chip.

lib/content-services/src/lib/search/components/search-chip-list/directives/search-chip.directive.ts
```diff
@@ -0,0 +1,35 @@
...
+@Directive({
+  selector: '[adf-a11y-search-chip]'
+})
+export class SearchChipDirective {
+    constructor(private el: ElementRef, private renderer: Renderer2) {
+    }
+
+    ngAfterViewInit() {
+        setTimeout(() => {
+            this.renderer.removeAttribute(this.el.nativeElement.parentNode.parentNode.parentNode.parentNode, 'role');
```
pionnegru: to avoid this `.parentNode.parentNode.parentNode.parentNode` why not give the directive a target element so one could understand what element is being affected ?

downzer0:
@pionnegru i was able to make your requested changes, but by targeting the element i got lint errors (because they aren't prefixed with adf-). i ignored those lines, which i know isn't ideal. let me know how you handle this with the linter.

pionnegru:
@downzer0 
the idea was that from `SearchChipListDirective` and `SearchChipDirective` could benefit other apps instances if they were to reside in `core` lib. Maybe `A11yChipListDirective` and `A11yChipDirective` would be more generic.

lib/content-services/src/lib/search/components/search-chip-list/directives/search-chip.directive.ts
```diff
@@ -0,0 +1,38 @@
...
+  // tslint:disable-next-line: directive-selector
+  selector: 'mat-chip'
+})
+export class SearchChipDirective {
+    constructor(private el: ElementRef, private renderer: Renderer2) {
+    }
+
+    ngAfterViewInit() {
+        setTimeout(() => {
+            const chipListWrapper = document.querySelector('.mat-chip-list-wrapper');
```
pionnegru: maybe not duplicate what parent component directive already does https://github.com/Alfresco/alfresco-ng2-components/pull/5416/files#diff-887ade528360c14ec6dbe52ac689b4f1R29

lib/core/directives/search-chip-list.directive.ts
```diff
@@ -0,0 +1,33 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { Directive, ElementRef, Renderer2 } from '@angular/core';
+
+@Directive({
+  // tslint:disable-next-line: directive-selector
+  selector: 'mat-chip-list'
```
pionnegru: the problem with this selector is that anyone who will be using mat-chip-list and mat-chip will have these DOM changes even if its not desired. If someone wants just to display in a chip list some visual tags
```html
<mat-chip-list>
  <mat-chip>Tag 1</mat-chip>
  <mat-chip>Tag 2</mat-chip>
  <mat-chip>Tag 3</mat-chip>
</mat-chip-list>
```
will result in 

```html
<mat-chip-list>
  <mat-chip role="button">Tag 1</mat-chip>
  <mat-chip role="button">Tag 2</mat-chip>
  <mat-chip role="button">Tag 3</mat-chip>
</mat-chip-list>
```

lib/core/directives/search-chip.directive.ts
```diff
@@ -0,0 +1,38 @@
...
+  // tslint:disable-next-line: directive-selector
+  selector: 'mat-chip'
+})
+export class SearchChipDirective {
+    constructor(private el: ElementRef, private renderer: Renderer2) {
+    }
+
+    ngAfterViewInit() {
+        setTimeout(() => {
+            const chipListWrapper = document.querySelector('.mat-chip-list-wrapper');
```
pionnegru: this should have been left in chip-list. right now, for every chip  we will have a DOM manipulation for the same element

lib/core/directives/search-chip-list.directive.ts
downzer0: @pionnegru so is there a way to do this level of selectivity and still make this directive a core directive?

pionnegru: since we want a to change the attributes in case of a list of chips that are actionable we can have ```selector: ['adf-a11y-action-chip-list']``` or something close to underline the intent of this directive. Same with `mat-chip`, can be ```selector: ['adf-a11y-action-chip']```.  
Then we can use them like normal directives

eromano:
Are we sure we can not leverage those problems with the standard API of this component https://material.angular.io/components/chips/api ?  Post attribute rendering modify are really bad

DenysVuika:
I agree with @eromano. Google guys spend a lot of time supporting a11y in the Angular Components, and hacking those components via the custom directives does not look like a clean solution.

DenysVuika:
Closing as the underlying component already provides necessary a11y support:
https://material.angular.io/components/chips/overview#accessibility

```
A chip-list behaves as a role="listbox", with each chip being a role="option". 
The chip input should have a placeholder or be given a meaningful label via aria-label or aria-labelledby.
```

PR #5432, opened 2020-01-30
[ACA-2568] Permission-List - Removal of List ARIA

Steady5063:

> - [ X] The commit message follows our [guidelines](https://github.com/Alfresco/alfresco-ng2-components/wiki/Commit-format)

> - [ X] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

Currently items in the table are reading as a list.

The items in the table will maintain their table markup and semantic meaning.

> - [ ] Yes
> - [X ] No

**Other information**:

### DIFF
--- a/lib/content-services/src/lib/permission-manager/components/mat-chip.directive.ts
+++ b/lib/content-services/src/lib/permission-manager/components/mat-chip.directive.ts
@@ -1,0 +1,33 @@
+/*!
+ * @license
+ * Copyright 2019 Alfresco Software, Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* tslint:disable:no-input-rename  */
+import { Directive, ElementRef, Renderer2 } from '@angular/core';
+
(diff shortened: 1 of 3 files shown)


lib/content-services/src/lib/permission-manager/components/mat-chip.directive.ts
```diff
@@ -0,0 +1,33 @@
...
+
+@Directive({
+    selector: '[adf-list-a11y]'
+})
+export class MatChipDirective {
+    constructor(private el: ElementRef, private renderer: Renderer2) {}
+
+    ngAfterViewInit() {
+        setTimeout(() => {
+            this.renderer.removeAttribute(this.el.nativeElement, 'role');
```
pionnegru: what would be the difference with what we do [here](https://github.com/Alfresco/alfresco-ng2-components/pull/5416/files#diff-a19aedfa9dcd553f1e38dedc522a1f02R29-R31)

lib/content-services/src/lib/permission-manager/components/mat-chip.directive.ts
```diff
@@ -0,0 +1,33 @@
...
+            this.renderer.removeAttribute(this.el.nativeElement.querySelector('.mat-chip'), 'role');
```
pionnegru: [attr.role]="null" on <mat-chip> could do this trick

lib/content-services/src/lib/permission-manager/components/mat-chip.directive.ts
Steady5063: If we end up making Chris' change global, then we would not need this specifically. We would just need to add the attribute he uses for selector on this and it should work

DenysVuika:
Closing this PR as Google already provides all the necessary aria elements for this component:
https://material.angular.io/components/chips/overview#accessibility

```
A chip-list behaves as a role="listbox", with each chip being a role="option". 
The chip input should have a placeholder or be given a meaningful label via aria-label or aria-labelledby.
```

PR #5454, opened 2020-02-06
[ACA-2619][ACA-2616] a11y fixes for Share Link dialog

Steady5063:

> - [ X] The commit message follows our [guidelines](https://github.com/Alfresco/alfresco-ng2-components/wiki/Commit-format)

> - [X ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

After switching on the expiration toggle, the calendar appears. This is not proper from an accessibility standpoint. 

The field will be disabled if there is not date added into it. and when the switch toggle is turned on the field will become enabled. The action of the calendar appearing on click has been removed. 

> - [ ] Yes
> - [ X] No

**Other information**:

### DIFF
--- a/e2e/content-services/share-file/share-file.e2e.ts
+++ b/e2e/content-services/share-file/share-file.e2e.ts
@@ -121,6 +121,7 @@ describe('Share file', () => {
         it('[C286578] Should disable today option in expiration day calendar', async () => {
             await contentServicesPage.clickShareButton();
             await shareDialog.checkDialogIsDisplayed();
+            await shareDialog.clickExpireToggle();
             await shareDialog.clickDateTimePickerButton();
             await shareDialog.calendarTodayDayIsDisabled();
             await BrowserActions.closeMenuAndDialogs();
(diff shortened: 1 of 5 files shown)


lib/content-services/src/lib/content-node-share/content-node-share.dialog.ts
```diff
@@ -126,6 +126,12 @@ export class ShareDialogComponent implements OnInit, OnDestroy {
         }
     }
 
+    ngAfterViewInit() {
```
eromano: all this change afterViewInit should be avoided

DenysVuika: I have removed that part

e2e/content-services/share-file/share-file.e2e.ts
```diff
@@ -142,18 +144,11 @@ describe('Share file', () => {
             await BrowserActions.closeMenuAndDialogs();
         });
 
-        it('[C286578] Should disable today option in expiration day calendar', async () => {
```
adinapitul: why are you removing this test?

DenysVuika: it's a copy-pasted one, the same test is above :)

adinapitul: hehe 😄

PR #5506, opened 2020-02-24
[AAE-1884] e2e shared file and refactor custom sources

eromano:

> - [ ] Bugfix
> - [ ] Feature
> - [x] Build related changes
> - [ ] Documentation

E2E hared file and refactor custom sources

> - [ ] Yes
> - [ ] No

**Other information**:


e2e/content-services/share-file/share-file.e2e.ts
```diff
@@ -199,5 +201,19 @@ describe('Share file', () => {
             await BrowserActions.getUrl(sharedLink);
             await viewerPage.checkFileNameIsDisplayed(pngFileModel.name);
         });
+
+        it('[C260153] Should shared files listed in share files custom resources', async () => {
```
adinapitul: this test name is not very clear

e2e/content-services/share-file/share-file.e2e.ts
```diff
@@ -199,5 +201,19 @@ describe('Share file', () => {
...
+            await contentListPage.selectRow(pngFileModel.name);
+            await contentServicesPage.clickShareButton();
+            await shareDialog.checkDialogIsDisplayed();
+            await shareDialog.clickShareLinkButton();
+
+            await BrowserActions.closeMenuAndDialogs();
+            await browser.sleep(30000); //it get really long to update the shared link file list
```
adinapitul: why not 27000 ? :)

adinapitul: there must be another way of waiting, like maybe querying the server a few times until total number of items is the expected one

e2e/content-services/share-file/share-file.e2e.ts
adinapitul: what is the purpose of this file? to test the Share action from UI or to check that a shared file appears in the Shared Files list?

if the purpose is to test that a shared file appears in the Shared Files list, then the creating and sharing of the file would be a precondition, meaning the sharing of the file can be done using an API call, not using the UI.

eromano: at this point is not anymore an e2e it becomes a unit test. E2E is to test a behavior flow this is the scope of it.  I see often this problem we write E2E thinking like a unit test. In this flow, we share a file and check if it appears in the list of the shared file.

 Should shared files listed in share files custom resources do you have a better name ?

e2e/content-services/share-file/share-file.e2e.ts
eromano: I don't see how this thing could improve the test

e2e/content-services/share-file/share-file.e2e.ts
adinapitul: Maybe something like "Shared files are listed in the Shared custom source"

adinapitul: OK, so from your comment above I understand you consider this an integration test, which in general are useful.

But somehow, for this specific case, I don't think it brings much value, cause:
- when you share a file from the UI, what basically happens is you set some node properties in the BE. This is covered by the other tests from this suite
- when you check the files are listed in the Shared files custom source, you're basically testing the API endpoint response, and for testing this, doesn't matter how the file is shared, if it's by a BE call or using the UI

e2e/content-services/share-file/share-file.e2e.ts
adinapitul: having hard coded values for browser.sleep is never a good idea.

This test will always wait 30 secs, no matter what.

I did a check using a retry mechanism that queries the backend every 1 sec, and the average I got was around 5-6 seconds. 
Of course, at times it might take longer, but at times it can also take less. Smart waits are always prefered instead of waiting for a fixed amount of time, and 30 seconds for a single test is a lot!

e2e/content-services/share-file/share-file.e2e.ts
eromano: End-to-end testing is a methodology used to test whether the flow of an application is performing as designed from start to finish. The purpose of carrying out end-to-end tests is to identify system dependencies and to ensure that the right information is passed between various system components and systems

If you are testing only one behaviour of one component you can archive a better result with a unit test without spin up the environment backend and frontend. This is the reason why really few e2e make sense in ADF. This e2e was written and manual executed if you think it doesn't make sense as e2e we can deprecated but please not confuse e2e with unit test. If we etite e2e to test unit test scenario we are going only to have slower unit test instead of e2e

e2e/content-services/share-file/share-file.e2e.ts
eromano: @adinapitul do you already have this mechanism ?

adinapitul: yes, see this in ACA:
https://github.com/Alfresco/alfresco-content-app/blob/development/e2e/utilities/repo-client/apis/shared-links/shared-links-api.ts#L92

e2e/pages/adf/demo-shell/custom-sources.page.ts
```diff
@@ -56,18 +51,19 @@ export class CustomSources {
...
 
-    async clickOnSourceType(): Promise<void> {
-        await BrowserActions.click(this.sourceTypeDropdown);
-    }
-
     async selectMySitesSourceType(): Promise<void> {
-        await this.clickOnSourceType();
-        await BrowserActions.click(this.getSourceType(source.mySites));
+        await this.selectModeDropdown.clickDropdown();
+        await this.selectModeDropdown.selectOption(source.mySites);
```
arditdomi: there is a method called selectDropdownWithOption that is wrapping this 2 methods. Can we use that one?

e2e/pages/adf/demo-shell/custom-sources.page.ts
```diff
@@ -56,18 +51,19 @@ export class CustomSources {
...
     }
 
     async selectFavoritesSourceType(): Promise<void> {
-        await this.clickOnSourceType();
-        await BrowserActions.click(this.getSourceType(source.favorites));
+        await this.selectModeDropdown.clickDropdown();
+        await this.selectModeDropdown.selectOption(source.favorites);
```
arditdomi: same here

e2e/pages/adf/demo-shell/custom-sources.page.ts
```diff
@@ -56,18 +51,19 @@ export class CustomSources {
...
+    }
+
+    async selectSharedLinksSourceType(): Promise<void> {
+        await this.selectModeDropdown.clickDropdown();
+        await this.selectModeDropdown.selectOption(source.sharedLinks);
```
arditdomi: same here

PR #5511, opened 2020-02-25
filter facets : replace startsWith by includes

loictrichaud:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

when you filter a facet, you use a "startsWith" which doesn't make sense for a search. If you have values that all start with the same word (example: Type A, Type B, Type C), you must first type "Type" before you can filter on the letter A, B or C.

Facet Type :  "Type A, Type A1, Type A2, Type B1".
If you search for "A" in the search filter, you'll get the facets "Type A, Type A1, Type A2".

And I'd like to take this opportunity to switch the function to protected. If it had been in protected, I could have overloaded this method to replace the "startsWith" by the "Includes".
Except that your private function prevents me from working and satisfying my customer! 
Your methods should all be in protected ! 

And

> - [ ] Yes
> - [X] No

**Other information**:

### DIFF
--- a/lib/content-services/src/lib/search/components/search-filter/search-filter.component.ts
+++ b/lib/content-services/src/lib/search/components/search-filter/search-filter.component.ts
@@ -397,12 +397,12 @@ export class SearchFilterComponent implements OnInit, OnDestroy {
         });
     }
 
-    private getBucketFilterFunction(bucketList) {
+    protected getBucketFilterFunction(bucketList) {
         return (bucket: FacetFieldBucket): boolean => {
             if (bucket && bucketList.filterText) {
                 const pattern = (bucketList.filterText || '').toLowerCase();
                 const label = (this.translationService.instant(bucket.display) || this.translationService.instant(bucket.label)).toLowerCase();
-                return this.queryBuilder.config.filterWithContains ? label.indexOf(pattern) !== -1 : label.startsWith(pattern);
+                return this.queryBuilder.config.filterWithContains ? label.indexOf(pattern) !== -1 : label.includes(pattern);
             }
             return true;
         };


lib/content-services/src/lib/search/components/search-filter/search-filter.component.ts
```diff
@@ -397,12 +397,12 @@ export class SearchFilterComponent implements OnInit, OnDestroy {
...
     }
 
-    private getBucketFilterFunction(bucketList) {
+    protected getBucketFilterFunction(bucketList) {
         return (bucket: FacetFieldBucket): boolean => {
             if (bucket && bucketList.filterText) {
                 const pattern = (bucketList.filterText || '').toLowerCase();
                 const label = (this.translationService.instant(bucket.display) || this.translationService.instant(bucket.label)).toLowerCase();
-                return this.queryBuilder.config.filterWithContains ? label.indexOf(pattern) !== -1 : label.startsWith(pattern);
+                return this.queryBuilder.config.filterWithContains ? label.indexOf(pattern) !== -1 : label.includes(pattern);
```
DenysVuika: what's the difference then compared to the other condition and `filterWithContains` ?

loictrichaud: What other condition ? Sorry I don't understand? :/
are you talking about `label.indexOf(pattern)` ? I don't know but which is for sure, only the "starts with" works. Currently, if you search for "A" in a list [Type A, Type B, Type C], the search does not work. You have to search for "Type A" to get a result.
We already have 4 customers who are against this. 
If the method had been "protected" I would have been able to correct the problem only from our side without disturbing you :/

DenysVuika: The "includes" is essentially doing the "contains" comparison, the "indexOf" is also checking for "contains", so the code logic can be read as:
```
if filterWithContains === true then check contains
else check contains
```

loictrichaud: image

DenysVuika: I think you have fixed the wrong condition branch, might be 
```
return this.queryBuilder.config.filterWithContains 
  ? label.includes(pattern)
  : label.startsWith(pattern);
```

DenysVuika: try to enable the `filterWithContains` in the config. I think you are trying to fix a non-existing problem

loictrichaud: OK, I understand ! I didn't add "filterWithContains" in "app.config.json" !  
Sorry :(  It's actually working.

DenysVuika: no worries, I'm glad it works for you

PR #5593, opened 2020-04-07
[ACA-2983] Changed the labels from the inputs name and type to always floating

ursedaniel:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://issues.alfresco.com/jira/browse/ACA-2983
The labels from creating a new process (process name and type) are in the same position as the place holder.

Added floating labels from the 2 inputs and switched the positions of the 2 inputs.

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/process-services/src/lib/process-list/components/start-process.component.scss
+++ b/lib/process-services/src/lib/process-list/components/start-process.component.scss
@@ -1,60 +1,61 @@
-.adf {
-    &-start-process {
-        width: 96%;
-        margin-left: auto;
-        margin-right: auto;
-        margin-top: 10px;
-        .mat-select-trigger {
-            font-size: 14px !important;
+@mixin adf-process-services-create-theme($theme) {
+    .adf {
+        &-start-process {
+            width: 96%;
+            margin-left: auto;
+            margin-right: auto;
+            margin-top: 10px;
+            .mat-select-trigger {
+                font-size: 14px !important;
+            }
+            mat-form-field {
+                width: 100%;
+            }
+            mat-select {
+                width: 100%;
+                padding: 16px 0 0;
+            }
+            .mat-form-field-label {
+                color: map_get($mat-grey,A400);
+            }
+            .mat-content-actions {
+                text-align: right;
+            }
+            .mat-button {
+                text-transform: uppercase !important;
+            }
         }
-        mat-form-field {
-            width: 100%;
+        &-title {
+            padding-bottom: 1.25em;
         }
-        mat-select {
-            width: 100%;
-            padding: 16px 0 0;
+        &-process-input-container {
+            mat-form-field {
+                width: 100%;
+            }
         }
-        .mat-form-field-label {
-            color: #bababa;
+        &-process-input-autocomplete {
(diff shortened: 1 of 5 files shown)


lib/process-services/src/lib/process-list/components/start-process.component.scss
```diff
@@ -14,13 +14,19 @@
             width: 100%;
             padding: 16px 0 0;
         }
+        .mat-form-field-label {
+            color: #bababa;
```
DenysVuika, line 18: this should be taken from the theme

lib/process-services/src/lib/process-list/components/start-process.component.scss
```diff
@@ -1,54 +1,61 @@
...
-            text-align: right;
-        }
-        .mat-button {
-            text-transform: uppercase !important;
-        }
-    }
-    &-process-input-container {
-        mat-form-field {
-            width: 100%;
+@mixin adf-process-services-create-theme($theme) {
```
DenysVuika, line 1: If you change the style to mixin, it should be called where other mixins are, and also component should not reference it in the styles attribute.

DenysVuika: check the _index.scss for examples

ursedaniel: sorry forgot to commit the file for import

DenysVuika: also, the .ts file should not contain the following line now:
```
styleUrls: ['./start-form.component.scss'],
```

lib/process-services/src/lib/process-list/components/start-process.component.scss
```diff
@@ -1,54 +1,61 @@
...
+            }
+            mat-form-field {
+                width: 100%;
+            }
+            mat-select {
+                width: 100%;
+                padding: 16px 0 0;
+            }
+            .mat-form-field-label {
+                color: map_get($mat-grey,A400);
```
DenysVuika, line 19: I think this is not correct, we are using "map-get" in other files, see 
https://github.com/Alfresco/alfresco-ng2-components/blob/development/lib/content-services/src/lib/breadcrumb/breadcrumb.component.scss#L18

lib/process-services/src/lib/process-list/components/start-process.component.scss
```diff
@@ -1,54 +1,61 @@
...
+    .adf {
+        &-start-process {
+            width: 96%;
+            margin-left: auto;
+            margin-right: auto;
+            margin-top: 10px;
+            .mat-select-trigger {
+                font-size: 14px !important;
+            }
+            mat-form-field {
```
eromano, line 11: do we need to overwrite all this mat- class?

ursedaniel: Can you be more explicit about what exact lines of code are you reffering to?

PR #5676, opened 2020-05-05
[ACA-3245] Fetch alfresco repo name from an API

arditdomi:

> - [ ] Bugfix
> - [X] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://issues.alfresco.com/jira/browse/ACA-3245

The repository name is fetched by an API and not from the application config

> - [ ] Yes
> - [X] No

**Other information**:

### DIFF
--- a/lib/core/form/services/activiti-alfresco.service.ts
+++ b/lib/core/form/services/activiti-alfresco.service.ts
@@ -55,10 +55,12 @@ export class ActivitiContentService {
     /**
      * Returns a list of all the repositories configured
      * @param includeAccounts
+     * @param tenantId
      */
-    getAlfrescoRepositories(includeAccounts: boolean): Observable<any> {
+    getAlfrescoRepositories(includeAccounts: boolean, tenantId?: number): Observable<any> {
         const apiService: AlfrescoApiCompatibility = this.apiService.getInstance();
         const opts = {
+            tenantId,
             includeAccounts
         };
         return from(apiService.activiti.alfrescoApi.getRepositories(opts))
--- a/lib/core/form/services/activiti-alfresco.service.ts
+++ b/lib/core/form/services/activiti-alfresco.service.ts
@@ -54,27 +54,12 @@ export class ActivitiContentService {
 
     /**
      * Returns a list of all the repositories configured
-     *
+     * @param includeAccounts
      */
-    getAlfrescoRepositories(): Observable<any> {
+    getAlfrescoRepositories(includeAccounts: boolean): Observable<any> {
         const apiService: AlfrescoApiCompatibility = this.apiService.getInstance();
         const opts = {
-            includeAccounts: false
-        };
-        return from(apiService.activiti.alfrescoApi.getRepositories(opts))
-            .pipe(
-                map(this.toJsonArray),
-                catchError((err) => this.handleError(err))
-            );
-    }
-
-    /**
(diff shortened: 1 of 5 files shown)


lib/core/form/services/activiti-alfresco.service.ts
```diff
@@ -55,14 +55,26 @@ export class ActivitiContentService {
     /**
      * Returns a list of all the repositories configured
      *
-     * @param accountId
-     * @param folderId
      */
-    getAlfrescoRepositories(tenantId: number, includeAccount: boolean): Observable<any> {
```
eromano, line 61: will be better leave the old method with the option, this is a breaking change and honestly have two method just to change a parameter I don't like is better have the possibility to pass it has option includeAccounts

lib/core/form/services/activiti-alfresco.service.ts
```diff
@@ -54,15 +54,12 @@ export class ActivitiContentService {
 
     /**
      * Returns a list of all the repositories configured
-     *
-     * @param accountId
-     * @param folderId
+     * @param includeAccounts
      */
-    getAlfrescoRepositories(tenantId: number, includeAccount: boolean): Observable<any> {
+    getAlfrescoRepositories(includeAccounts: boolean): Observable<any> {
```
eromano, line 59: tenantId is missing

arditdomi: I'm not sure if is needed. In all places that we use this function we pass tenantId = null. The api is returning the same if we don't pass anything as tenantId or if we pass only the one that the user is part of (if he is part of any tenant). The only place that tenantId could be useful is when you are an admin user and you can query by using tenantId but I think this is not a valid case in our apps.

arditdomi: I can re-add it it's just a thought. Thnx @eromano

mauriziovitale: it was not used. the UI was passing null

lib/core/form/services/activiti-alfresco.service.ts
```diff
@@ -54,15 +54,14 @@ export class ActivitiContentService {
...
     /**
      * Returns a list of all the repositories configured
-     *
-     * @param accountId
-     * @param folderId
+     * @param includeAccounts
+     * @param tenantId
      */
-    getAlfrescoRepositories(tenantId: number, includeAccount: boolean): Observable<any> {
+    getAlfrescoRepositories(includeAccounts: boolean, tenantId?: number): Observable<any> {
```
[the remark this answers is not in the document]
mauriziovitale: both of them needs to be optionals

lib/core/form/services/activiti-alfresco.service.ts
```diff
@@ -54,15 +54,14 @@ export class ActivitiContentService {
...
     /**
      * Returns a list of all the repositories configured
-     *
-     * @param accountId
-     * @param folderId
+     * @param tenantId
+     * @param includeAccounts
      */
-    getAlfrescoRepositories(tenantId: number, includeAccount: boolean): Observable<any> {
+    getAlfrescoRepositories(tenantId?: number, includeAccounts?: boolean): Observable<any> {
```
eromano, line 60: includeAccount no includeAccounts

PR #5710, opened 2020-05-18
[ACA-45] Drag and Drop a New Version

ursedaniel:

> - [ ] Bugfix
> - [x] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://issues.alfresco.com/jira/browse/ACA-45
Currently I can not drop a file over another file to update it's version, I can only drop a file into a folder to upload it there.

I can drag&drop a file over another file and the dialog for updating it's version will appear and after I specific the version(minor, major) and adding a comment (as optional) i can upload it's version right away.
Also I've added an option in the demo shell to enable/disable this feature.

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/demo-shell/src/app/components/files/files.component.ts
+++ b/demo-shell/src/app/components/files/files.component.ts
@@ -65,7 +65,6 @@ export class FilesComponent implements OnInit, OnChanges, OnDestroy {
     showViewer = false;
     showVersions = false;
     allowDropFiles = true;
-    allowDropFileVersion = true;
     displayMode = DisplayMode.List;
     includeFields = ['isFavorite', 'isLocked', 'aspectNames'];
 
@@ -238,10 +237,7 @@ export class FilesComponent implements OnInit, OnChanges, OnDestroy {
     }
 
     toggleAllowDropFiles() {
-        this.documentList.reload();
-    }
-
-    toggleAllowDropFileVersion() {
+        this.allowDropFiles = !this.allowDropFiles;
         this.documentList.reload();
     }
 
--- a/demo-shell/src/app/components/files/files.component.ts
+++ b/demo-shell/src/app/components/files/files.component.ts
@@ -65,6 +65,7 @@ export class FilesComponent implements OnInit, OnChanges, OnDestroy {
     showViewer = false;
     showVersions = false;
     allowDropFiles = true;
+    allowDropFileVersion = true;
     displayMode = DisplayMode.List;
     includeFields = ['isFavorite', 'isLocked', 'aspectNames'];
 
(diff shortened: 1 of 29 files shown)


demo-shell/src/app/components/files/files.component.ts
```diff
@@ -591,6 +596,27 @@ export class FilesComponent implements OnInit, OnChanges, OnDestroy {
         this.logService.log(event);
     }
 
+    onUploadNewVersion(ev) {
```
DenysVuika, line 599: does this mean every ADF application will have to have this code implemented on the app level?

lib/content-services/src/lib/upload/components/upload-version-button.component.ts
```diff
@@ -35,6 +35,10 @@ export class UploadVersionButtonComponent extends UploadButtonComponent implemen
     @Input()
     node: Node;
 
+    /** (**Required**) The drag&dropped new file to be versioned. */
+    @Input()
+    newFileVersion: File;
```
DenysVuika, line 40: why creating second property and then assigning it to the `file`? I suggest using base.file for that

lib/content-services/src/lib/upload/components/upload-version-button.component.ts
```diff
@@ -47,6 +51,7 @@ export class UploadVersionButtonComponent extends UploadButtonComponent implemen
     }
 
     ngOnInit() {
+        this.file = this.newFileVersion;
```
DenysVuika, line 54: having 2 properties in this case does not make sense, just use `file`

lib/core/datatable/components/datatable/datatable.component.html
```diff
@@ -60,7 +60,7 @@
                 (select)="onEnterKeyPressed(row, $event)"
                 (keyup)="onRowKeyUp(row, $event)"
                 (keydown)="onRowEnterKeyDown(row, $event)"
-                [adf-upload]="allowDropFiles && rowAllowsDrop(row)"
+                [adf-upload]="((isFolder(row) && allowDropFiles) || (!isFolder(row) && allowDropFileVersion)) && rowAllowsDrop(row)"
```
DenysVuika, line 63: the expression is too long, can you move it to a function and use a single function call in the html? like "canUpload()"

lib/core/datatable/components/datatable/datatable.component.ts
```diff
@@ -668,6 +679,10 @@ export class DataTableComponent implements AfterContentInit, OnChanges, DoCheck,
         return this.selectionMode && this.selectionMode.toLowerCase() === 'multiple';
     }
 
+    isFolder(row: DataRow): boolean {
```
DenysVuika, line 682: I think this is incorrect, the datatable component is a generic one, it should not have a notion of Folder.

lib/core/datatable/components/datatable/datatable.component.html
eromano, line 63: looks quite complicate assertion for a template maybe is better incapsulate in  a method

demo-shell/src/app/components/files/files.component.ts
ursedaniel: yes it does

lib/content-services/src/lib/upload/components/upload-version-button.component.ts
ursedaniel: you're right, changed that, my bad.

PR #5715, opened 2020-05-20
[ACA-2989]Creating data-table-component PO to support all kind of columns, including custom

cristinaj:
…om adf testing package]. Currently the majority of the methods that we have on this class are working only on 'text' column type. The scope of this PR was to make them work on all types of columns (text, image, date, fileSize, location, json) and also custom columns that may be created on different applications or other pages.

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

> - [ ] Yes
> - [ ] No

**Other information**:

### DIFF
--- a/lib/testing/src/lib/core/pages/data-table/column.ts
+++ b/lib/testing/src/lib/core/pages/data-table/column.ts
@@ -15,20 +15,17 @@
  * limitations under the License.
  */
 
-export * from './bread-crumb.page';
-export * from './header.page';
-export * from './user-info.page';
-export * from './login.page';
-export * from './settings.page';
-export * from './login-sso.page';
-export * from './data-table-component.page';
-export * from './pagination.page';
-export * from './error.page';
-export * from './notification-history.page';
-export * from './form/public-api';
-export * from './card-view/public-api';
-export * from './viewer.page';
-export * from './config-editor-page';
-export * from './about.page';
-export * from './snackbar.page';
-export * from './data-table/public-api';
+export abstract class Column {
+    columnName: string;
+    columnType: string;
+    locator: string;
+
+    constructor(columnName: string) {
+        this.columnName = columnName;
+    }
+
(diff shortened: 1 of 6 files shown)


lib/testing/src/lib/core/pages/data-table/column.ts
```diff
@@ -0,0 +1,31 @@
...
+export abstract class Column {
+    columnName: string;
+    columnType: string;
+    locator: string;
+
+    constructor(columnName: string) {
+        this.columnName = columnName;
+    }
+
+    // this will work for any type of columnType: 'text', 'image', 'date', 'fileSize', 'location', and 'json'.
```
eromano, line 27: this comment is not necessary

lib/testing/src/lib/core/pages/data-table/data-table-builder.ts
```diff
@@ -0,0 +1,44 @@
...
+                case 'date': {
+                    datatable.addItem(new DateColumn(column.columnName));
+                    break;
+                }
+                case 'custom': {
+                    datatable.addItem(column);
+                    break;
+                }
+                default: datatable.addItem(new TextColumn(column.columnName));
+                // here cases for 'image', 'fileSize', 'location', and 'json' can be added when needed
```
eromano, line 39: this comment is not needed

mauriziovitale:
@DenysVuika could you give us your though ?

mauriziovitale:
> For me looks ok as far we refactor completely and we don't keep two pages

I agree that we should not have 2 pages but is not feasible to change all the apps in one go. So I would suggest to merge the new code, start to using it on ADW and once we are 100% that we are able to cover all the different scenario we can create a story related to the refactor and change 1 app at time. @eromano @DenysVuika @cristinaj

DenysVuika:
Yeah, let’s add new code and remove old one in the second round. In that case we may have time to update all the apps without broken e2e pipeline?

PR #5741, opened 2020-05-29
[ACA-3390]Add startProcessWithApplication method on StartProcessPage

cristinaj:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

> - [ ] Yes
> - [ ] No

**Other information**:

### DIFF
--- a/lib/testing/src/lib/process-services/pages/start-process.page.ts
+++ b/lib/testing/src/lib/process-services/pages/start-process.page.ts
@@ -27,7 +27,6 @@ export class StartProcessPage {
     processNameInput: ElementFinder = element(by.id('processName'));
     disabledSelectProcessDropdown: ElementFinder = element(by.css('input[id="processDefinitionName"][disabled]'));
     selectProcessDropdownArrow: ElementFinder = element(by.css('button[id="adf-select-process-dropdown"]'));
-    selectApplicationDropdownArrow: ElementFinder = element(by.css('[data-automation-id*="start-process-app"] div[class="mat-select-arrow"]'));
     cancelProcessButton: ElementFinder = element(by.id('cancel_process'));
     formStartProcessButton: ElementFinder = element(by.css('button[data-automation-id="adf-form-start process"]'));
     startProcessButton: ElementFinder = element(by.css('button[data-automation-id="btn-start"]'));
@@ -36,6 +35,7 @@ export class StartProcessPage {
     processDefinitionOptionsPanel: ElementFinder = element(by.css('div[class*="mat-autocomplete-panel"]'));
 
     dropdownPage = new DropdownPage();
+    applicationDropdownPage = new DropdownPage(element(by.css('[data-automation-id*="start-process-app"] div[class="mat-select-arrow"]')));
 
     async checkNoProcessMessage(): Promise<void> {
         await BrowserVisibility.waitUntilElementIsVisible(this.noProcess);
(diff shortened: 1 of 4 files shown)


lib/testing/src/lib/process-services/pages/start-process.page.ts
```diff
@@ -145,9 +152,21 @@ export class StartProcessPage {
         return new FormFields();
     }
 
+    async checkSelectProcessField(): Promise<void> {
+        await BrowserVisibility.waitUntilElementIsNotVisible(this.disabledSelectProcessDropdown);
```
arditdomi, line 156: can we have a name like checkSelectProcessDropdownIsEnabled?

lib/testing/src/lib/process-services/pages/start-process.page.ts
```diff
@@ -34,6 +35,7 @@ export class StartProcessPage {
     processDefinitionOptionsPanel: ElementFinder = element(by.css('div[class*="mat-autocomplete-panel"]'));
 
     dropdownPage = new DropdownPage();
+    applicationDropdownPage = new DropdownPage(element(by.css('[data-automation-id*="start-process-app"] div[class="mat-select-arrow"]')));
```
arditdomi, line 38: just a small improvement. Now that we have 2 dropdown we could rename to processDefinitionDropdown and applicationDropdownPage. There is no need to wrap the method like we do here https://github.com/Alfresco/alfresco-ng2-components/blob/74aadb162ecc2a83c364f83be4030db550b022a9/lib/testing/src/lib/process-services/pages/start-process.page.ts#L90 instead we could directly use `this.applicationDropdownPage.selectOption` and `this.processDefinitionDropdown.selectOption`

lib/testing/src/lib/process-services/pages/start-process.page.ts
```diff
@@ -145,9 +152,21 @@ export class StartProcessPage {
         return new FormFields();
     }
 
+    async checkSelectProcessField(): Promise<void> {
+        await BrowserVisibility.waitUntilElementIsNotVisible(this.disabledSelectProcessDropdown);
+    }
+
     async startProcess({name, processName }) {
```
eromano, line 159: avoid use object in the firm of the method use two input parameters, we never use this kind of declaration

lib/testing/src/lib/process-services/pages/start-process.page.ts
```diff
@@ -145,9 +152,21 @@ export class StartProcessPage {
...
         await this.enterProcessName(name);
         await this.selectFromProcessDropdown(processName);
         await this.clickStartProcessButton();
     }
+
+    async startProcessWithApplication({name, applicationName, processName }) {
```
eromano, line 165: avoid use object in the firm of the method use three input parameters, we never use this kind of declaration

lib/testing/src/lib/process-services/pages/start-process.page.ts
cristinaj: This is how it was implemented previously. I've seen this method it's not much used, so i can refactor.

lib/testing/src/lib/process-services/pages/start-process.page.ts
cristinaj: Renamed

lib/testing/src/lib/process-services/pages/start-process.page.ts
cristinaj: I've renamed dropdownPage to processDefinitionDropdown and applicationDropdownPage to applicationDropdown. About the wrapping, it seems selectOption method is used in some tests too and there i wouldn't want to use the class member directly in the test.

PR #5747, opened 2020-06-02
[ACA_3292] - Add queued task as part of default task filter

SilviuCPopa:

> - [ ] Bugfix
> - [x] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://issues.alfresco.com/jira/browse/ACA-3292

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/process-services-cloud/src/lib/task/task-filters/services/task-filter-cloud.service.ts
+++ b/lib/process-services-cloud/src/lib/task/task-filters/services/task-filter-cloud.service.ts
@@ -277,6 +277,16 @@ export class TaskFilterCloudService {
                 order: 'DESC'
             }),
             new TaskFilterCloudModel({
+                name: 'ADF_CLOUD_TASK_FILTERS.QUEUED_TASKS',
+                key: 'queued-tasks',
+                icon: 'queue',
+                appName: appName,
+                status: 'CREATED',
+                assignee: '',
+                sort: 'createdDate',
+                order: 'DESC'
+            }),
+            new TaskFilterCloudModel({
                 name: 'ADF_CLOUD_TASK_FILTERS.COMPLETED_TASKS',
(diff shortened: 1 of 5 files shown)


lib/process-services-cloud/src/lib/task/task-filters/services/task-filter-cloud.service.ts
```diff
@@ -276,6 +276,16 @@ export class TaskFilterCloudService {
                 sort: 'createdDate',
                 order: 'DESC'
             }),
+            new TaskFilterCloudModel({
+                name: 'ADF_CLOUD_TASK_FILTERS.QUEUED_TASKS',
+                key: 'queued-tasks',
+                icon: 'queue',
+                appName: appName,
+                status: 'CREATED',
+                assignee: '',
```
mauriziovitale, line 285: I don't think this filter works as expected. A queued task isn't just a task that has assignee null, but a task that has candidates. It would be better to 
1 see how APS1  works 
2 create an example on cloud and make sure we return the valid tasks

erdemedeiros: @mauriziovitale, I was checking with @SilviuCPopa, and unless I'm missing something a task that is not assigned will always have candidates because in the modeler you cannot release a project for which a user task doesn't have `assignee` or `candidates` set. So for activiti-cloud, filtering on status `CREATED` should be enough.

popovicsandras: The other issue with a more complicated filter criteria is that the BE at the moment doesn't support more. In APS1 we have the request param: `assignment=candidate`, which we doesn't have in APS2.

PR #5903, opened 2020-07-23
[AAE-3084] Add retrieve metadata option in upload widget

pmartinezga:

> - [ ] Bugfix
> - [X] Feature
> - [ ] Build related changes
> - [ ] Documentation

There is no way to refresh the form values based on the value of another component

When the upload widget that is marked with the retrieve metadata flag changes its value, then the other form fields will be updated with the value of the properties that match the name of the property

> - [ ] Yes
> - [X] No

**Other information**:

### DIFF
--- a/lib/process-services-cloud/src/lib/i18n/en.json
+++ b/lib/process-services-cloud/src/lib/i18n/en.json
@@ -269,7 +269,7 @@
     "ERROR": {
       "INVALID_DESTINATION_FOLDER_PATH": "Invalid destination folder path"
     }
-  }
+  },
   "ADF_CLOUD_FORM_COMPONENT": {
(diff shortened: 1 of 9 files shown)


lib/process-services-cloud/src/lib/i18n/en.json
```diff
@@ -271,5 +270,7 @@
       "INVALID_DESTINATION_FOLDER_PATH": "Invalid destination folder path"
     }
   }
-
+  "ADF_CLOUD_FORM_COMPONENT": {
```
DenysVuika, line 273: coma is missing

pmartinezga: Done

PR #5936, opened 2020-07-30
[ACA-3762] Task/Process filters not working with updatePagination

mcchrys:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://issues.alfresco.com/jira/browse/ACA-3762

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/process-services-cloud/src/lib/process/process-list/components/process-list-cloud.component.ts
+++ b/lib/process-services-cloud/src/lib/process/process-list/components/process-list-cloud.component.ts
@@ -237,10 +237,23 @@ export class ProcessListCloudComponent extends DataTableSchema implements OnChan
         return !this.rows || this.rows.length === 0;
     }
 
+    resetPaginationValues(pagination?: PaginationModel) {
+        if (pagination) {
+            this.size = pagination.maxItems;
+            this.skipCount = pagination.skipCount;
+            this.pagination.next(pagination);
+        } else {
+            this.skipCount = 0;
+            this.size = this.userPreferences.paginationSize;
+            this.pagination.next({
+                skipCount: 0,
+                maxItems: this.size
+            });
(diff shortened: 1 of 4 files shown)


lib/process-services-cloud/src/lib/process/process-list/components/process-list-cloud.component.ts
```diff
@@ -233,10 +233,32 @@ export class ProcessListCloudComponent extends DataTableSchema implements OnChan
         return !this.rows || this.rows.length === 0;
     }
 
+    /**
+     * Resets the pagination values
+     * @param pagination Pagination values to be set (optional)
+     */
+    resetPaginationValues(pagination?: PaginationModel) {
```
SilviuCPopa, line 240: Can we use this function just to reset the pagination without taking any parameter? Also can we call it just resetPagination?

mcchrys: The optional parameter is necessary since this method is called from _updatePagination_ as well.
I think it's best to handle the reset cases within the same method and the name _resetPaginationValues_ sounds appropriate since we change only the values, unlike _updatePagination_ where we reload the list too.

SilviuCPopa: Can we let updatePagination unchanged? I mean, not to call resetPagination inside updatePagination function? 
From the name of the function, resetPagination should only reset the values. Now is also update them if we pass a parameter. It can be confusing

mcchrys: Okay. Updated.

PR #5944, opened 2020-08-03
[AAE-3209]Fix e2e about attach file from local

cristinaj:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

> - [ ] Yes
> - [ ] No

**Other information**:

### DIFF
--- a/e2e/process-services-cloud/process-task-attach-content-file-cloud.e2e.ts
+++ b/e2e/process-services-cloud/process-task-attach-content-file-cloud.e2e.ts
@@ -20,8 +20,6 @@ import {
     ApiService,
     AppListCloudPage,
     ContentNodeSelectorDialogPage,
-    IdentityService,
-    GroupIdentityService,
     LoginPage,
     ProcessCloudWidgetPage,
     ProcessDefinitionsService,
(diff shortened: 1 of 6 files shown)


e2e/process-services-cloud/process-task-attach-content-file-cloud.e2e.ts
```diff
@@ -55,7 +55,7 @@ describe('Process Task - Attach content file', () => {
     const processDefinitionName = browser.params.resources.ACTIVITI_CLOUD_APPS.SIMPLE_APP.processes.uploadSingleMultipleFiles;
     const uploadWidgetId = browser.params.resources.ACTIVITI_CLOUD_APPS.SIMPLE_APP.forms.uploadSingleMultiple.widgets.contentMultipleAttachFileId;
     const taskName = browser.params.resources.ACTIVITI_CLOUD_APPS.SIMPLE_APP.tasks.uploadSingleMultipleFiles;
-    const folderName = StringUtil.generateRandomString(5);
+    const folderName = '0' + StringUtil.generateRandomString(5);
```
eromano, line 58: why do we have this 0 at the beginning?

cristinaj: This test uses hruser and there are a lot of folders created[by other people, other tests, i don't know]. attachFileFromContentNode() method is checking if the content[folder in this case] is displayed, without taking in consideration to click on 'Show more' in case is displayed. So if the folder is not displayed in the first ones, the test will fail. The easiest way for me was to fix it like this. Other options would be to use a new created user instead of hruser or modify the attachFileFromContentNode() method to take in consideration cases where multiple folders are created.

eromano: I think is better to use a clean user

cristinaj: It seems i can't use clean user without changing the process definition. The user task has hruser as assignee. We can raise separate ticket for this.

Issue #5970, opened 2020-08-11
Custom Stencil doesn't work on Start Form


uvukasinovic:
**Type of issue:**  (check with "[x]")

> - [ ] New feature request
> - [x] Bug  
> - [ ] Support request
> - [ ] Documentation

**Current behaviour:**
Custom Stencil doesn't work with 3.9 version (it works with 3.7)

**Expected behavior:**
The stencil should be displayed in the Start Form

**Steps to reproduce the issue:**
Create a new process with a custom stencil field and start it in ADF

**Component name and version:** 
start-process.component.ts

<img width="1291" alt="Screen Shot 2020-08-10 at 12 55 16" src="https://user-images.githubusercontent.com/6500524/89872512-5afa3080-dbb9-11ea-81c9-a9634ac00c82.png">

dhrn:
@uvukasinovic have you tried this https://github.com/Alfresco/alfresco-ng2-components/issues/5678 ?

uvukasinovic:
Thanks @dhrn 

I tried #5678, but it doesn't work.

dhrn:
Raised jira ticket. https://issues.alfresco.com/jira/browse/ADF-5218

Thanks for reporting

PR #5999, opened 2020-08-14
[ADF-5218] Fix custom stencil usage on Start Process Form

davidcanonieto:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://issues.alfresco.com/jira/browse/ADF-5218

Custom stencils where overwritten buy the provider injection inside the Start Form and Task Form components.
This has been fixed and this injection comes from the module which allows custom stencils to overwrite it.

> - [ ] Yes
> - [x] No

**Other information**:
fixes https://github.com/Alfresco/alfresco-ng2-components/issues/5970


lib/process-services/src/lib/task-list/components/task-details.component.ts
```diff
@@ -43,11 +43,15 @@ import { TaskDetailsModel } from '../models/task-details.model';
...
 import { TaskFormComponent } from './task-form/task-form.component';
+import { ProcessFormRenderingService } from '../../form';
 
 @Component({
     selector: 'adf-task-details',
     templateUrl: './task-details.component.html',
-    styleUrls: ['./task-details.component.scss']
+    styleUrls: ['./task-details.component.scss'],
+    providers: [
+        { provide: FormRenderingService, useClass: ProcessFormRenderingService }
```
DenysVuika, line 53: this is going to be a problem, it will create new copies each time

demo-shell/src/app/components/files/files.component.ts
```diff
@@ -53,7 +54,10 @@ const DEFAULT_FOLDER_TO_SHOW = '-my-';
 @Component({
     selector: 'app-files-component',
     templateUrl: './files.component.html',
-    styleUrls: ['./files.component.scss']
+    styleUrls: ['./files.component.scss'],
```
davidcanonieto, line 57: @DenysVuika I had to add this provider to this component to make it work... This component is using the adf-start-form component and without this the widgets coming from ProcessFormRenderingService would not work, I don't think it's the correct implementation but I don't know what else we could do here.

I noticed that we're also injecting this provider in the Form Component in the demoshell for the same reason.

Any idea?

PR #6098, opened 2020-09-07
[ACA-3312] Add user-filter util class

iuliaib:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

- Add user-filter.util class that corresponds to [user-filters : Retrieve and manage user filters](http://adwdev.envalfresco.com/activiti-app/api-explorer.html#/user-filters) section in APS1.X.
- Add methods that **check if a filter is not highlighted** in order to avoid waiting seven seconds for `isFilterHighlighted()` to return `false`.

> - [ ] Yes
> - [ ] No

**Other information**:

### DIFF
--- a/lib/testing/src/lib/process-services/actions/user-filters.util.ts
+++ b/lib/testing/src/lib/process-services/actions/user-filters.util.ts
@@ -17,7 +17,7 @@
 
 import { Logger } from '../../core/utils/logger';
 import { ApiService } from '../../core/actions/api.service';
-import { UserProcessInstanceFilterRepresentation, UserFilterOrderRepresentation } from '@alfresco/js-api';
+import { UserFilterOrderRepresentation, UserTaskFilterRepresentation } from '@alfresco/js-api';
 
 export class UserFiltersUtil {
 
@@ -27,9 +27,9 @@ export class UserFiltersUtil {
         this.apiService = apiService;
     }
 
-    async createATaskFilter(newTaskFilterName: string, sortType?: string,  stateType?: string, assignmentType?: string, appId?: number, iconName?: string): Promise<any> {
+    async createATaskFilter(newTaskFilterName: string, sortType?: string,  stateType?: string, assignmentType?: string, iconName?: string, appId?: number): Promise<any> {
         try {
-            return this.apiService.getInstance().activiti.userFiltersApi.createUserTaskFilter(new UserProcessInstanceFilterRepresentation(
+            return this.apiService.getInstance().activiti.userFiltersApi.createUserTaskFilter(new UserTaskFilterRepresentation(
                 {appId: appId, name: newTaskFilterName, icon: iconName, filter: {sort: sortType, state: stateType, assignment: assignmentType}}));
         } catch (error) {
             Logger.error('Create Task Filter - Service error, Response: ', JSON.parse(JSON.stringify(error)));
@@ -52,6 +52,18 @@ export class UserFiltersUtil {
         }
     }
 
+    async getUserTaskFilterByName(taskFilterName: string, appId?: number): Promise<any> {
+        try {
+            const taskFiltersList = this.apiService.getInstance().activiti.userFiltersApi.getUserTaskFilters({appId: appId});
+            const chosenTaskFilter = (await taskFiltersList).data.find( (taskFilter) => {
+                return taskFilter.name === taskFilterName;
+            });
+            return chosenTaskFilter;
+        } catch (error) {
+            Logger.error('Get user task filters by name - Service error, Response: ', JSON.parse(JSON.stringify(error)));
+        }
+    }
+
     async deleteUserTaskFilter(filterId: number): Promise<any> {
         try {
             return this.apiService.getInstance().activiti.userFiltersApi.deleteUserTaskFilter(filterId);
@@ -59,4 +71,13 @@ export class UserFiltersUtil {
             Logger.error('Delete a task filter - Service error, Response: ', JSON.parse(JSON.stringify(error)));
         }
     }
+
+    async updateUserTaskFilter(filterId: number, updatedTaskFilterName?: string, updatedSortType?: string, updatedStateType?: string, updatedAssignmentType?: string, updatedIconName?: string, appId?: number): Promise<any> {
+        try {
+            return this.apiService.getInstance().activiti.userFiltersApi.updateUserTaskFilter(filterId, new UserTaskFilterRepresentation(
+                {appId: appId, name: updatedTaskFilterName, icon: updatedIconName, filter: {sort: updatedSortType, state: updatedStateType, assignment: updatedAssignmentType}}));
+        } catch (error) {
+            Logger.error('Delete a task filter - Service error, Response: ', JSON.parse(JSON.stringify(error))); }
+    }
+
 }
--- a/lib/testing/src/lib/process-services/actions/user-filters.util.ts
+++ b/lib/testing/src/lib/process-services/actions/user-filters.util.ts
@@ -1,0 +1,62 @@
+/*!
+ * @license
+ * Copyright 2019 Alfresco Software, Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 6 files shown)


lib/testing/src/lib/process-services/actions/user-filters.util.ts
```diff
@@ -0,0 +1,83 @@
...
+    constructor(apiService: ApiService) {
+        this.apiService = apiService;
+    }
+
+    async createATaskFilter(newTaskFilterName: string, sortType?: string,  stateType?: string, assignmentType?: string, iconName?: string, appId?: number): Promise<any> {
+        try {
+            return this.apiService.getInstance().activiti.userFiltersApi.createUserTaskFilter(new UserTaskFilterRepresentation(
+                {appId: appId, name: newTaskFilterName, icon: iconName, filter: {sort: sortType, state: stateType, assignment: assignmentType}}));
+        } catch (error) {
+            Logger.error('Create Task Filter - Service error, Response: ', JSON.parse(JSON.stringify(error)));
```
popovicsandras, line 35: In these cases, don't we want rethrow the error, beside displaying them? Like this, even if an error occurs, we just show a message, but the program's execution will continue. Is it exactly what we want here?

lib/testing/src/lib/process-services/actions/user-filters.util.ts
```diff
@@ -0,0 +1,83 @@
...
+        }
+    }
+
+    async orderUserTaskFilters(filtersIdOrder: number[], appId?: number): Promise<any> {
+        try {
+            return this.apiService.getInstance().activiti.userFiltersApi.orderUserTaskFilters(new UserFilterOrderRepresentation({appId: appId, order: filtersIdOrder}));
+        } catch (error) {
+            Logger.error('Re-order the list of user task filters - Service error, Response: ', JSON.parse(JSON.stringify(error)));
```
popovicsandras, line 43: Same here.

lib/testing/src/lib/process-services/actions/user-filters.util.ts
```diff
@@ -0,0 +1,83 @@
...
+        }
+    }
+
+    async getUserTaskFilters(appId?: number): Promise<any> {
+        try {
+            return this.apiService.getInstance().activiti.userFiltersApi.getUserTaskFilters({appId: appId});
+        } catch (error) {
+            Logger.error('List task filters - Service error, Response: ', JSON.parse(JSON.stringify(error)));
```
popovicsandras, line 51: Same here.

lib/testing/src/lib/process-services/actions/user-filters.util.ts
```diff
@@ -0,0 +1,83 @@
...
+
+    async getUserTaskFilterByName(taskFilterName: string, appId?: number): Promise<any> {
+        try {
+            const taskFiltersList = this.apiService.getInstance().activiti.userFiltersApi.getUserTaskFilters({appId: appId});
+            const chosenTaskFilter = (await taskFiltersList).data.find( (taskFilter) => {
+                return taskFilter.name === taskFilterName;
+            });
+            return chosenTaskFilter;
+        } catch (error) {
+            Logger.error('Get user task filters by name - Service error, Response: ', JSON.parse(JSON.stringify(error)));
```
popovicsandras, line 63: Same here.

lib/testing/src/lib/process-services/actions/user-filters.util.ts
```diff
@@ -0,0 +1,83 @@
...
+        }
+    }
+
+    async deleteUserTaskFilter(filterId: number): Promise<any> {
+        try {
+            return this.apiService.getInstance().activiti.userFiltersApi.deleteUserTaskFilter(filterId);
+        } catch (error) {
+            Logger.error('Delete a task filter - Service error, Response: ', JSON.parse(JSON.stringify(error)));
```
popovicsandras, line 71: Same here.

lib/testing/src/lib/process-services/actions/user-filters.util.ts
```diff
@@ -0,0 +1,83 @@
...
+        }
+    }
+
+    async updateUserTaskFilter(filterId: number, updatedTaskFilterName?: string, updatedSortType?: string, updatedStateType?: string, updatedAssignmentType?: string, updatedIconName?: string, appId?: number): Promise<any> {
+        try {
+            return this.apiService.getInstance().activiti.userFiltersApi.updateUserTaskFilter(filterId, new UserTaskFilterRepresentation(
+                {appId: appId, name: updatedTaskFilterName, icon: updatedIconName, filter: {sort: updatedSortType, state: updatedStateType, assignment: updatedAssignmentType}}));
+        } catch (error) {
+            Logger.error('Delete a task filter - Service error, Response: ', JSON.parse(JSON.stringify(error))); }
```
popovicsandras, line 80: Same here.

popovicsandras (changes requested):

lib/testing/src/lib/process-services/actions/user-filters.util.ts
DenysVuika: Indeed, the `catch` block will never be executed in case of the errors in the async `createUserTaskFilter`. I suggest not using try/catch here at all, as it's a test and we will see the error + stack trace anyway

iuliaib: @popovicsandras, @DenysVuika thanks for the comments. Regarding the `catch's logger.error`... I was writing this in the same idea of what we already have (eg. [lib/process-services/src/lib/process-list/services/process.service.ts](https://github.com/Alfresco/alfresco-ng2-components/blob/develop/lib/process-services/src/lib/process-list/services/process.service.ts)) and I was keeping it like this because I was wondering that if not, we will stop the execution of the tests and the entire class will fail, right? I'm not sure if I'm actually right.
Currently, in a class we have multiple tests, most of them with different preconditions [this is available at least for all the tests from adw->Process Services extension]. If one precondition (required for just one test) will fail, currently the tests will keep running. From my point of view, we could keep using this approach until we will find a way to avoid making an entire class fail because of one precondition needed by a single test. 🤔  

@DenysVuika I'm not sure if I've understood correctly the concern, but `createUserTaskFilter` is an API used directly from js-api package ([node_modules/@alfresco/js-api/src/api/activiti-rest-api/api/userFilters.api.ts](https://github.com/Alfresco/alfresco-js-api/blob/develop/src/api/activiti-rest-api/api/userFilters.api.ts#L73)). If something happens there, shouldn't be caught in `async createATaskFilter` `catch's Logger.error`? 🤔

lib/testing/src/lib/process-services/actions/user-filters.util.ts
mauriziovitale: JSON.parse(JSON.stringify(obj)) is a hack to deep clone an object I don't see any reason why we should have it here

mauriziovitale (changes requested):
check the comments

lib/testing/src/lib/process-services/actions/user-filters.util.ts
```diff
@@ -0,0 +1,83 @@
...
+            Logger.error('Delete a task filter - Service error, Response: ', error);
+        }
+    }
+
+    async updateUserTaskFilter(filterId: number, updatedTaskFilterName?: string, updatedSortType?: string, updatedStateType?: string, updatedAssignmentType?: string, updatedIconName?: string, appId?: number): Promise<any> {
+        try {
+            return this.apiService.getInstance().activiti.userFiltersApi.updateUserTaskFilter(filterId, new UserTaskFilterRepresentation(
+                {appId: appId, name: updatedTaskFilterName, icon: updatedIconName, filter: {sort: updatedSortType, state: updatedStateType, assignment: updatedAssignmentType}}));
+        } catch (error) {
+            Logger.error('Delete a task filter - Service error, Response: ', error);
```
mauriziovitale, line 80: you are not deleting you are updating

mauriziovitale (approved):

lib/testing/src/lib/process-services/actions/user-filters.util.ts
iuliaib: I've changed it. Thanks.

PR #6134, opened 2020-09-16
[AAE-3199] - Add support for searchable content model properties (new search filters panel in the attach file widget)

mauriziovitale:

> - [ ] Bugfix
> - [x] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://issues.alfresco.com/jira/browse/AAE-3199

A new panel with filters has been added in the attach file widget. The first part of the story is during modeling time. The modeler should be able to select which filters (content model properties) are going to be displayed for the end user so he can then easily search for them). This PR is addressing the end user part where the user should be able to filter using the properties defined by the modeler as shown in the attached image.
Screenshot 2020-09-28 at 2 59 56 pm

> - [ ] Yes
> - [ ] No

**Other information**:

### DIFF
--- a/lib/content-services/src/lib/content-node-selector/content-node-selector-panel.component.ts
+++ b/lib/content-services/src/lib/content-node-selector/content-node-selector-panel.component.ts
@@ -15,7 +15,7 @@
  * limitations under the License.
  */
 
-import { Component, EventEmitter, Input, OnInit, Output, ViewChild, ViewEncapsulation, OnDestroy, Inject } from '@angular/core';
+import { Component, EventEmitter, Input, OnInit, Output, ViewChild, ViewEncapsulation, OnDestroy } from '@angular/core';
 import {
     HighlightDirective,
     UserPreferencesService,
@@ -37,8 +37,7 @@ import { debounceTime, takeUntil, scan } from 'rxjs/operators';
 import { CustomResourcesService } from '../document-list/services/custom-resources.service';
 import { NodeEntryEvent, ShareDataRow } from '../document-list';
 import { Subject } from 'rxjs';
-import { SearchPanelQueryBuilderService } from '../search/search-panel-query-builder.service';
-import { SEARCH_QUERY_SERVICE_TOKEN } from '../search/search-query-service.token';
+import { SearchQueryBuilderService } from '../search';
 
 export type ValidationFunction = (entry: Node) => boolean;
 
@@ -236,8 +235,7 @@ export class ContentNodeSelectorPanelComponent implements OnInit, OnDestroy {
 
     constructor(private contentNodeSelectorService: ContentNodeSelectorService,
                 private customResourcesService: CustomResourcesService,
-                @Inject(SEARCH_QUERY_SERVICE_TOKEN)
-                private queryBuilderService: SearchPanelQueryBuilderService,
+                private queryBuilderService: SearchQueryBuilderService,
                 private userPreferencesService: UserPreferencesService,
                 private nodesApiService: NodesApiService,
                 private uploadService: UploadService,
@@ -261,6 +259,10 @@ export class ContentNodeSelectorPanelComponent implements OnInit, OnDestroy {
             )
             .subscribe(searchValue => this.search(searchValue));
 
+        this.contentNodeSelectorService.executed
+            .pipe(takeUntil(this.onDestroy$))
+            .subscribe(this.showSearchResults.bind(this));
+
         this.queryBuilderService.executed
             .pipe(takeUntil(this.onDestroy$))
             .subscribe(this.showSearchResults.bind(this));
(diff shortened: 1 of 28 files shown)


VitoAlbano:
I don't see the love, where are the ❤️ ?

lib/content-services/src/lib/content-node-selector/content-node-selector-panel.component.ts
```diff
@@ -37,6 +47,7 @@ import { debounceTime, takeUntil, scan } from 'rxjs/operators';
 import { CustomResourcesService } from '../document-list/services/custom-resources.service';
 import { NodeEntryEvent, ShareDataRow } from '../document-list';
 import { Subject } from 'rxjs';
+import { SEARCH_QUERY_SERVICE_TOKEN, SearchQueryBuilderService } from '../search';
```
mauriziovitale, line 50: can we import those from the specific files?

lib/content-services/src/lib/content-node-selector/content-node-selector.module.ts
```diff
@@ -24,10 +24,12 @@ import { ContentNodeSelectorPanelComponent } from './content-node-selector-panel
 import { ContentNodeSelectorComponent } from './content-node-selector.component';
 import { SitesDropdownModule } from '../site-dropdown/sites-dropdown.module';
 import { BreadcrumbModule } from '../breadcrumb/breadcrumb.module';
+import { SearchModule } from '../search/search.module';
 import { CoreModule } from '@alfresco/adf-core';
 import { DocumentListModule } from '../document-list/document-list.module';
 import { NameLocationCellComponent } from './name-location-cell/name-location-cell.component';
 import { UploadModule } from '../upload/upload.module';
+import { SearchQueryBuilderService } from '../search';
```
mauriziovitale, line 32: import not correct.

lib/core/services/search.service.ts
```diff
@@ -28,8 +28,8 @@ export class SearchService {
 
     dataLoaded: Subject<ResultSetPaging> = new Subject();
 
-    constructor(private apiService: AlfrescoApiService,
-                private searchConfigurationService: SearchConfigurationService) {
+    constructor(protected apiService: AlfrescoApiService,
```
mauriziovitale, line 31: do we need this change ?

lib/content-services/src/lib/content-node-selector/content-node-selector.module.ts
```diff
@@ -50,6 +53,7 @@ import { UploadModule } from '../upload/upload.module';
         ContentNodeSelectorPanelComponent,
         NameLocationCellComponent,
         ContentNodeSelectorComponent
-    ]
+    ],
+    providers: [SearchQueryBuilderService]
```
DenysVuika, line 57: is it needed? the service is provided in root

lib/content-services/src/lib/content-node-selector/content-node-selector.service.ts
```diff
@@ -67,15 +49,14 @@ export class ContentNodeSelectorService {
...
             },
             filterQueries: [
-                { query: `TYPE:'cm:folder'${ showFiles ? " OR TYPE:'cm:content'" : '' }` },
+                { query: `TYPE:'cm:folder'${showFiles ? " OR TYPE:'cm:content'" : ''}` },
                 { query: 'NOT cm:creator:System' },
                 ...parentFiltering
             ],
             scope: {
-                locations: ['nodes']
+                locations: 'nodes'
```
DenysVuika, line 57: why it's not an array?

lib/content-services/src/lib/search/components/search-panel/search-panel.component.scss
```diff
@@ -0,0 +1,3 @@
+.adf-search-panel {
```
DenysVuika, line 1: not needed

lib/content-services/src/lib/search/search-query-builder.service.ts
```diff
@@ -21,7 +21,7 @@ import { SearchConfiguration } from './search-configuration.interface';
 import { BaseQueryBuilderService } from './base-query-builder.service';
 
 @Injectable({
-    providedIn: 'root'
+    providedIn: 'any'
```
DenysVuika, line 24: is that on purpose?

DenysVuika (changes requested):

lib/content-services/src/lib/search/search-query-builder.service.ts
arditdomi: yes it is on purpose. We need a different service instance between the global search and the content node selector otherwise what you search on one it will remain as search term on the other

DenysVuika: Just remove the parameters then, and leave 
```
@Injectable()
```

lib/content-services/src/lib/search/components/search-panel/search-panel.component.scss
DenysVuika: I suggest removing this file and references to it, and adding only when we really need some new styles to be added

lib/content-services/src/lib/content-node-selector/content-node-selector.service.ts
arditdomi: https://github.com/Alfresco/alfresco-js-api/blob/develop/src/api/search-rest-api/model/requestScope.ts#L37 should be a string I think not an array

DenysVuika: thanks for clarifying

PR #6277, opened 2020-10-23
[AAE-3750] - add possibility to exclude user from selection on people…

SilviuCPopa:
… cloud component

> - [ ] Bugfix
> - [x] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://issues.alfresco.com/jira/browse/AAE-3750

> - [ ] Yes
> - [x] No

**Other information**:


lib/process-services-cloud/src/lib/people/components/people-cloud.component.ts
```diff
@@ -275,6 +281,13 @@ export class PeopleCloudComponent implements OnInit, OnChanges, OnDestroy {
         return false;
     }
 
+    private isExcludedUser(searchUser: IdentityUserModel): boolean {
+        if (this.excludedUsers && this.excludedUsers.length > 0) {
```
arditdomi, line 285: can we use `if (this.excludedUsers?.length > 0)` and declare a local variable isExcluded for the result?
Something like:

```
let isExcluded = false;
if (this.excludedUsers?.length > 0) {
isExcluded = !!this.excludedUsers.find(excludedUser => this.compare(excludedUser, searchUser));
}
return isExcluded;
```

it would make debugging easier

arditdomi (changes requested):

lib/process-services-cloud/src/lib/people/components/people-cloud.component.ts
SilviuCPopa: I'm not sure how will will make the debugging easier by creating a local variable.
Also I don't think using a local variable will make any sense in this function since this function has only one condition and the debugging purpose is not enough to add it in my opinion.

arditdomi: I personally think that having multiple exits on a function makes it more difficult to understand and debug but that's not a rule it is just my personal opinion. Please take a look at this article https://www.tomdalling.com/blog/coding-tips/coding-tip-have-a-single-exit-point/ and proceed as you wish

popovicsandras: Although I do agree, having a single exit point is considered to be a good practice with (especially complex) functions, which can help making the code easier to understand, however with such a simple function (which consists of 4 lines), I think the readability and understandability isn't damaged.

I agree with using the optional chaining, it is essentially the same, but less code:
```
this.excludedUsers?.length > 0
```

popovicsandras (changes requested):

arditdomi (approved):

popovicsandras (approved):

eromano (approved):

PR #6296, opened 2020-10-29
[ACA-4057]Add method to create process with values

cristinaj:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

> - [ ] Yes
> - [ ] No

**Other information**:


lib/testing/src/lib/process-services/actions/process.util.ts
```diff
@@ -43,6 +43,24 @@ export class ProcessUtil {
         }
     }
 
+    async startProcessByDefinitionNameWithValues(appName: string, processDefinitionName: string, values: any, processName?: string): Promise<any> {
```
eromano, line 46: can we just enhance the method above with an optional parameter?

eromano (changes requested):

lib/testing/src/lib/process-services-cloud/resources/resources.json
```diff
@@ -262,6 +262,50 @@
       "processchild": "processchild",
       "processparent": "processparent"
     },
+    "contentModels": {
+      "contentmodel1": {
+        "name": "contentmodel1",
```
eromano, line 267: will it be better to add some example use case name instead of cardinal numbers?

cristinaj: ok, i'll try to find better names

lib/testing/src/lib/process-services/actions/process.util.ts
cristinaj: As discussed with you, this will remain like this for now

eromano (approved):

PR #6348, opened 2020-11-14
[ADF-5284] Restore original plugin functionality

pionnegru:

> - [ ] Bugfix
> - [x] Feature
> - [ ] Build related changes
> - [ ] Documentation

[ADF-5284](https://alfresco.atlassian.net/browse/ADF-5284)

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/extensions/src/lib/services/extension-loader.service.ts
+++ b/lib/extensions/src/lib/services/extension-loader.service.ts
@@ -31,7 +31,7 @@ export class ExtensionLoaderService {
     constructor(private http: HttpClient) {
     }
 
-    load(configPath: string, pluginsPath: string, extensions?: ExtensionConfig[]): Promise<ExtensionConfig> {
+    load(configPath: string, pluginsPath: string, extensions?: string[]): Promise<ExtensionConfig> {
         return new Promise<any>((resolve) => {
             this.loadConfig(configPath, 0).then((result) => {
                 if (result) {
@@ -40,6 +40,10 @@ export class ExtensionLoaderService {
                     const override = sessionStorage.getItem('app.extension.config');
                     if (override) {
                         config = JSON.parse(override);
+                    }
+
+                    if (extensions && extensions.length) {
+                        config.$references.push(...this.filterIgnoredExtensions(extensions, config));
                     }
 
                     if (config.$references && config.$references.length > 0) {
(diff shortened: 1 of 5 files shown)


eromano (approved):

DenysVuika (approved):

lib/extensions/src/lib/services/extension-loader.service.ts
```diff
@@ -166,4 +166,12 @@ export class ExtensionLoaderService {
         }
         return action;
     }
+
+    private filterIgnoredExtensions(extensions: string[], config: ExtensionConfig): string[] {
```
popovicsandras, line 170: Why passing down the whole config object? I assume the $ignoreReferenceList would be enough, this is the only thing what we use from this object, as I see.

lib/extensions/src/lib/services/extension-loader.service.ts
```diff
@@ -42,6 +42,10 @@ export class ExtensionLoaderService {
                         config = JSON.parse(override);
                     }
 
+                    if (extensions && extensions.length) {
+                        config.$references.push(...this.filterIgnoredExtensions(extensions, config));
```
popovicsandras, line 46: If we push, there is no way to override the the order. As I see, here we need to do is something similar to this:\\\\\\\\\

```js
// No $references defined in the app.extensions.json, so we can continue the dynamically registered ones
// coming from the extensions variable, paying attention to the ignored ones as well
if (!config.$references || !config.$references.length) {
	config.$references = this.filterIgnoredExtensions(extensions, config);
}
// User defined their own array of $references, we should proceed with that, paying attention
// to the ignored ones as well
else {
	config.$references = this.filterIgnoredExtensions(config.$references, config);
}
```

For this, the application (ACA/ADW) should not have anything in their $references by default, which means, we need to modify even ACA to dynamically register its plugins either.

popovicsandras (changes requested):

lib/extensions/src/lib/services/extension-loader.service.ts
pionnegru: the  `push` is just to load the registered plugins.  Isn't the order set by the subsequent logic https://github.com/Alfresco/alfresco-ng2-components/pull/6348/files/a0fb40821839ddf565aa5f4b33e3cc885daac35f#diff-c6a1c872a637e729391870f0795194fb846e560b34090f242e98be038057977bL50-L54 ?

lib/extensions/src/lib/services/extension-loader.service.ts
pionnegru: ah, yes

popovicsandras (approved):

PR #6446, opened 2020-12-10
ACA-4175 Show number of user's home folders during environment scan

TomikaArome:

> - [ ] Bugfix
> - [x] Feature
> - [ ] Build related changes
> - [ ] Documentation

The environment scan shows only number of activated and deactivated users.

As well as the above information, the environment scan also show the number of user's home folders.

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/cli/scripts/scan-env.ts
+++ b/lib/cli/scripts/scan-env.ts
@@ -76,9 +76,7 @@ async function getPeopleCount(skipCount: number = 0): Promise<PeopleTally> {
 async function getHomeFoldersCount(): Promise<number> {
     try {
         const nodesApi = new NodesApi(jsApiConnection);
-        const rootApiResult = await nodesApi.listNodeChildren('-root-');
-        const userHomesFolderId = rootApiResult.list.entries.find(associationEntry => associationEntry.entry.name === 'User Homes').entry.id;
-        const homesFolderApiResult = await nodesApi.listNodeChildren(userHomesFolderId);
+        const homesFolderApiResult = await nodesApi.listNodeChildren('-root-', { relativePath: 'User Homes' });
         return homesFolderApiResult.list.pagination.totalItems;
     } catch (error) {
         console.log(error);
--- a/lib/cli/scripts/scan-env.ts
+++ b/lib/cli/scripts/scan-env.ts
@@ -1,4 +1,4 @@
-const { AlfrescoApiCompatibility, PeopleApi } = require('@alfresco/js-api');
+const { AlfrescoApiCompatibility, PeopleApi, NodesApi } = require('@alfresco/js-api');
 const program = require('commander');
 
 const MAX_ATTEMPTS = 10;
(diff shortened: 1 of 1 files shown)


andrea-ligios (approved):
LGTM!

lib/cli/scripts/scan-env.ts
```diff
@@ -69,6 +73,18 @@ async function getPeopleCount(skipCount: number = 0): Promise<PeopleTally> {
     }
 }
 
+async function getHomeFoldersCount(): Promise<number> {
+    try {
+        const nodesApi = new NodesApi(jsApiConnection);
+        const rootApiResult = await nodesApi.listNodeChildren('-root-');
+        const userHomesFolderId = rootApiResult.list.entries.find(associationEntry => associationEntry.entry.name === 'User Homes').entry.id;
+        const homesFolderApiResult = await nodesApi.listNodeChildren(userHomesFolderId);
```
andrea-ligios, line 81: On second thought, let's chain the API usage for a more fluent approach. 
You can take [this](https://github.com/Alfresco/alfresco-digital-workspace-app/blob/develop/scripts%2Fci%2Fcheck-env%2Fcleanup-gs-env.js#L67-L73) as an example. 

Hence instead of a style like:
```
const peopleApi = new PeopleApi(jsApiConnection);
const apiResult = await peopleApi.listPeople({
```

it's better to use something chained like for example: 

```
const groups = await alfrescoJsApi.core.groupsApi.getGroups(opt);
```

andrea-ligios (changes requested):

lib/cli/scripts/scan-env.ts
DenysVuika: All the `alfrescoJsApi` members will be deprecated in the future (because it prevents proper tree-shaking and optimizations), the first usage is more accurate. I suggest to avoid using any `alfrescoJsApi.XXX` methods for now

andrea-ligios: Good to know @DenysVuika ! Let's go with the former one then, @TomikaArome

andrea-ligios (approved):
LGTM

lib/cli/scripts/scan-env.ts
```diff
@@ -69,6 +73,18 @@ async function getPeopleCount(skipCount: number = 0): Promise<PeopleTally> {
     }
 }
 
+async function getHomeFoldersCount(): Promise<number> {
+    try {
+        const nodesApi = new NodesApi(jsApiConnection);
+        const rootApiResult = await nodesApi.listNodeChildren('-root-');
```
mauriziovitale, line 79: are we sure that there is no way to pass a relative path such as 
`listNodeChildren('-root-', {
            'relativePath': 'User Homes',
            'maxItems': 500
        });`

mauriziovitale: In addition to that keep in mind that the api will return the matItems so you have to iterate to get all of them!

TomikaArome: Oh yes you're correct, I had missed that thanks. Updating now

mauriziovitale (changes requested):
check the comments

lib/cli/scripts/scan-env.ts
mauriziovitale: can we please have the User Homes as a constant

TomikaArome: Only saw this after committing. Updating again

mauriziovitale (approved):

PR #6599, opened 2021-01-29
[AAE-3992] Process Filter bug fixes and improvements

DenysVuika:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

With this change, it is now possible to fully restore the state of the process filter from within the application after page reloads. This is much needed for the "Admin App" and similar, when filters are stored in the URL as query parameters and need to be restored if following the link or reloading the page.

- fix: old bug with Cloud components and the need to provide `entry.key` for all the table presets (this was a very painful bug for the APPS, now should have clean implementation)
- fix: issue with filter not able to initialize properly after page reload
- fix: issue with the "save as" not working if form is not changed
- fix: issue with "save" not working in some cases
- fix: issue with the multiple app versions picker when restoring state
- fix: non-translated text in the filter UI
- lots of other bug fixes for the Process Filter component
- remove unit tests that make no sense after refactoring
- remove dead code in e2e pages classes

Screenshot 2021-01-29 at 15 35 04

> - [ ] Yes
> - [x] No

**Other information**:
https://alfresco.atlassian.net/browse/AAE-3992


mauriziovitale (approved):

lib/process-services-cloud/src/lib/process/process-filters/components/edit-process-filter-cloud.component.spec.ts
```diff
@@ -211,112 +210,19 @@ describe('EditProcessFilterCloudComponent', () => {
...
-            it('should disable saveAs button if the process filter is not changed for custom filter', async(() => {
-                component.toggleFilterActions = true;
-                const expansionPanel = fixture.debugElement.nativeElement.querySelector('mat-expansion-panel-header');
-                expansionPanel.click();
-                fixture.detectChanges();
-                fixture.whenStable().then(() => {
-                    const saveButton = fixture.debugElement.nativeElement.querySelector('[data-automation-id="adf-filter-action-saveAs"]');
-                    expect(saveButton.disabled).toEqual(true);
-                });
-            }));
```
arditdomi, line 318: The unit tests that were removed were redundant/duplicated/not needed/covered by e2e?

DenysVuika: Not needed or redundant, yes. For example, "save as" should not depend on the dirty state of the form, I remove that linking and some tests for that become irrelevant. Same for other areas. There are still a few useless tests in the e2e (or duplicates), but I will address that later I hope.

arditdomi: Thnx 👍

arditdomi (approved):

davidcanonieto (approved):

e2e/process-services-cloud/edit-process-filters-component.e2e.ts
```diff
@@ -81,10 +81,6 @@ describe('Edit process filters cloud', () => {
         await editProcessFilter.checkSaveButtonIsDisplayed();
         await editProcessFilter.checkSaveAsButtonIsDisplayed();
         await editProcessFilter.checkDeleteButtonIsDisplayed();
-        await expect(await editProcessFilter.checkSaveButtonIsEnabled()).toEqual(false);
-        await expect(await editProcessFilter.checkSaveAsButtonIsEnabled()).toEqual(false);
-        await expect(await editProcessFilter.checkDeleteButtonIsEnabled()).toEqual(false);
```
arditdomi, line 86: is it necessary to remove those checks that check if the buttons are enabled/disabled? Also should we leave the test without any expectation? I know that those methods check the visibility of the button and if it is not there it will fail but if let's say someone refactors the method `checkDeleteButtonIsDisplayed ` in the future to not fail but to catch the error and return a value true/fails. Could it lead to an ever-green test

lib/process-services-cloud/src/lib/process/process-filters/services/process-filter-cloud.service.ts
```diff
@@ -102,14 +153,20 @@ export class ProcessFilterCloudService {
...
+        return this.getProcessFiltersByKey(appName, key).pipe(
             switchMap((filters: ProcessFilterCloudModel[]) => {
                 if (filters && filters.length === 0) {
-                    return this.createProcessFilters(newFilter.appName, key, [newFilter]);
+                    return this.createProcessFilters(appName, key, [newFilter]);
                 } else {
+                    const existing = filters.find(filter => filter.name === name);
+                    if (existing) {
+                        filters.splice(filters.indexOf(existing), 1);
+                    }
```
arditdomi, line 167: Could we use something like this to achieve the same result? the .find and .indexOf will iterate 2 times the filters Array. One to find the element and one to find the index. Instead I think this way is a little bit more efficient. What do you think?
```
const index = filters.findIndex(filter => filter.name === name);
if (index !== -1) {
filters.splice(index, 1);
}
```

lib/process-services-cloud/src/lib/process/process-list/services/process-list-cloud.service.ts
```diff
@@ -33,29 +34,40 @@ export class ProcessListCloudService extends BaseCloudService {
...
+                })
+            );
         } else {
             this.logService.error('Appname is mandatory for querying task');
             return throwError('Appname not configured');
         }
     }
 
-    private isPropertyValueValid(requestNode: any, property: string) {
+    protected isPropertyValueValid(requestNode: any, property: string) {
```
arditdomi, line 66: missing return type

arditdomi (changes requested):

e2e/process-services-cloud/edit-process-filters-component.e2e.ts
DenysVuika: My concern is that the test is already doing lots of stuff that is not in the test name... if someone changes something in the future, that someone will need to write a proper test? :)

lib/process-services-cloud/src/lib/process/process-filters/services/process-filter-cloud.service.ts
DenysVuika: makes sense, thanks!

PR #6626, opened 2021-02-04
[ATS-854] Add media tracks to player from webvtt rendition

pmartinezga:

> - [ ] Bugfix
> - [X] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://alfresco.atlassian.net/browse/ATS-854

https://alfresco.atlassian.net/browse/ATS-854

> - [ ] Yes
> - [X] No

**Other information**:

### DIFF
--- a/lib/core/viewer/components/viewer.component.ts
+++ b/lib/core/viewer/components/viewer.component.ts
@@ -32,6 +32,7 @@ import { Subscription } from 'rxjs';
 import { ViewUtilService } from '../services/view-util.service';
 import { AppExtensionService, ViewerExtensionRef } from '@alfresco/adf-extensions';
 import { filter } from 'rxjs/operators';
+import { Track } from '../models/viewer.model';
 
 @Component({
     selector: 'adf-viewer',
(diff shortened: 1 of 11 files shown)


lib/core/viewer/components/viewer.component.ts
```diff
@@ -416,6 +419,10 @@ export class ViewerComponent implements OnChanges, OnInit, OnDestroy {
             }
         }
 
+        if (this.viewerType === 'media') {
+            this.mediaTracks = await this.viewUtils.generateMediaTracks(nodeData.id);
```
eromano, line 423: should this thing be in adf-media-player ?

eromano (changes requested):

lib/core/viewer/components/viewer.component.ts
pmartinezga: We can do it here or in the media player. I decided to do it here and keep media without any logic to follow the same approach that has been done with the source url, that is not calculated in media player but here in the viewer.

I can do it as you prefer, @eromano media or viewer?

eromano: to me looks a media player concern and you are right in this class there are other concern to move

eromano (approved):

PR #6664, opened 2021-02-11
[AAE-4429] Add readOnly breadcrumb to upload from device tab

TomikaArome:

> - [ ] Bugfix
> - [x] Feature
> - [ ] Build related changes
> - [ ] Documentation

image

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/content-services/src/lib/content-node-selector/content-node-selector-panel.component.ts
+++ b/lib/content-services/src/lib/content-node-selector/content-node-selector-panel.component.ts
@@ -223,6 +223,10 @@ export class ContentNodeSelectorPanelComponent implements OnInit, OnDestroy {
     @Output()
     folderLoaded: EventEmitter<any> = new EventEmitter<any>();
 
+    /** Emitted when the breadcrumb folder title has changed */
+    @Output()
+    breadcrumbFolderTitleEvent = new EventEmitter<string>();
+
     @ViewChild('documentList', { static: true })
     documentList: DocumentListComponent;
 
@@ -239,9 +243,6 @@ export class ContentNodeSelectorPanelComponent implements OnInit, OnDestroy {
     _chosenNode: Node [] = null;
     folderIdToShow: string | null = null;
     startSiteGuid: string | null = null;
-
-    @Output()
-    breadcrumbFolderTitleEvent = new EventEmitter<string | null>();
     breadcrumbFolderTitle: string | null = null;
 
     @ViewChild(InfinitePaginationComponent, { static: true })
--- a/lib/content-services/src/lib/content-node-selector/content-node-selector-panel.component.ts
+++ b/lib/content-services/src/lib/content-node-selector/content-node-selector-panel.component.ts
@@ -238,8 +238,11 @@ export class ContentNodeSelectorPanelComponent implements OnInit, OnDestroy {
     inDialog: boolean = false;
     _chosenNode: Node [] = null;
     folderIdToShow: string | null = null;
-    breadcrumbFolderTitle: string | null = null;
     startSiteGuid: string | null = null;
+
+    @Output()
+    breadcrumbFolderTitleEvent = new EventEmitter<string | null>();
+    breadcrumbFolderTitle: string | null = null;
 
     @ViewChild(InfinitePaginationComponent, { static: true })
     infinitePaginationComponent: InfinitePaginationComponent;
@@ -533,6 +536,7 @@ export class ContentNodeSelectorPanelComponent implements OnInit, OnDestroy {
         this.showingSearchResults = false;
         this.infiniteScroll = false;
         this.breadcrumbFolderTitle = null;
+        this.breadcrumbFolderTitleEvent.emit(this.breadcrumbFolderTitle);
         this.preselectedNodes = [];
         this.clearSearch();
         this.navigationChange.emit($event);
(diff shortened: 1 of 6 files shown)


lib/content-services/src/lib/content-node-selector/content-node-selector-panel.component.ts
```diff
@@ -238,9 +238,12 @@ export class ContentNodeSelectorPanelComponent implements OnInit, OnDestroy {
     inDialog: boolean = false;
     _chosenNode: Node [] = null;
     folderIdToShow: string | null = null;
-    breadcrumbFolderTitle: string | null = null;
     startSiteGuid: string | null = null;
 
+    @Output()
+    breadcrumbFolderTitleEvent = new EventEmitter<string | null>();
```
arditdomi, line 244: Do we need the `| null`? Also can you leave an empty line between the event emitter and the next variable declaration? It would be nice to move this on top of the class under all the other @Outputs , usually all the Inputs and Outputs are grouped together in our components which I think has it's pros

lib/content-services/src/lib/content-node-selector/content-node-selector-panel.component.ts
```diff
@@ -533,6 +536,7 @@ export class ContentNodeSelectorPanelComponent implements OnInit, OnDestroy {
         this.showingSearchResults = false;
         this.infiniteScroll = false;
         this.breadcrumbFolderTitle = null;
+        this.breadcrumbFolderTitleEvent.emit(this.breadcrumbFolderTitle);
```
arditdomi, line 539: we could add a unit test that checks: When the folder is changed the new event emitter emits the value that it should, in this case the folder title

lib/content-services/src/lib/content-node-selector/content-node-selector-panel.component.ts
```diff
@@ -608,6 +612,7 @@ export class ContentNodeSelectorPanelComponent implements OnInit, OnDestroy {
         } else {
             this.breadcrumbFolderTitle = null;
         }
+        this.breadcrumbFolderTitleEvent.emit(this.breadcrumbFolderTitle);
```
arditdomi, line 615: same here we could add a similar unit test

arditdomi (changes requested):

lib/content-services/src/lib/content-node-selector/content-node-selector-panel.component.ts
TomikaArome: What happens if breadcrumbFolderTitle is null? The original property has type `string | null`, implying there would be a chance we'd emit null value

arditdomi: theoretically every Event emitter could emit null if something goes wrong. To be honest not sure if it's needed as a type declaration. @mauriziovitale any idea?

lib/content-services/src/lib/content-node-selector/content-node-selector.component.spec.ts
```diff
@@ -393,4 +393,32 @@ describe('ContentNodeSelectorComponent', () => {
...
+
+    describe('Breadcrumb', () => {
+        let contentNodePanel;
+
+        beforeEach(() => {
+            contentNodePanel = fixture.debugElement.query(By.directive(ContentNodeSelectorPanelComponent));
+            spyOn(component, 'onBreadcrumbFolderTitleEvent');
+        });
+
+        it('should emit a breadcrumbFolderTitleEvent when the folder is changed', () => {
```
arditdomi, line 405: The event emitter is part of the content-node-selector-panel component and the unit tests should be part of that component and not part of the content-node-selector component. In the panel component you can spy on the event emitter to check what is emitted

lib/content-services/src/lib/content-node-selector/content-node-selector.component.spec.ts
```diff
@@ -393,4 +393,32 @@ describe('ContentNodeSelectorComponent', () => {
...
+            const fakeNodeEntry = new Node({ id: 'fakeId' });
+            const nodeEntryEvent = new NodeEntryEvent(fakeNodeEntry);
+
+            contentNodePanel.componentInstance.onFolderChange(nodeEntryEvent);
+            fixture.detectChanges();
+
+            expect(component.onBreadcrumbFolderTitleEvent).toHaveBeenCalled();
```
arditdomi, line 412: I suggest you use toHaveBeenCalledWith(something) instead of checking that it was just called

arditdomi (changes requested):

lib/content-services/src/lib/content-node-selector/content-node-selector-panel.component.ts
```diff
@@ -223,6 +223,10 @@ export class ContentNodeSelectorPanelComponent implements OnInit, OnDestroy {
     @Output()
     folderLoaded: EventEmitter<any> = new EventEmitter<any>();
 
+    /** Emitted when the breadcrumb folder title has changed */
+    @Output()
+    breadcrumbFolderTitleEvent = new EventEmitter<string>();
```
eromano, line 228: This event has few wrong things IMO:
1. The name we never call it Event at the end. We try to not use suffix. 
2. The name is not really an indicator of what happened in the component. If the Title change is because we change folder or also for other reason?

eromano (changes requested):

lib/content-services/src/lib/content-node-selector/content-node-selector-panel.component.ts
TomikaArome: Thanks for confirming, this has just been pointed out to me. Just about to update to remove the event.

arditdomi (approved):

mauriziovitale (approved):

PR #6731, opened 2021-02-25
[AAE-3543] Attach button enabled only when all files uploaded

TomikaArome:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

The "Apply" button in the attach file dialog is always enabled as long as a file is selected, even if there are still files being uploaded.

As long as the queue of files hasn't finished uploading, the "Apply" button will be disabled.

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/core/services/upload.service.spec.ts
+++ b/lib/core/services/upload.service.spec.ts
@@ -105,6 +105,34 @@ describe('UploadService', () => {
         expect(service.getQueue().length).toEqual(2);
     });
 
+    it('should have the queue finished uploading if all files are complete, cancelled, aborted, errored or deleted', () => {
+        const file1 = new FileModel(<File> { name: 'fake-file-1', size: 10 });
+        const file2 = new FileModel(<File> { name: 'fake-file-2', size: 20 });
+        const file3 = new FileModel(<File> { name: 'fake-file-3', size: 30 });
+        const file4 = new FileModel(<File> { name: 'fake-file-4', size: 40 });
+        const file5 = new FileModel(<File> { name: 'fake-file-5', size: 50 });
+        file1.status = FileUploadStatus.Complete;
+        file2.status = FileUploadStatus.Cancelled;
+        file3.status = FileUploadStatus.Aborted;
+        file4.status = FileUploadStatus.Error;
+        file5.status = FileUploadStatus.Deleted;
+        service.addToQueue(file1, file2, file3, file4, file5);
+        expect(service.isQueueFinishedUploading()).toBe(true);
+    });
+
+    it('should not have the queue finished uploading if some files are still pending, starting, in progress or uploading', () => {
+        const file1 = new FileModel(<File> { name: 'fake-file-1', size: 10 });
+        const file2 = new FileModel(<File> { name: 'fake-file-2', size: 20 });
+        service.addToQueue(file1, file2);
+        file1.status = FileUploadStatus.Complete;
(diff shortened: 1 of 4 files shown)


lib/core/services/upload.service.ts
```diff
@@ -95,6 +95,13 @@ export class UploadService {
         return !!this.activeTask;
     }
 
+    isQueueFinishedUploading(): boolean {
+        const finishedFileStates = [FileUploadStatus.Complete, FileUploadStatus.Cancelled, FileUploadStatus.Aborted, FileUploadStatus.Error, FileUploadStatus.Deleted];
+        return this.queue.reduce((finishedUploading: boolean, currentFile: FileModel) => {
+            return finishedUploading && finishedFileStates.indexOf(currentFile.status) > -1;
+        }, true);
+    }
+
```
arditdomi, line 104: there is already a method called isUploading() and I think there is no need to create a new one

arditdomi (dismissed):

lib/core/services/upload.service.ts
TomikaArome: I'd agree, but using that method intermittently causes the button to be reenabled in between one upload having finished and the next one starting. This method takes into consideration the Pending and Starting states too. Basically it's a method to check if the queue still has things left to do or if it's idle, I wouldn't know how better to name it though.

arditdomi: thnx

mauriziovitale: just thinking if we really need the first case and why.
probably the isUploading is hiding bugs
can u try to use u code and run the unit test?

mauriziovitale (changes requested):
check what breaks and why

arditdomi (approved):

lib/core/services/upload.service.spec.ts
```diff
@@ -105,6 +105,34 @@ describe('UploadService', () => {
...
+        const file1 = new FileModel(<File> { name: 'fake-file-1', size: 10 });
+        const file2 = new FileModel(<File> { name: 'fake-file-2', size: 20 });
+        service.addToQueue(file1, file2);
+        file1.status = FileUploadStatus.Complete;
+        file2.status = FileUploadStatus.Pending;
+        expect(service.isUploading()).toBe(true);
+        file2.status = FileUploadStatus.Starting;
+        expect(service.isUploading()).toBe(true);
+        file2.status = FileUploadStatus.Progress;
+        expect(service.isUploading()).toBe(true);
```
arditdomi, line 133: Just a small thing. Please check `Rule Style 05-14` of this page https://alfresco.atlassian.net/wiki/spaces/AP/pages/471060/Code+Style+E2E

arditdomi (changes requested):

lib/core/services/upload.service.spec.ts
TomikaArome: Sure, but if I were to do it this way, I'd have a new line every line, or I would have (in this case) lines 131 and 132 grouped with new lines before and after. If we want new lines, wouldn't it still be better to group the assignment of the status with the expect that follows it?

arditdomi: as you wish, both ways are ok and reading it you can easily identify the expectations. I just want to avoid the one block test

arditdomi (approved):

mauriziovitale (approved):

PR #6738, opened 2021-02-26
[AAE-4697] Restrict content node selector breadcrumb when destination folder path is wrong

sivakumar414ram:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://alfresco.atlassian.net/browse/AAE-4697

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/process-services-cloud/src/lib/form/components/widgets/attach-file/attach-file-cloud-widget.component.ts
+++ b/lib/process-services-cloud/src/lib/form/components/widgets/attach-file/attach-file-cloud-widget.component.ts
@@ -152,14 +152,16 @@ export class AttachFileCloudWidgetComponent extends UploadCloudWidgetComponent i
 
         if (!rootNodeId) {
             let nodeId: string;
-            if (destinationFolderPath.path) {
-                nodeId = await this.contentNodeSelectorService.fetchNodeIdFromRelativePath(destinationFolderPath.alias, { relativePath: destinationFolderPath.path });
+            try {
+                if (destinationFolderPath.path) {
+                    nodeId = await this.contentNodeSelectorService.fetchNodeIdFromRelativePath(destinationFolderPath.alias, { relativePath: destinationFolderPath.path });
+                }
+                if (!nodeId) {
+                    nodeId = await this.contentNodeSelectorService.fetchAliasNodeId(destinationFolderPath.alias);
+                }
+            } catch (error) {
+                this.logService.error(error);
             }
-
-            if (!nodeId) {
-                nodeId = await this.contentNodeSelectorService.fetchAliasNodeId(destinationFolderPath.alias);
-            }
-
             rootNodeId = nodeId ? nodeId : destinationFolderPath.alias;
         }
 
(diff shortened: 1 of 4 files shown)


mauriziovitale (approved):

lib/process-services-cloud/src/lib/form/services/content-cloud-node-selector.service.ts
```diff
@@ -54,11 +54,17 @@ export class ContentCloudNodeSelectorService {
...
-            .getNode(alias, opts)
-            .catch((err) => this.handleError(err));
+        const relativePathNodeEntry = await this.fetchNodeId(alias, opts);
+        return relativePathNodeEntry?.entry?.id;
+    }
+
+    async fetchAliasNodeId(alias: string): Promise<string> {
+        const aliasNodeEntry = await this.fetchNodeId(alias);
+        return aliasNodeEntry?.entry?.id;
+    }
```
arditdomi, line 64: fetchNodeIdFromRelativePath and fetchAliasNodeId are the same methods, the only difference is the opts parameter which you can make optional and keep just one. What do you think?

arditdomi: Maybe having has it's benefits as from the names you understand the flow

sivakumar414ram: yup, In the first place i had single method but it's little hard to understand the flow. May we can have more generic name so that we can have optional param.

lib/process-services-cloud/src/lib/form/components/widgets/attach-file/attach-file-cloud-widget.component.ts
```diff
@@ -152,14 +152,16 @@ export class AttachFileCloudWidgetComponent extends UploadCloudWidgetComponent i
 
         if (!rootNodeId) {
             let nodeId: string;
```
eromano, line 154: can we move what is inside this if in a separate method? try catch block should leave as good practice in a method and not find it in the middle of the code of your method. This is because try-catch is a concern and for the single responsibility principle any method should do only one thing

eromano (changes requested):

lib/process-services-cloud/src/lib/form/components/widgets/attach-file/attach-file-cloud-widget.component.ts
sivakumar414ram: Updated. Thanks

lib/process-services-cloud/src/lib/form/services/content-cloud-node-selector.service.ts
sivakumar414ram: I would like to have two method to keep flow understandable. Please do suggest if you have a better name. Thanks

arditdomi:
There is a problem not related to this issue regarding the affected libs. As you see in the travis run no e2e were executed. The following PR contains a fix so we should wait until it gets merged https://github.com/Alfresco/alfresco-ng2-components/pull/6842/files

arditdomi (approved):

PR #6755, opened 2021-03-01
[ACA-4308] More flexible content-node-selector-dialog-content height to avoid scrollbar

TomikaArome:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

Scrollbar appears when height of content-node-selector-dialog-content is greater than its css defined value.

Changed to min-height to allow for more flexible content. Addition of a dynamicHeight property to smoothen out the change of height between tabs in case both heights are not the same.

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/content-services/src/lib/content-node-selector/content-node-selector.component.html
+++ b/lib/content-services/src/lib/content-node-selector/content-node-selector.component.html
@@ -7,7 +7,8 @@
 <mat-tab-group class="adf-content-node-selector-dialog-content"
                mat-align-tabs="start"
                (selectedIndexChange)="onTabSelectionChange($event)"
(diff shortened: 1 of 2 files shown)


andrea-ligios (approved):
LGTM, as long as it works with the resolutions of modern smartphones

arditdomi (changes requested):
there is one problem with this implementation, the second tab will be always growing the more files you upload and it will be different from the height the first tab will have
Screenshot 2021-03-02 at 10 28 48 am

TomikaArome:
Reverting original change of the equal height between the two tabs (Repository & Upload from device) awaiting a better fix

PR #6764, opened 2021-03-02
Revert "[AAE-4637] Fix varying height of dialog when clicking on the Upload from your device tab"

TomikaArome:
Reverts Alfresco/alfresco-ng2-components#6690

Reverting to come up with a better fix for later


andrea-ligios (approved):
LGTM

PR #6807, opened 2021-03-10
[ACA-4202] - forcing firefox to ignore cache when document version ha…

VitoAlbano:
…s changed

> - [X] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

Only for Firefox whenever the user updates a new version with the same name of a doc document whilst viewing it, the document is not refreshed but it's taken from the FF cache.

Whenever a new version steps in, we set the header of pdf-js to 'no-cache' and this will force the browser to actual fetch the new version

> - [ ] Yes
> - [ ] No

**Other information**:
https://alfresco.atlassian.net/browse/ACA-4202

### DIFF
--- a/lib/core/viewer/components/viewer.component.html
+++ b/lib/core/viewer/components/viewer.component.html
@@ -212,6 +212,7 @@
                                             [blobFile]="blobFile"
                                             [urlFile]="urlFileContent"
                                             [nameFile]="displayName"
+                                            [cacheType]="cacheTypeForContent"
                                             (error)="onUnsupportedFile()"></adf-pdf-viewer>
                         </ng-container>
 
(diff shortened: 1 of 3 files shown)


lib/core/viewer/components/viewer.component.html
```diff
@@ -212,6 +212,7 @@ <h2>{{ 'ADF_VIEWER.LOADING' | translate }}</h2>
                                             [blobFile]="blobFile"
                                             [urlFile]="urlFileContent"
                                             [nameFile]="displayName"
+                                            [cacheType]="cacheTypeForContent"
```
eromano, line 215: do we need a new input? or we can add it always?

VitoAlbano: Due the fact that this affects the caching for when we get the file it might be better to have a flexible strategy. Maybe the no-cache is something that makes sense for us but not in some other cases when the viewer is used.

eromano (approved):

PR #6814, opened 2021-03-11
[ADF-5360] Fix App Subscription Initialization

davidcanonieto:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://alfresco.atlassian.net/browse/ADF-5360

App notifications are initialized before the login happens. Because of that there is no token to be sent to the graphql endpoint to authenticate. We need to wait for the login to happen before we can start the subscription

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/demo-shell/src/app/services/app-notifications.service.ts
+++ b/demo-shell/src/app/services/app-notifications.service.ts
@@ -14,11 +14,12 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import { AppConfigService, NotificationService, NotificationModel, AlfrescoApiService, IdentityUserService } from '@alfresco/adf-core';
+import { AppConfigService, NotificationService, NotificationModel, AlfrescoApiService, IdentityUserService, AuthenticationService } from '@alfresco/adf-core';
 import { NotificationCloudService } from '@alfresco/adf-process-services-cloud';
 import { Injectable } from '@angular/core';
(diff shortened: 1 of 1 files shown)


dhrn (approved):

demo-shell/src/app/services/app-notifications.service.ts
```diff
@@ -43,9 +44,16 @@ export class AppNotificationsService {
         private notificationService: NotificationService,
         private translateService: TranslateService,
         private identityUserService: IdentityUserService,
-        private alfrescoApiService: AlfrescoApiService
+        private alfrescoApiService: AlfrescoApiService,
+        private authenticationService: AuthenticationService
     ) {
-        this.alfrescoApiService.alfrescoApiInitialized.subscribe(() => {
+        merge(this.alfrescoApiService.alfrescoApiInitialized, this.authenticationService.onLogin)
```
eromano, line 50: this.authenticationService.onLogin seems not be triggered when is in SSO probably we need to fix something there as well.

eromano (changes requested):

demo-shell/src/app/services/app-notifications.service.ts
davidcanonieto: I've checked with SSO and it works fine, let me have another look

eromano:
I already fixed it in antoher PR let's close this

PR #6816, opened 2021-03-12
[ACA-3188] Add methods for ADW+APS1 e2e

iuliaib:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

> - [ ] Yes
> - [ ] No

**Other information**:


lib/testing/src/lib/core/pages/form/widgets/attach-file-widget.page.ts
```diff
@@ -122,6 +122,15 @@ export class AttachFileWidgetPage {
         return this.removeFileOptionButton.isEnabled();
     }
 
+    async isRemoveFileOptionDisplayed(): Promise<boolean> {
+        try {
+            await this.removeFileOptionButton.isDisplayed();
```
eromano, line 127: isDisplayed doesn't need a try-catch and if you convert this element in TestElement the TestElement Class has already all the options.  A better option is convert all the element of this page in TestElement

eromano (changes requested):

lib/testing/src/lib/process-services/pages/start-process.page.ts
```diff
@@ -32,6 +32,7 @@ export class StartProcessPage {
     formStartProcessButton = element(by.css('button[data-automation-id="adf-form-start process"]'));
     startProcessButton = element(by.css('button[data-automation-id="btn-start"]'));
     noProcess = element(by.css('.adf-empty-content__title'));
+    startProcessForm = element(by.css('.adf-start-form-container'));
```
cristinaj, line 35: This element is part of StartFormComponent ADF component, that is used on start-process component. Maybe it would be good to have a different PO for it.

iuliaib: I've created a new PO for Start Form component. Thanks

lib/testing/src/lib/core/pages/form/widgets/attach-file-widget.page.ts
iuliaib: I've changed the check over the element, thanks.
And about changing all the elements from this page, I think it would be better to have a different task for refactoring

cristinaj (approved):

lib/testing/src/lib/core/pages/form/widgets/attach-file-widget.page.ts
```diff
@@ -122,6 +122,15 @@ export class AttachFileWidgetPage {
         return this.removeFileOptionButton.isEnabled();
     }
 
+    async isRemoveFileOptionDisplayed(): Promise<boolean> {
+        try {
+            await BrowserVisibility.waitUntilElementIsVisible(this.removeFileOptionButton);
```
eromano, line 127: can you make removeFileOptionButton a testelement has already that method inside

lib/testing/src/lib/core/pages/form/start-form.page.ts
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+import { element, by } from 'protractor';
+import { BrowserVisibility } from '../../utils/browser-visibility';
+
+export class StartFormPage {
+    startProcessForm = element(by.css('.adf-start-form-container'));
+
+    async checkStartProcessFormIsDisplayed(): Promise<void> {
+        await BrowserVisibility.waitUntilElementIsVisible(this.startProcessForm);
```
eromano, line 25: can you make startProcessForm a testelement has already that method inside

eromano (changes requested):

lib/testing/src/lib/core/pages/form/widgets/attach-file-widget.page.ts
iuliaib: I've changed the `removeFileOptionButton` locator to TestElement

lib/testing/src/lib/core/pages/form/start-form.page.ts
iuliaib: I've changed `startProcessForm` locator to TestElement

lib/testing/src/lib/core/pages/form/start-form.page.ts
```diff
@@ -0,0 +1,27 @@
...
+ * limitations under the License.
+ */
+
+import { TestElement } from '../../test-element';
+
+export class StartFormPage {
+    startProcessForm = TestElement.byCss('.adf-start-form-container');
+
+    async checkStartProcessFormIsDisplayed(): Promise<void> {
+        await this.startProcessForm.isDisplayed();
```
eromano, line 24: this wrapper is not needed just use it where you need startFormPage.startProcessForm.isDisplayed()

eromano (changes requested):

lib/testing/src/lib/core/pages/form/start-form.page.ts
iuliaib: Why should I use the locator directly in the test class? It's not a good practice from many points of view:

- What if the name of the locator change? -> Then we should rename the variable in all the tests;
- What if somewhere in the future, the UI design changes and a single check over an element is no longer enough or valid [for example: before performing the "isDisplayed" check I need to add a "wait for something" or another action is needed] -> Then again we should update all the tests that are performing this check;
- Following this proposal is not ensuring code reusability and maintainability;
- This proposal is not respecting the **Encapsulation OOP principle**. Even if our variables are not of type private, it's still not correct to use the data of a class in another class;
- This proposal is not respecting the [Page Object principles](https://www.selenium.dev/documentation/en/guidelines_and_recommendations/page_object_models/).

adinapitul: @iuliaib , you wouldn't be using the actual locator in the test class. 
In the test you would have something like: 
`const startFormPage = new StartFormPage();`
and then in the test, assertions like:
`expect(await startFormPage.startProcessForm.isDisplayed()).toBe(true, ...)`
The actual selector is still in the StartFormPage class.

iuliaib: Yes @adinapitul, I've understood this, thanks. I was wondering why to use a variable from the PO in the test class because of the reasons from above.

adinapitul: If you don't want to access the member variable **startProcessForm** directly, you can create a get.

```
get startProcessForm()
{
    return this.startProcessForm;
}
```

iuliaib: @adinapitul What would be the difference between having a get in the **StartFormPage** PO and having directly the method in that PO? Why would be better to use the `get` approach instead of using the method?

Also, what about the other points mentioned above?

mauriziovitale: The locator should always be private. 
The StartFormPage should have methods like 

checkFormIsVisible()
checkStartIsEnabled()
cancel()
start()

eromano: We are asking ourselves here if we need a get method to access a property of an object in Typescript to not violate Demeter Principle or Encapsulation as you were saying. 
IMO in Typescript adding the get method to access property is not needed for the nature public of it unless we don't force and specify them as read-only. 

- If the name of the locator change probably you will need to change also the method since are really similar
- For the future, this is anticipation we try to avoid anticipation
- It actually the opposite is less code IMO to maintain
- I agree on the OOP principle sho you should have a getStartProcessForm especially in a language like Java or C++ where usually class property is private but in Typescript this looks like an antipattern adding some code not really needed to access to a property that is already accessible

cristinaj: The locators are accessible right now, but we should make them private. I see more benefit in protecting them and respecting POO principles + Page Object pattern. Also It's much more easier for a person who is writing a new test to call such methods 'checkStartProcessFormIsDisplayed()', than starting to ask himself what is needed in order to check the form is displayed: what locator should i check/use? is enough to check only for startProcessForm locator in this case? do i need to check for some other element too? Not sure if i was clear in my explanation, but i see only benefits in adding these kind of methods.

iuliaib: > If the name of the locator change probably you will need to change also the method since are really similar

Yes, but we will need to change only the method, not to rename the variable in all the E2Es.
> For the future, this is anticipation we try to avoid anticipation

At the same time, when we write a method we need to make it as general as possible so we can easily reuse it. Right? Following this, from my point of view, It's better and safer to take precautions. Most of all when we're changing the approach from the current one where it's easier to avoid these kinds of anticipated behaviours which will generate changes in the affected E2E.

> It actually the opposite is less code IMO to maintain

In "the old way", if something changes, the only place that needs to be updated is the `StartFormPage` PO. Isn't this even less code to maintain?

adinapitul: - the tests are the ones that need to decide what to check and how to check, not the page objects, so, as the selenium guideline says, methods like "checkThis", or "checkThat" do not belong in the page objects
- the simpler the code is, the easier would be to maintain, debug and understand it. So I don't see the benefit of having a method calling another method that calls another method, when all this can be written in a single line without all those extra wrappers
- the "what if"s, even if they are valid, can be handled when that actually happens

cristinaj: https://github.com/SeleniumHQ/selenium/wiki/PageObjects: 'Try not to expose the internals of the page'.

cristinaj: Where exactly is the selenium guideline saying that the methods like "checkThis"  do not belong in the page objects?

adinapitul: > Where exactly is the selenium guideline saying that the methods like "checkThis" do not belong in the page objects?

in the fragment I pasted above. "checkThis" is a verification / assertion.

iuliaib: > Page objects themselves should never make verifications or assertions. This is part of your test and should always be within the test’s code, never in an page object.

From my point of view, this is referring to not use expect() in the PO like in this example: https://github.com/Alfresco/alfresco-ng2-components/blob/develop/e2e/core/pages/content-services.page.ts#L559 

> the tests are the ones that need to decide what to check and how to check, not the page objects, so, as the selenium guideline says, methods like "checkThis", or "checkThat" do not belong in the page objects

I agree. I also think the assertion needs to be performed in the tests and not in the PO. So I've changed this method in `isStartProcessFormDisplayed(): Promise<boolean>` which would be used in the E2E as:
`expect(await startFormPage.isStartProcessFormDisplayed()).toBe(true, 'The form is not displayed on Start Process page.');`

adinapitul: yeap, with this renaming, you are addressing the concern of having verifications inside the page object. Good. Thanks

cristinaj: I think this refers to not have asserts 

> > Where exactly is the selenium guideline saying that the methods like "checkThis" do not belong in the page objects?
> 
> in the fragment I pasted above. "checkThis" is a verification / assertion.

I think this refers to assertions, which are done in the test in our case. 
At this link https://github.com/SeleniumHQ/selenium/wiki/PageObjects i found this 'rule': 'Try not to expose the internals of the page'. For me this is the correct approach, for all the arguments listed above.

iuliaib:
@eromano based on the meeting that we had yesterday I've deleted the `isDisplayed()` methods.
Could you please, review again these changes?
Thanks

eromano (approved):
Thanks

PR #6881, opened 2021-03-30
[AAE-4428] Add selected file counter to attach file widget

TomikaArome:

> - [ ] Bugfix
> - [x] Feature
> - [ ] Build related changes
> - [ ] Documentation

No way to tell exactly how many files are selected without counting them manually.

A counter in the toolbar to the right of the breadcrumb with the number of files selected is displayed.
image

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/content-services/src/lib/content-node-selector/content-node-selector-panel.component.html
+++ b/lib/content-services/src/lib/content-node-selector/content-node-selector-panel.component.html
@@ -59,7 +59,7 @@
             </adf-dropdown-breadcrumb>
             <div
                 data-automation-id="content-node-selector-selected-counter">
-                {{ getSelectedCountTranslation() }}
+                {{ 'NODE_SELECTOR.SELECTED_COUNT' | translate: { count: getSelectedCount() } }}
             </div>
         </adf-toolbar-title>
     </adf-toolbar>
--- a/lib/content-services/src/lib/content-node-selector/content-node-selector-panel.component.html
+++ b/lib/content-services/src/lib/content-node-selector/content-node-selector-panel.component.html
@@ -57,6 +57,10 @@
                 [root]="breadcrumbFolderTitle"
                 data-automation-id="content-node-selector-content-breadcrumb">
             </adf-dropdown-breadcrumb>
+            <div
+                data-automation-id="content-node-selector-selected-counter">
+                {{ getSelectedCountTranslation() }}
+            </div>
         </adf-toolbar-title>
     </adf-toolbar>
 
--- a/lib/content-services/src/lib/content-node-selector/content-node-selector-panel.component.html
+++ b/lib/content-services/src/lib/content-node-selector/content-node-selector-panel.component.html
@@ -43,7 +43,7 @@
         </adf-search-panel>
         <div class="adf-content-node-selector-document-list-container">
     <adf-toolbar>
-        <adf-toolbar-title>
+        <adf-toolbar-title [adf-node-counter]="getSelectedCount()">
             <ng-container *ngIf="!showBreadcrumbs()">
                 <span role="heading" aria-level="3" class="adf-search-results-label">{{ 'NODE_SELECTOR.SEARCH_RESULTS' | translate }}</span>
             </ng-container>
@@ -57,10 +57,6 @@
                 [root]="breadcrumbFolderTitle"
                 data-automation-id="content-node-selector-content-breadcrumb">
             </adf-dropdown-breadcrumb>
-            <div
-                data-automation-id="content-node-selector-selected-counter">
-                {{ 'NODE_SELECTOR.SELECTED_COUNT' | translate: { count: getSelectedCount() } }}
-            </div>
         </adf-toolbar-title>
     </adf-toolbar>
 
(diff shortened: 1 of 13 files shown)


eromano (changes requested):
missing unit test

lib/content-services/src/lib/content-node-selector/content-node-selector-panel.component.ts
```diff
@@ -278,6 +280,14 @@ export class ContentNodeSelectorPanelComponent implements OnInit, OnDestroy {
         return this._chosenNode;
     }
 
+    getSelectedCountTranslation(): string {
+        return this.translation.instant(`NODE_SELECTOR.SELECTED_COUNT`, { count: this.getSelectedCount() });
```
arditdomi, line 284: you could do the same in the html and let the translate pipe do the job by doing:

`'NODE_SELECTOR.SELECTED_COUNT' | translate: { count: getSelectedCount() }`

Not much different from what you are doing though, just less number of services injected in the constructor in this way as you don't need to inject the translation service, but up to you to decide which one you want to use

TomikaArome: Good idea, I will change it

lib/content-services/src/lib/content-node-selector/content-node-selector-panel.component.html
```diff
@@ -57,6 +57,10 @@
                 [root]="breadcrumbFolderTitle"
                 data-automation-id="content-node-selector-content-breadcrumb">
             </adf-dropdown-breadcrumb>
+            <div
+                data-automation-id="content-node-selector-selected-counter">
```
arditdomi, line 61: this data-automation-id is not used anywhere

lib/content-services/src/lib/content-node-selector/content-node-selector-panel.component.spec.ts
```diff
@@ -1304,5 +1304,24 @@ describe('ContentNodeSelectorPanelComponent', () => {
                 expect(component.sorting).toEqual(['createdAt', 'desc']);
             });
         });
+
+        describe('Selected nodes counter', () => {
```
arditdomi, line 1308: Could you add an extra check that checks the binding with the html? Right now your test is checking the getselectedCount() method but if someone changes the html that you added, the counter could be broken without any tests highlighting a problem

arditdomi (changes requested):

lib/content-services/src/lib/content-node-selector/content-node-selector.component.spec.ts
```diff
@@ -393,4 +394,28 @@ describe('ContentNodeSelectorComponent', () => {
...
+            component.onSelect([new Node({ id: 'fake' })]);
+
+            expect(component.getSelectedCount()).toBe(1);
+        });
+
+        it('should have injected the counter text', () => {
+            fixture.detectChanges();
+            fixture.whenStable().then(() => {
+                const breadcrumb = fixture.debugElement.query(By.directive(DropdownBreadcrumbComponent));
+                expect(breadcrumb).not.toBeNull('Breadcrumb not found');
```
arditdomi, line 413: Why do we need to check the breadcrumb?

lib/content-services/src/lib/content-node-selector/content-node-selector-panel.component.html
mauriziovitale, line 61: is the counter visible on the local upload tab too? 
I was expecting to have a directive

lib/content-services/src/lib/content-node-selector/content-node-selector.component.spec.ts
TomikaArome: Not sure why I left that in. I think I assumed the breadcrumb would have to be present for the counter to also be present, when that's not true. Removing.

arditdomi (approved):

mauriziovitale (approved):

lib/content-services/src/lib/directives/node-counter.directive.ts
```diff
@@ -0,0 +1,55 @@
...
+
+    componentRef: NodeCounterComponent;
+
+    constructor(
+        private resolver: ComponentFactoryResolver,
+        public viewContainerRef: ViewContainerRef
+    ) {}
+
+    ngOnInit() {
+        const componentFactory = this.resolver.resolveComponentFactory(NodeCounterComponent);
```
DenysVuika, line 35: why do you need dynamically creating same component?

arditdomi: It is inside the directive that we create the NodeCounterComponent. Not sure if I answered your question or not 😅

DenysVuika (approved):

eromano (changes requested):
missing documentation of the new directive

TomikaArome:
> missing documentation of the new directive

Documentation added. Do we need to wait for the build again? It was green before.

eromano (approved):

PR #6924, opened 2021-04-13
[ADF-5379] Add a way to disable WS notifications

davidcanonieto:

> - [ ] Bugfix
> - [x] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://alfresco.atlassian.net/browse/ADF-5379

Add a way to disable websockets from the app.config.json so that we can stop them when running e2e tests and improve performance.

Two ways have been added:

- Directly from the component's input `enableNotifications`
- From your app.config.json 

> - [ ] Yes
> - [ ] No

**Other information**:


demo-shell/src/app.config.json
```diff
@@ -1009,6 +1009,9 @@
     "expandedSidenav": true,
     "preserveState": true
   },
+  "adf-cloud-task-filters": {
+    "ws-notifications": true
```
mauriziovitale, line 1013: I think that we should have a ws-notification true false across components. right now task list is the only one that uses it but as next task we will have the same for process.

mauriziovitale: @eromano @popovicsandras @DenysVuika what do u guys think?

lib/process-services-cloud/src/lib/task/task-filters/components/task-filters-cloud.component.ts
```diff
@@ -48,12 +55,18 @@ export class TaskFiltersCloudComponent extends BaseTaskFiltersCloudComponent imp
     currentFilter: TaskFilterCloudModel;
 
     constructor(private taskFilterCloudService: TaskFilterCloudService,
-                private translationService: TranslationService) {
+                private translationService: TranslationService,
+                appConfigService: AppConfigService) {
         super();
+        if (this.enableNotifications === undefined) {
+            this.enableNotifications = appConfigService.get(TaskFiltersCloudComponent.WS_NOTIFICATIONS_KEY, false);
```
mauriziovitale, line 62: move this to the init

demo-shell/src/app.config.json
DenysVuika: what about centralizing around the "ws-notifications" as a top feature with some params, like:

```
{
  "notifications": {
    "task-filters": true,
    "process-filters": true
}
```

it's more like iOS/Android notification center settings as for me.

Also, we should not embed the technology stack into names, ws (web sockets) should not be in the name, what if we use different stack in the future... should be just "notifications" as for me

davidcanonieto: I like that @DenysVuika, It'll allow for future expandability

mauriziovitale: I agree to have a general notification property but I don't see to point of having the ability to turn one on and another off. We should probably call the property _**web-socket-notification**_

lib/process-services-cloud/src/lib/task/task-filters/components/task-filters-cloud.component.ts
```diff
@@ -31,6 +31,13 @@ import { TaskCloudEngineEvent } from '../../../models/engine-event-cloud.model';
     styleUrls: ['base-task-filters-cloud.component.scss']
 })
 export class TaskFiltersCloudComponent extends BaseTaskFiltersCloudComponent implements OnInit, OnChanges {
+
+    static WS_NOTIFICATIONS_KEY = 'notifications.task-filters';
+
+    /** Enable Bubble Notifications on Task Filter Count */
+    @Input()
+    enableNotifications: boolean;
```
mauriziovitale, line 39: Do we really need this input? From my point of view, either is the parent to check the app config.
```this.enableNotifications = this.appConfigService.get(TaskFiltersCloudComponent.WS_NOTIFICATIONS_KEY, false);```
and pass the information down or the component itself

**Review by mauriziovitale — CHANGES_REQUESTED** [review]
comment

File: demo-shell/src/app/services/app-notifications.service.ts [hunk 4]
```diff
@@ -54,7 +54,7 @@ export class AppNotificationsService {
         private alfrescoApiService: AlfrescoApiService
     ) {
         this.alfrescoApiService.alfrescoApiInitialized.subscribe(() => {
-            if (this.isProcessServicesEnabled()) {
+            if (this.isProcessServicesEnabled() && this.areWSNotificationsEnabled()) {
```
DenysVuika, line 57: can we reduce technology from the naming and have it called `notificationsEnabled` function?

demo-shell/src/app/services/app-notifications.service.ts
```diff
@@ -76,10 +76,14 @@ export class AppNotificationsService {
         });
     }
 
-    private isProcessServicesEnabled() {
+    private isProcessServicesEnabled(): boolean {
         return this.authenticationService.isLoggedIn() && (this.authenticationService.isBPMProvider() || this.authenticationService.isALLProvider());
     }
 
+    private areWSNotificationsEnabled(): boolean {
```
DenysVuika, line 83: I suggest `notificationsEnabled` to make it more generic

DenysVuika (commented):
Looks good, but check the comments

mauriziovitale (approved):

eromano (approved):

PR #6946, opened 2021-04-23
[ADF-5380] Check Acs status in aps init script

Giovanni007:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

The init-aps-env script doesn't check that ACS is reachable before authorising the content on activiti-app. So sometimes it happens that it fails when trying to authorize users if there are problems with ACS. Right now, that is not clear from the logs.

The init script checks that ACS is reachable before starting the init of the env. If ACS is not reachable the init of the env will fail. So It now supposes that there is always a running instance of ACS and fails with a clear message before creating repositories and authorizing the users

> - [ ] Yes
> - [ ] No

**Other information**:

### DIFF
--- a/lib/cli/scripts/init-aps-env.ts
+++ b/lib/cli/scripts/init-aps-env.ts
@@ -140,6 +140,7 @@ async function checkEnv() {
             checkEnv();
         }
     }
+    await checkAcsEnv();
 }
 
 async function hasDefaultTenant(tenantId, tenantName) {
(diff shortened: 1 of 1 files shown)


lib/cli/scripts/init-aps-env.ts
```diff
@@ -434,3 +435,24 @@ function sleep(delay) {
     const start = new Date().getTime();
     while (new Date().getTime() < start + delay) {  }
 }
+
+async function checkAcsEnv() {
+    try {
+        const repositoryEntry = await alfrescoJsApiRepo.oauth2Auth.callCustomApi(
```
eromano, line 441: we have another script that jsut cehck ACS why don't just use that before instead of coupling and duplicate? the script is check-cs-env

eromano (changes requested):

lib/cli/scripts/init-aps-env.ts
Giovanni007: Do you mean I should use check-cs-env in the single pipeline script [initialize_services.sh](https://github.com/Alfresco/terraform-alfresco-pipeline/blob/cfc1f1190be3299bfc85e07bc94da3323f7882fb/scripts/travis/build/initialize_services.sh#L18) before initializing APS?

eromano: yep

PR #7021, opened 2021-05-12
[ACA-2864] Site Managers are not able to disable inherit permissions …

dhrn:
…in DW/ACA

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://alfresco.atlassian.net/browse/ACA-2864

Sitemanger can't toggle Inherit permission

fixed the problem

What is the fix?
We should include the site_manger group to node before toggling the inherit permission

why are you doing this hack?
As soon as we toggle the inherit permission user can't get the access to the node. so we should include to the locallyset before toggling the inherit permission. so backend is not letting the user to do

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/content-services/src/lib/permission-manager/components/permission-list/permission-list.service.ts
+++ b/lib/content-services/src/lib/permission-manager/components/permission-list/permission-list.service.ts
@@ -39,11 +39,13 @@ export class PermissionListService {
     nodeWithRoles$: Subject<{ node: Node, roles: RoleModel[] }> = new Subject();
     data$: Observable<NodePermissionsModel> = this.nodeWithRoles$.pipe(
         map(({ node, roles}) => {
+            const nodeLocalPermissions = this.nodePermissionService.getLocalPermissions(node);
+            const localPermissions = this.updateReadOnlyPermission(node, nodeLocalPermissions);
             return {
                 node,
                 roles,
+                localPermissions,
                 inheritedPermissions: this.nodePermissionService.getInheritedPermission(node),
-                localPermissions: this.nodePermissionService.getLocalPermissions(node)
             };
         })
     );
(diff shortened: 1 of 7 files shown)


lib/content-services/src/lib/permission-manager/components/permission-list/permission-list.service.ts
```diff
@@ -201,6 +216,23 @@ export class PermissionListService {
         this.nodeWithRoles$.next({ node: this.node, roles: this.roles });
     }
 
+    private hasLocalPermission(node: Node) {
```
eromano, line 219: missing return types

eromano (changes requested):

adinapitul (changes requested):
I want to review this approach cause it sounds a bit dangerous. Please allow me time till tomorrow morning.

@dhrn , why do you say this "As soon as we toggle the inherit permission user can't get the access to the node." ?

dhrn:
@adinapitul this is how share  is doing.

dhrn:
> I want to review this approach cause it sounds a bit dangerous. Please allow me time till tomorrow morning.
> 
> @dhrn , why do you say this "As soon as we toggle the inherit permission user can't get the access to the node." ?

@adinapitul backend is smart. it won't allow the user to make this possible. what i mean if we didn't include then togglerpermission api will be failed.

adinapitul (approved):
@dhrn , @eromano , I performed some checks and I think this is fine after all. 
My concern was mainly related to complex scenarios involving adding then removing user from the site, or adding then removing user from the system group.... 
But indeed, this is what share is doing also. 

We're fine. Thank you!

eromano (approved):

PR #7024, opened 2021-05-13
[ACA-4423] VIEWER - when closing the upload dialog the filename changes to "Unknown"

ursedaniel:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

Forgot in the original PR to remove the blob line and after update of the node the viewer acts as a blob instead of the image from ACS.

> - [ ] Yes
> - [x] No

**Other information**: https://alfresco.atlassian.net/browse/ACA-4423


montgolfiere (approved):

PunchedChimera (approved):

lib/core/viewer/components/viewer.component.spec.ts
```diff
@@ -962,25 +965,27 @@ describe('ViewerComponent', () => {
...
+                spyOn(uploadService, 'uploadFilesInTheQueue').and.callFake(() => {});
                 component.readOnly = false;
                 component.nodeEntry = new NodeEntry({ entry: { name: 'fakeImage.png', id: '12', content: { mimeType: 'txt' } } });
                 const data = atob('iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAADUlEQVR42mNk+M9QDwADhgGAWjR9awAAAABJRU5ErkJggg==');
                 const fakeBlob = new Blob([data], { type: 'image/png' });
                 component.onSubmitFile(fakeBlob);
                 fixture.detectChanges();
 
-                expect(component.blobFile).toEqual(fakeBlob);
+                expect(uploadService.uploadFilesInTheQueue).toHaveBeenCalled();
```
mauriziovitale, line 976: why are we changing that? before we were checking the file and now only that a method is called which does not make any sense.

lib/core/viewer/components/viewer.component.spec.ts
```diff
@@ -962,25 +965,27 @@ describe('ViewerComponent', () => {
...
+                spyOn(uploadService, 'uploadFilesInTheQueue').and.callFake(() => {});
                 component.readOnly = true;
                 component.nodeEntry = new NodeEntry({ entry: { name: 'fakeImage.png', id: '12', content: { mimeType: 'txt' } } });
                 const data = atob('iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAADUlEQVR42mNk+M9QDwADhgGAWjR9awAAAABJRU5ErkJggg==');
                 const fakeBlob = new Blob([data], { type: 'image/png' });
                 component.onSubmitFile(fakeBlob);
                 fixture.detectChanges();
 
-                expect(component.blobFile).toEqual(undefined);
+                expect(uploadService.uploadFilesInTheQueue).not.toHaveBeenCalled();
```
mauriziovitale, line 988: same as before

mauriziovitale (changes requested):
comments

eromano (changes requested):
can we avoid reloading the image viewer when we save? I don't think is necessary. This is actually also creating some flickering.

lib/core/viewer/components/viewer.component.spec.ts
ursedaniel: because that line with blob was added by mistake, we don't need that

ursedaniel: i'll also add an expect to check if the upload was called with the right BLOB.

ursedaniel:
as investigated with @eromano for the moment ACA always refreshes the viewer based on the subscription for uploadService so we will investigate in the next release the possibility to stop always refreshing the viewer after upload

DenysVuika (approved):

eromano (approved):

mauriziovitale (approved):

mauriziovitale (approved):

PR #7095, opened 2021-06-07
[ACA-4436] remove None option from the nodeType dropdown

adinapitul:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

None option is available and selectable in the nodeType dropdown, which causes ACA-4436

remove the None option from the nodeType select

> - [ ] Yes
> - [x] No

**Other Information**:

### DIFF
--- a/lib/core/card-view/components/card-view-selectitem/card-view-selectitem.component.html
+++ b/lib/core/card-view/components/card-view-selectitem/card-view-selectitem.component.html
@@ -15,7 +15,7 @@
 
                     <adf-select-filter-input *ngIf="showInputFilter" (change)="onFilterInputChange($event)"></adf-select-filter-input>
 
-                    <mat-option *ngIf="showNoneOption() && property.key !== 'nodeType'">{{ 'CORE.CARDVIEW.NONE' | translate }}</mat-option>
+                    <mat-option *ngIf="showNoneOption()">{{ 'CORE.CARDVIEW.NONE' | translate }}</mat-option>
                     <mat-option *ngFor="let option of list$ | async"
                                 [value]="option.key">
                         {{ option.label | translate }}
--- a/lib/core/card-view/components/card-view-selectitem/card-view-selectitem.component.html
+++ b/lib/core/card-view/components/card-view-selectitem/card-view-selectitem.component.html
@@ -15,7 +15,7 @@
 
                     <adf-select-filter-input *ngIf="showInputFilter" (change)="onFilterInputChange($event)"></adf-select-filter-input>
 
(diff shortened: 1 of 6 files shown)


lib/core/card-view/components/card-view-selectitem/card-view-selectitem.component.html
```diff
@@ -15,7 +15,7 @@
 
                     <adf-select-filter-input *ngIf="showInputFilter" (change)="onFilterInputChange($event)"></adf-select-filter-input>
 
-                    <mat-option *ngIf="showNoneOption()">{{ 'CORE.CARDVIEW.NONE' | translate }}</mat-option>
+                    <mat-option *ngIf="showNoneOption() && property.key !== 'nodeType'">{{ 'CORE.CARDVIEW.NONE' | translate }}</mat-option>
```
dhrn, line 18: The problem here is `Cardviewselectmodel` doesn't handle for `dispalynone` option. So the fix will be updating the [model]( https://github.com/Alfresco/alfresco-ng2-components/blob/develop/lib/core/card-view/models/card-view-selectitem.model.ts#L25) and use this while constructing the cardview[ here]( https://github.com/Alfresco/alfresco-ng2-components/blob/develop/lib/content-services/src/lib/content-metadata/services/content-type-property.service.ts#L83)

dhrn (changes requested):

lib/core/card-view/components/card-view-selectitem/card-view-selectitem.component.html
adinapitul: @dhrn , I don't understand why displayNoneOption should be part of the model. Can you explain a bit? What use case do you have in mind?

lib/core/card-view/components/card-view-selectitem/card-view-selectitem.component.ts
```diff
@@ -98,7 +98,7 @@ export class CardViewSelectItemComponent extends BaseCardView<CardViewSelectItem
     }
 
     showNoneOption() {
-        return this.displayNoneOption;
+        return this.displayNoneOption && this.property.key !== 'nodeType';
```
eromano, line 101: this is a generic component we can not embed a business logic like node type here

eromano (changes requested):

lib/core/card-view/components/card-view-selectitem/card-view-selectitem.component.ts
eromano: we need to pass displayNoneOption to false where we use it for the content type

PR #7173, opened 2021-07-13
[AAE-4740] FE - [ADF] Destination picker it's not displaying my files folder if the destination Folder path is set to a folder part of trash

sivakumar414ram:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://alfresco.atlassian.net/browse/AAE-4740

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/process-services-cloud/src/lib/form/components/widgets/attach-file/attach-file-cloud-widget.component.ts
+++ b/lib/process-services-cloud/src/lib/form/components/widgets/attach-file/attach-file-cloud-widget.component.ts
@@ -152,7 +152,7 @@ export class AttachFileCloudWidgetComponent extends UploadCloudWidgetComponent i
         if (!rootNodeId) {
             try {
                 const nodeEntry = await this.getNodeBasedOnPath(destinationFolderPath);
-                rootNodeId = nodeEntry && nodeEntry.entry ? nodeEntry.entry.id : destinationFolderPath.alias;
+                rootNodeId = nodeEntry?.entry ? nodeEntry.entry.id : destinationFolderPath.alias;
             } catch (error) {
                 this.logService.error(error);
             }
--- a/lib/process-services-cloud/src/lib/form/components/widgets/attach-file/attach-file-cloud-widget.component.ts
+++ b/lib/process-services-cloud/src/lib/form/components/widgets/attach-file/attach-file-cloud-widget.component.ts
@@ -29,7 +29,7 @@ import {
     AlfrescoApiService,
     UploadWidgetContentLinkModel
 } from '@alfresco/adf-core';
-import { Node, RelatedContentRepresentation } from '@alfresco/js-api';
+import { Node, NodeEntry, RelatedContentRepresentation } from '@alfresco/js-api';
 import { ContentCloudNodeSelectorService } from '../../../services/content-cloud-node-selector.service';
 import { ProcessCloudContentService } from '../../../services/process-cloud-content.service';
 import { UploadCloudWidgetComponent } from './upload-cloud.widget';
@@ -151,8 +151,8 @@ export class AttachFileCloudWidgetComponent extends UploadCloudWidgetComponent i
 
         if (!rootNodeId) {
             try {
-                const nodeId = await this.getNodeIdBasedOnPath(destinationFolderPath);
-                rootNodeId = nodeId ? nodeId : destinationFolderPath.alias;
+                const nodeEntry = await this.getNodeBasedOnPath(destinationFolderPath);
+                rootNodeId = nodeEntry && nodeEntry.entry ? nodeEntry.entry.id : destinationFolderPath.alias;
             } catch (error) {
                 this.logService.error(error);
             }
@@ -161,16 +161,17 @@ export class AttachFileCloudWidgetComponent extends UploadCloudWidgetComponent i
         return rootNodeId;
     }
 
-    private async getNodeIdBasedOnPath(destinationFolderPath: DestinationFolderPathModel) {
-        let nodeId: string;
+    private async getNodeBasedOnPath(destinationFolderPath: DestinationFolderPathModel): Promise<NodeEntry> {
+        let nodeEntry: NodeEntry;
         if (destinationFolderPath.path) {
-            nodeId = await this.contentNodeSelectorService.fetchNodeIdFromRelativePath(destinationFolderPath.alias, { relativePath: destinationFolderPath.path });
-        }
-        if (!nodeId) {
-            nodeId = await this.contentNodeSelectorService.fetchAliasNodeId(destinationFolderPath.alias);
+            nodeEntry = await this.contentNodeSelectorService.fetchNodeFromRelativePath(destinationFolderPath.alias, { relativePath: destinationFolderPath.path });
         }
 
-        return nodeId;
+        if (!nodeEntry?.entry?.id) {
+            nodeEntry = await this.contentNodeSelectorService.fetchAliasNode(destinationFolderPath.alias);
+        }
+
+        return nodeEntry;
     }
 
     getAliasAndRelativePathFromDestinationFolderPath(destinationFolderPath: string): DestinationFolderPathModel {
--- a/lib/process-services-cloud/src/lib/form/components/widgets/attach-file/attach-file-cloud-widget.component.ts
+++ b/lib/process-services-cloud/src/lib/form/components/widgets/attach-file/attach-file-cloud-widget.component.ts
@@ -29,7 +29,7 @@ import {
     AlfrescoApiService,
     UploadWidgetContentLinkModel
 } from '@alfresco/adf-core';
-import { Node, NodeEntry, RelatedContentRepresentation } from '@alfresco/js-api';
+import { Node, RelatedContentRepresentation } from '@alfresco/js-api';
 import { ContentCloudNodeSelectorService } from '../../../services/content-cloud-node-selector.service';
 import { ProcessCloudContentService } from '../../../services/process-cloud-content.service';
 import { UploadCloudWidgetComponent } from './upload-cloud.widget';
@@ -151,8 +151,8 @@ export class AttachFileCloudWidgetComponent extends UploadCloudWidgetComponent i
 
         if (!rootNodeId) {
             try {
-                const nodeEntry = await this.getNodeBasedOnPath(destinationFolderPath);
-                rootNodeId = nodeEntry?.entry ? nodeEntry.entry.id : destinationFolderPath.alias;
+                const nodeId = await this.getNodeIdBasedOnPath(destinationFolderPath);
+                rootNodeId = nodeId ? nodeId : destinationFolderPath.alias;
             } catch (error) {
                 this.logService.error(error);
             }
@@ -161,17 +161,16 @@ export class AttachFileCloudWidgetComponent extends UploadCloudWidgetComponent i
         return rootNodeId;
     }
 
-    private async getNodeBasedOnPath(destinationFolderPath: DestinationFolderPathModel): Promise<NodeEntry> {
-        let nodeEntry: NodeEntry;
+    private async getNodeIdBasedOnPath(destinationFolderPath: DestinationFolderPathModel) {
+        let nodeId: string;
         if (destinationFolderPath.path) {
-            nodeEntry = await this.contentNodeSelectorService.fetchNodeFromRelativePath(destinationFolderPath.alias, { relativePath: destinationFolderPath.path });
+            nodeId = await this.contentNodeSelectorService.fetchNodeIdFromRelativePath(destinationFolderPath.alias, { relativePath: destinationFolderPath.path });
+        }
+        if (!nodeId) {
+            nodeId = await this.contentNodeSelectorService.fetchAliasNodeId(destinationFolderPath.alias);
         }
 
-        if (!nodeEntry?.entry?.id) {
-            nodeEntry = await this.contentNodeSelectorService.fetchAliasNode(destinationFolderPath.alias);
-        }
-
-        return nodeEntry;
+        return nodeId;
     }
 
     getAliasAndRelativePathFromDestinationFolderPath(destinationFolderPath: string): DestinationFolderPathModel {
--- a/lib/process-services-cloud/src/lib/form/services/content-cloud-node-selector.service.spec.ts
+++ b/lib/process-services-cloud/src/lib/form/services/content-cloud-node-selector.service.spec.ts
@@ -105,8 +105,6 @@ describe('ContentCloudNodeSelectorService', () => {
         service.sourceNodeNotFound = true;
         service.openUploadFileDialog('nodeId', 'single', true, true);
 
-        expect(openDialogSpy).toHaveBeenCalled();
-
         service.close();
 
         expect(service.sourceNodeNotFound).toBe(false);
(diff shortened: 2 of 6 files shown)


lib/process-services-cloud/src/lib/form/components/widgets/attach-file/attach-file-cloud-widget.component.ts
```diff
@@ -151,8 +151,8 @@ export class AttachFileCloudWidgetComponent extends UploadCloudWidgetComponent i
 
         if (!rootNodeId) {
             try {
-                const nodeId = await this.getNodeIdBasedOnPath(destinationFolderPath);
-                rootNodeId = nodeId ? nodeId : destinationFolderPath.alias;
+                const nodeEntry = await this.getNodeBasedOnPath(destinationFolderPath);
+                rootNodeId = nodeEntry && nodeEntry.entry ? nodeEntry.entry.id : destinationFolderPath.alias;
```
arditdomi, line 155:
```suggestion
                rootNodeId = nodeEntry?.entry ? nodeEntry.entry.id : destinationFolderPath.alias;
```

lib/process-services-cloud/src/lib/form/services/content-cloud-node-selector.service.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+        const relativePathNodeEntry = await service.fetchNodeFromRelativePath('mock-alias', { relativePath: 'mock-relativePath' });
+        service.openUploadFileDialog('nodeId', 'single', true, true).subscribe(() => {
+            expect(openDialogSpy).toHaveBeenCalled();
+            expect(service.sourceNodeNotFound).toBe(true);
+            expect(showWarningSpy).toHaveBeenCalledWith('success');
+            expect(relativePathNodeEntry).toBeNull();
+        });
+    });
+
+    it('should not be able to show an notification if the relative path is valid', () => {
```
arditdomi, line 84: Can we change the title to `should not show a notification if the relative path is valid` ?

lib/process-services-cloud/src/lib/form/services/content-cloud-node-selector.service.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+        expect(showWarningSpy).not.toHaveBeenCalled();
+    });
+
+    it('should be able to fetch given relative path node id', async () => {
+        getNodeSpy.and.returnValue(Promise.resolve(relativePathNodeResponseBody));
+        await service.fetchNodeFromRelativePath('mock-alias', { relativePath: 'mock-relativePath' });
+
+        expect(getNodeSpy).toHaveBeenCalledWith('mock-alias', {
+            relativePath: 'mock-relativePath'
+        });
```
arditdomi, line 99: I have the feeling we can enrich those tests, instead of calling a method manually and expect for something to be returned or called can we test something more meaningful? Here we are checking that a wrapper method performs the API call with the right parameters. I am not fully into the context and I might be wrong but please take a moment to think if we can test something more meaningful.

lib/process-services-cloud/src/lib/form/services/content-cloud-node-selector.service.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+    });
+
+    it('should be able to fetch given alias node id', async () => {
+        getNodeSpy.and.returnValue(Promise.resolve(aliasNodeResponseBody));
+        await service.fetchAliasNode('mock-alias');
+
+        expect(getNodeSpy).toHaveBeenCalledWith('mock-alias');
```
arditdomi, line 106: same here again we pass something and just test that the wrapper function just calls the API with the parameters

lib/process-services-cloud/src/lib/form/services/content-cloud-node-selector.service.ts
```diff
@@ -16,66 +16,85 @@
...
+                this.handleError(error);
+                return of(<NodeEntry> {});
+            })
+        ).toPromise();
     }
 
-  private openContentNodeDialog(data: ContentNodeSelectorComponentData, currentPanelClass: string, chosenWidth: string) {
-    this.dialog.open(ContentNodeSelectorComponent, { data, panelClass: currentPanelClass, width: chosenWidth });
-  }
+    private openContentNodeDialog(data: ContentNodeSelectorComponentData, currentPanelClass: string, chosenWidth: string) {
```
arditdomi, line 80: are you sure the indentation you are applying is correct and the previous one was wrong?

lib/process-services-cloud/src/lib/form/services/content-cloud-node-selector.service.ts
```diff
@@ -16,66 +16,85 @@
...
+            })
+        ).toPromise();
     }
 
-    async fetchAliasNodeId(alias: string): Promise<string> {
-        const aliasNodeEntry: any = await this.apiService.getInstance().node
-        .getNode(alias)
-        .catch((err) => this.handleError(err));
-        return aliasNodeEntry?.entry?.id;
+    fetchAliasNode(alias: string): Promise<NodeEntry> {
```
arditdomi, line 71: Why do we need to change this method to return the Node instead of the nodeId that it was returning before? What is the benefit of it, are we solving any problem by this?

lib/process-services-cloud/src/lib/form/services/content-cloud-node-selector.service.ts
```diff
@@ -16,66 +16,85 @@
...
+        return from(this.apiService.nodesApi.getNode(alias)).pipe(
+            catchError((error) => {
+                this.handleError(error);
+                return of(<NodeEntry> {});
```
arditdomi, line 75: typecasting an empty object to avoid compilation errors I am not sure is the right thing to do. Can we investigate what is a good practice for that and what do we do in other places?

arditdomi (changes requested):

lib/process-services-cloud/src/lib/form/services/content-cloud-node-selector.service.ts
sivakumar414ram: Yes, the new Indentation is the correct one. I guess the previous indentation was wrong I changed it after checking with other services indentation.

lib/process-services-cloud/src/lib/form/services/content-cloud-node-selector.service.ts
sivakumar414ram: rollbacked

lib/process-services-cloud/src/lib/form/services/content-cloud-node-selector.service.ts
sivakumar414ram: rollbacked

lib/process-services-cloud/src/lib/form/services/content-cloud-node-selector.service.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+        });
+        getNodeSpy = spyOn(apiService.nodesApi, 'getNode');
+    });
+
+    it('should be able to open the content node select panel dialog', () => {
+        service.openUploadFileDialog('nodeId', 'single', true, true);
+        expect(openDialogSpy).toHaveBeenCalled();
+    });
+
+    it('should be able to show an notification if the relative path is invalid/deleted', async () => {
```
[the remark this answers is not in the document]
sivakumar414ram: Updated, Please review again

lib/process-services-cloud/src/lib/form/services/content-cloud-node-selector.service.spec.ts
sivakumar414ram: Actually, this test is not checking mocked responses. It's checking getNodeSpy(JS-API) has been called with the right parameters.

lib/process-services-cloud/src/lib/form/services/content-cloud-node-selector.service.spec.ts
sivakumar414ram: Actually, this test is not checking mocked responses. It's checking getNodeSpy(JS-API) has been called with the right parameters.

lib/process-services-cloud/src/lib/form/services/content-cloud-node-selector.service.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+            relativePath: 'mock-wrong-relativePath'
+        });
+        expect(service.sourceNodeNotFound).toBe(true);
+    });
+
+    it('should be able to set sourceNodeNotFound value to false after the dialog close', async () => {
+        service.sourceNodeNotFound = true;
+        service.openUploadFileDialog('nodeId', 'single', true, true);
+
+        expect(openDialogSpy).toHaveBeenCalled();
```
arditdomi, line 108: It is better not to add extra checks along the way of testing a particular thing. A unit test should test one thing. Here you called the `service.openUploadFileDialog` and you check that the dialog is opened when your test title says something else. Then you test that the sourceNodeNotFound should be false. 

I see 2 tests here (If opening the dialog is not covered already, it probably is). This causes a small refactoring to require a lot of changes when there is no true benefit for adding this check. If there is no test about that then there should be:
- 1 test that checks that the dialog opens.
- 1 test (your test) that should only check what the title says. `should be able to set sourceNodeNotFound value to false after the dialog close`

lib/process-services-cloud/src/lib/form/services/content-cloud-node-selector.service.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+            afterOpened: () => of({}),
+            afterClosed: () => of({}),
+            componentInstance: {
+                error: new Subject<any>()
+            }
+        });
+        getNodeSpy = spyOn(apiService.nodesApi, 'getNode');
+    });
+
+    it('should be able to open the content node select panel dialog', () => {
```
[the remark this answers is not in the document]
sivakumar414ram: Updated

lib/process-services-cloud/src/lib/form/services/content-cloud-node-selector.service.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+        service.close();
+
+        expect(service.sourceNodeNotFound).toBe(false);
+    });
+
+    it('should be able to show an notification if the relative path is invalid/deleted', () => {
+        service.sourceNodeNotFound = true;
+        service.openUploadFileDialog('nodeId', 'single', true, true);
+
+        expect(openDialogSpy).toHaveBeenCalled();
```
arditdomi, line 119: same here. Imagine if someone refactors the openUploadFileDialog method. Why will they need to change a test that is called `should be able to show a notification` ?

lib/process-services-cloud/src/lib/form/services/content-cloud-node-selector.service.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+
+        service.close();
+
+        expect(service.sourceNodeNotFound).toBe(false);
+    });
+
+    it('should be able to show an notification if the relative path is invalid/deleted', () => {
```
arditdomi, line 115:
```suggestion
    it('should be able to show a notification if the relative path is invalid/deleted', () => {
```

lib/process-services-cloud/src/lib/form/services/content-cloud-node-selector.service.ts
```diff
@@ -27,55 +27,67 @@ import { Observable, Subject, throwError } from 'rxjs';
...
+        .catch((err) => {
+            this.sourceNodeNotFound = true;
+            return this.handleError(err);
+        });
         return relativePathNodeEntry?.entry?.id;
     }
 
     async fetchAliasNodeId(alias: string): Promise<string> {
-        const aliasNodeEntry: any = await this.apiService.getInstance().node
+        const aliasNodeEntry: any = await this.apiService.nodesApi
```
arditdomi, line 68: why `this.apiService.getInstance().node` was a problem? Is there a problematic use case?

arditdomi (changes requested):

lib/process-services-cloud/src/lib/form/services/content-cloud-node-selector.service.ts
sivakumar414ram: Yup, `.getInstance().node`  is deprecated

lib/process-services-cloud/src/lib/form/services/content-cloud-node-selector.service.spec.ts
sivakumar414ram: I see this spec looks good. This spec is testing enter flow once you click on the widget.

1. Is Dialog opened
2. Close the dialog
3. Update sourceNodeNotFound flag after dialog close. (Without the first and second steps, we won't reach 3rd step right.)


It would be always better to check each step (like we do on e2e) instead of testing a single check. I can remove it but I do not see any benefit from it.

lib/process-services-cloud/src/lib/form/services/content-cloud-node-selector.service.spec.ts
sivakumar414ram: We are not showing notification before dialog opens, We show it after dialog opens. If someone changing/refactoring openUploadFileDialog they should change notification logic as well

lib/process-services-cloud/src/lib/form/services/content-cloud-node-selector.service.spec.ts
arditdomi: 1. You open the dialog
2. You close it
3. you check what you need to check
This is correct I am not saying it is wrong.

The thing that I think you should not do is check that the dialog is opened.
Right now the flow is:

1. You open the dialog
2. You check that the dialog is open (That is the one I am suggesting to remove)
3. You close it
4. you check what you need to check

To respect that each test should test one thing you should have 2 tests to cover them
Test 1) Open the dialog -> Check that the dialog is opened
Test 2) open the dialog -> Close the dialog -> Check that the sourceNodeNotFound is false

lib/process-services-cloud/src/lib/form/services/content-cloud-node-selector.service.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+    };
+
+    const aliasNodeResponseBody = {
+        entry: {
+            id: 'mock-alias-node-id'
+        }
+    };
+
+    setupTestBed({
+        imports: [TranslateModule.forRoot(), ProcessServiceCloudTestingModule, MatDialogModule]
```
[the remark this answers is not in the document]
sivakumar414ram: Updated

lib/process-services-cloud/src/lib/form/services/content-cloud-node-selector.service.spec.ts
sivakumar414ram: Ok Removed it 

Here is the separate test for Open the dialog

https://github.com/Alfresco/alfresco-ng2-components/pull/7173/files#diff-b9940e991191204dd31c27360bea21130ae12a0c32a376a75863c6e75a0ad5ccR71-R90

arditdomi (approved):

PR #7355, opened 2021-11-09
[AAE-6224] - fixed form field values when rest dropdown has loaded value

VitoAlbano:

> - [X] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

> - [ ] Yes
> - [ ] No

**Other information**:

### DIFF
--- a/lib/process-services-cloud/src/lib/form/components/widgets/dropdown/dropdown-cloud.widget.ts
+++ b/lib/process-services-cloud/src/lib/form/components/widgets/dropdown/dropdown-cloud.widget.ts
@@ -90,6 +90,7 @@ export class DropdownCloudWidgetComponent extends WidgetComponent implements OnI
                 .pipe(takeUntil(this.onDestroy$))
                 .subscribe((result: FormFieldOption[]) => {
                     this.field.options = result;
+                    this.field.updateForm();
                 }, (err) => this.handleError(err));
         }
     }
--- a/lib/process-services-cloud/src/lib/form/components/widgets/dropdown/dropdown-cloud.widget.ts
+++ b/lib/process-services-cloud/src/lib/form/components/widgets/dropdown/dropdown-cloud.widget.ts
@@ -108,7 +108,6 @@ export class DropdownCloudWidgetComponent extends WidgetComponent implements OnI
     private loadFieldOptionsForLinkedWidget() {
         const parentWidgetValue = this.getParentWidgetValue();
         this.parentValueChanged(parentWidgetValue);
-        this.field.updateForm();
     }
 
     private getParentWidgetValue(): string {
@@ -143,6 +142,7 @@ export class DropdownCloudWidgetComponent extends WidgetComponent implements OnI
             rulesEntries.forEach((ruleEntry: RuleEntry) => {
                 if (ruleEntry.key === value) {
                     this.field.options = ruleEntry.options;
+                    this.field.updateForm();
                 }
             });
         }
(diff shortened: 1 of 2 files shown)


dhrn (approved):

lib/process-services-cloud/src/lib/form/components/widgets/dropdown/dropdown-cloud.widget.ts
```diff
@@ -90,6 +90,7 @@ export class DropdownCloudWidgetComponent extends WidgetComponent implements OnI
                 .pipe(takeUntil(this.onDestroy$))
                 .subscribe((result: FormFieldOption[]) => {
                     this.field.options = result;
+                    this.field.updateForm();
```
mauriziovitale, line 93: we already have another place where we call the updateForm is not enough ?
```suggestion
                     if (this.isLinkedWidget()) {
            this.loadFieldOptionsForLinkedWidget();
```

arditdomi: The ability to add a default value has been removed in modeling, so here we are fixing only a scenario for old apps that contain the default value in their form definition. In the next release we are also changing the validation to not allow you to have a form with a dropdown and a default value https://alfresco.atlassian.net/browse/AAE-6250 Not sure at this point if we need this fix at all

VitoAlbano: Ok what we are missing here is the main point : A dropdown with a value set by a variable not a default value needs to add itself to the form values. If it doesn't add its value to the form values you cannot complete the task .

VitoAlbano: @mauriziovitale It looks like that, expecially 'cause if you don't have the options loaded like in a rest dropdown you don't save any values there.

arditdomi: Ok cool, the only point that I see an improvement here is in case the dropdown is linked and of type REST and the form is saved/completed, the updateForm will be called twice. Once here https://github.com/Alfresco/alfresco-ng2-components/blob/cd795c1344c2d6a964446f92ed5815d47f887cc7/lib/process-services-cloud/src/lib/form/components/widgets/dropdown/dropdown-cloud.widget.ts#L93 and once here https://github.com/Alfresco/alfresco-ng2-components/blob/cd795c1344c2d6a964446f92ed5815d47f887cc7/lib/process-services-cloud/src/lib/form/components/widgets/dropdown/dropdown-cloud.widget.ts#L111

Can we check if there is any way to avoid that double call?

arditdomi: If the task is saved/completed and you reload your browser, you will see that it is called twice

VitoAlbano: @arditdomi it's behind an if what IS called twice ? The options retrieval?

VitoAlbano: @arditdomi ok i got it :D 
`The call is better than anything`

arditdomi (approved):

lib/process-services-cloud/src/lib/form/components/widgets/dropdown/dropdown-cloud.widget.ts
arditdomi: We had a call with Vito and figured it out. Resolving the comment

PR #7362, opened 2021-11-10
[AAE-6382] User acronym is not displayed

BSekula:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://alfresco.atlassian.net/browse/AAE-6366

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/core/pipes/user-initial.pipe.spec.ts
+++ b/lib/core/pipes/user-initial.pipe.spec.ts
@@ -16,6 +16,7 @@
  */
 
 import { DomSanitizer } from '@angular/platform-browser';
+import { IdentityUserModel } from 'core/models/identity-user.model';
 import { UserProcessModel } from '../models/user-process.model';
 import { InitialUsernamePipe } from './user-initial.pipe';
 
@@ -85,6 +86,17 @@ describe('UserInitialPipe', () => {
         fakeUser.firstName = undefined;
         fakeUser.lastName = 'FAKE-SURNAME';
         const result = pipe.transform(fakeUser);
+        expect(result).toBe('<div id="user-initials-image" class="">F</div>');
+    });
+
+    it('should return a single letter from username if we do not have first and last name', () => {
+        const fakeIdentityUser: IdentityUserModel = {
+            firstName: undefined,
+            lastName: undefined,
+            username: 'FAKE-USERNAME'
+        };
+
+        const result = pipe.transform(fakeIdentityUser);
         expect(result).toBe('<div id="user-initials-image" class="">F</div>');
     });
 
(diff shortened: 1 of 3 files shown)


lib/core/pipes/user-initial.pipe.spec.ts
```diff
@@ -88,6 +89,17 @@ describe('UserInitialPipe', () => {
...
 
+    it('should return a single letter from username if we do not have first and last name', () => {
+        const fakeIdentityUser: IdentityUserModel = {
+            firstName: undefined,
+            lastName: undefined,
+            username: 'FAKE-USERNAME'
+        };
+
+        const result = pipe.transform(fakeIdentityUser);
+        expect(result).toBe('<div id="user-initials-image" class="">F</div>');
```
mauriziovitale, line 100: what if the username is lowercase ?

lib/core/pipes/user-initial.pipe.ts
```diff
@@ -34,15 +34,25 @@ export class InitialUsernamePipe implements PipeTransform {
...
+                delimiter,
+                user.username
+            );
             safeHtml = this.sanitized.bypassSecurityTrustHtml(`<div id="user-initials-image" class="${className}">${initialResult}</div>`);
         }
         return safeHtml;
     }
 
-    getInitialUserName(firstName: string, lastName: string, delimiter: string): string {
+    getInitialUserName(firstName: string, lastName: string, delimiter: string, username?: string): string {
```
mauriziovitale, line 48: I don't like this solution very much. 
Why don't we create a getInitials for unsername and one for firstnameLastname ?

mauriziovitale: another point that is not clear to me is from where this displayName comes from, I know that it was already there... but is very confusing

mauriziovitale (changes requested):
comments

lib/core/pipes/user-initial.pipe.spec.ts
BSekula: Transform text to uppercase (for `firstName` and `lastName` also)

lib/core/pipes/user-initial.pipe.ts
BSekula: Refactored the code.

We can see now exactly from where we are getting these properties.

Other solution would be just to add:

```
interface User {
    firstName?: string;
    lastName?: string;
    displayName?: string;
    username?: string;
}
```

lib/core/pipes/user-initial.pipe.spec.ts
mauriziovitale: ok my point is to have ```username: 'fake-username'```

PR #7564, opened 2022-03-24
[AAE-7819] Change column order for processes and tasks datatable 

BSekula:

> - [ ] Bugfix
> - [x] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://alfresco.atlassian.net/browse/AAE-7819

Added possibility to change order of columns for processes and task tables

- Added possibility to drag&drop columns in `adf-datatable` (datatable issuing an event after changing column order - columnOrderChanged).

- Added loading/saving columns order using preferences service for `process-list-cloud-ext.component` and `base-task-list-cloud.component`

- Preferences are saved using services which implements `PreferenceCloudServiceInterface` (currently we have two such a preferences, `LocalPreferenceCloudService` which saves preferences in localstorage and `UserPreferenceCloudService` - saves preferences on BE side)

- Columns order can be set in `DataTableSchema`, after setting the `DataTableSchema.columnsOrder`, specific order is applied to columns while creating `datatable` schema (`DataTableSchema.createDatatableSchema`).

> - [ ] Yes
> - [x] No

**Other information**:


lib/core/datatable/components/datatable/datatable.component.scss
```diff
@@ -315,66 +331,6 @@ $data-table-cell-min-width-file-size: $data-table-cell-min-width !default;
         }
     }
 
-    .adf-datatable-cell-header {
```
BSekula, line 318: While dragging header column, column is taken out from the context of data-table therefore moved this class outside of `adf-datatable-list`

lib/core/datatable/data/data-column.model.ts
```diff
@@ -44,4 +45,5 @@ export interface DataColumn {
     focus?: boolean;
     sortingKey?: string;
     header?: TemplateRef<any>;
+    draggable?: boolean;
```
BSekula, line 48: Here I am not sure if would be better to just enable drag and drop for specific column or just enable it for whole table at once, currently we can enable it for each column separately (@VitoAlbano if you think it is not necessary I can change it, should be quick change)

lib/process-services-cloud/src/lib/process/process-list/components/process-list-cloud.component.ts
```diff
@@ -210,7 +222,17 @@ export class ProcessListCloudComponent extends DataTableSchema implements OnChan
     }
 
     ngAfterContentInit() {
-        this.createDatatableSchema();
+        this.isLoadingColumnsOrder = true;
+        this.loadColumnsOrderPreferences()
```
BSekula, line 226: loading column order from preferences service. Other solution would be to pass `columnOrder` from parent component (e.g. in alfresco-apps/process-list-cloud-ext.component) but we would need to delay displaying of the datatable body e.g. by passing flag in order to showing the loading spinner (loading spinner is disappear right after loading processes).

PR #7567, opened 2022-04-04
[AAE-7819] Change column order - enable drag and drop for datatable - [1/3]

BSekula:
**Please check if the PR fulfills these requirements**

> - [ ] The commit message follows our [guidelines](https://github.com/Alfresco/alfresco-ng2-components/wiki/Commit-format)
> - [x] Tests for the changes have been added (for bug fixes / features)
> - [ ] Docs have been added / updated (for bug fixes / features)

**What kind of change does this PR introduce?** (check one with "x")

> - [ ] Bugfix
> - [x] Feature
> - [ ] Code style update (formatting, local variables)
> - [ ] Refactoring (no functional changes, no api changes)
> - [ ] Build related changes
> - [ ] Documentation
> - [ ] Other... Please describe:

**What is the current behaviour?** (You can also link to an open issue here)
https://alfresco.atlassian.net/browse/AAE-7819

**What is the new behaviour?**
This is the first part for [AAE-7819](https://alfresco.atlassian.net/browse/AAE-7819) (Allowing to change column order for processes and task table)

This PR allowing to drag and drop `datatable` column header and issuing an event for changing column order - `columnOrderChanged`

In order to enable drag and drop for column, we need to set this column as draggable.

Screenshot 2022-04-04 at 13 57 48

**Does this PR introduce a breaking change?** (check one with "x")

> - [ ] Yes
> - [x] No

If this PR contains a breaking change, please describe the impact and migration path for existing applications: ...

**Other information**:

Other PRs:
https://github.com/Alfresco/alfresco-ng2-components/pull/7567 [1/3] - Enable drag&drop for data-table
https://github.com/Alfresco/alfresco-ng2-components/pull/7568 [2/3] - saving/loading preferences for processes
https://github.com/Alfresco/alfresco-ng2-components/pull/7569 [3/3]  - saving/loading preferences for tasks

PR #7568, opened 2022-04-04
[AAE-7819] Change column order - load and save columns order preferences for PROCESSES - [2/3]

BSekula:

> - [ ] Bugfix
> - [x] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://alfresco.atlassian.net/browse/AAE-7819

This is the second part for [AAE-7819](https://alfresco.atlassian.net/browse/AAE-7819) (Allowing to change column order for processes and task table)

In this PR I am saving and loading columns order using preferences service.

Preferences are saved using services which implements PreferenceCloudServiceInterface (currently we have two such a preferences, LocalPreferenceCloudService which saves preferences in `localstorage` and UserPreferenceCloudService - saves preferences on BE side)

> - [ ] Yes
> - [x] No

**Other information**:
Other PRs:
https://github.com/Alfresco/alfresco-ng2-components/pull/7567 [1/3] - Enable drag&drop for data-table
https://github.com/Alfresco/alfresco-ng2-components/pull/7568 [2/3] - saving/loading preferences for processes
https://github.com/Alfresco/alfresco-ng2-components/pull/7569 [3/3]  - saving/loading preferences for tasks

### DIFF
--- a/lib/process-services-cloud/src/lib/process-services-cloud.module.ts
+++ b/lib/process-services-cloud/src/lib/process-services-cloud.module.ts
@@ -27,7 +27,9 @@ import {
     LocalPreferenceCloudService,
     PreferenceCloudServiceInterface,
     PROCESS_FILTERS_SERVICE_TOKEN,
-    TASK_FILTERS_SERVICE_TOKEN
+    TASK_FILTERS_SERVICE_TOKEN,
+    PROCESS_LISTS_PREFERENCES_SERVICE_TOKEN,
+    TASK_LIST_PREFERENCES_TOKEN
 } from './services/public-api';
 import { PeopleCloudModule } from './people/people-cloud.module';
 import { CloudFormRenderingService } from './form/components/cloud-form-rendering.service';
@@ -67,7 +69,10 @@ import { ProcessServicesCloudPipeModule } from './pipes/process-services-cloud-p
     ]
 })
 export class ProcessServicesCloudModule {
-    static forRoot(preferenceServiceInstance?: PreferenceCloudServiceInterface): ModuleWithProviders<ProcessServicesCloudModule> {
+    static forRoot(
+        filterPreferenceServiceInstance?: PreferenceCloudServiceInterface,
+        listPreferenceServiceInstance?: PreferenceCloudServiceInterface
+    ): ModuleWithProviders<ProcessServicesCloudModule> {
         return {
             ngModule: ProcessServicesCloudModule,
             providers: [
@@ -79,20 +84,19 @@ export class ProcessServicesCloudModule {
                         source: 'assets/adf-process-services-cloud'
                     }
                 },
-                { provide: PROCESS_FILTERS_SERVICE_TOKEN, useExisting: preferenceServiceInstance ?? LocalPreferenceCloudService },
-                { provide: TASK_FILTERS_SERVICE_TOKEN, useExisting: preferenceServiceInstance ?? LocalPreferenceCloudService },
+                { provide: PROCESS_FILTERS_SERVICE_TOKEN, useExisting: filterPreferenceServiceInstance ?? LocalPreferenceCloudService },
+                { provide: TASK_FILTERS_SERVICE_TOKEN, useExisting: filterPreferenceServiceInstance ?? LocalPreferenceCloudService },
+                { provide: PROCESS_LISTS_PREFERENCES_SERVICE_TOKEN, useExisting: listPreferenceServiceInstance ?? LocalPreferenceCloudService },
+                { provide: TASK_LIST_PREFERENCES_TOKEN, useExisting: listPreferenceServiceInstance ?? LocalPreferenceCloudService },
                 FormRenderingService,
                 { provide: FormRenderingService, useClass: CloudFormRenderingService }
             ]
         };
     }
 
-    static forChild(preferenceServiceInstance?: PreferenceCloudServiceInterface): ModuleWithProviders<ProcessServicesCloudModule> {
+    static forChild(): ModuleWithProviders<ProcessServicesCloudModule> {
         return {
-            ngModule: ProcessServicesCloudModule,
-            providers: [
-                { provide: PROCESS_FILTERS_SERVICE_TOKEN, useExisting: preferenceServiceInstance ?? LocalPreferenceCloudService }
-            ]
+            ngModule: ProcessServicesCloudModule
         };
     }
 }
--- a/lib/process-services-cloud/src/lib/process-services-cloud.module.ts
+++ b/lib/process-services-cloud/src/lib/process-services-cloud.module.ts
@@ -87,9 +87,12 @@ export class ProcessServicesCloudModule {
         };
     }
 
-    static forChild(): ModuleWithProviders<ProcessServicesCloudModule> {
+    static forChild(preferenceServiceInstance?: PreferenceCloudServiceInterface): ModuleWithProviders<ProcessServicesCloudModule> {
         return {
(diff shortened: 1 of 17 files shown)


lib/process-services-cloud/src/lib/process-services-cloud.module.ts
```diff
@@ -87,9 +87,12 @@ export class ProcessServicesCloudModule {
         };
     }
 
-    static forChild(): ModuleWithProviders<ProcessServicesCloudModule> {
+    static forChild(preferenceServiceInstance?: PreferenceCloudServiceInterface): ModuleWithProviders<ProcessServicesCloudModule> {
```
BSekula, line 90: We need to be able to provide preferences `UserPreferenceCloudService` service.

This is done in the same way as for root: https://github.com/Alfresco/alfresco-ng2-components/pull/7568/files#diff-d7355a313d50b1e32db0269c4a1e95720dde82f661ce21fdb6b84a5e29133c39R69-R89

BSekula: I removed it in next PR in part [3/3], so it stays as it was

https://github.com/Alfresco/alfresco-ng2-components/pull/7569/files#diff-d7355a313d50b1e32db0269c4a1e95720dde82f661ce21fdb6b84a5e29133c39R97-R99

VitoAlbano, line 90: This seems a place for a simple InjectionToken whenever you have to pass a service that is needed, and it is a little confusing adding/removing the providing

lib/process-services-cloud/src/lib/process/process-list/components/process-list-cloud.component.ts
```diff
@@ -188,15 +192,23 @@ export class ProcessListCloudComponent extends DataTableSchema implements OnChan
     skipCount: number = 0;
     currentInstanceId: string;
     selectedInstances: any[];
-    isLoading = true;
+    isLoadingProcesses = true;
+    isLoadingColumnsOrder = true;
+
+    get isLoading(): boolean {
+        return this.isLoadingProcesses || this.isLoadingColumnsOrder;
```
VitoAlbano, line 199: Can we unify this isLoading?

lib/process-services-cloud/src/lib/services/user-preference-cloud.service.ts
```diff
@@ -37,10 +38,19 @@ export class UserPreferenceCloudService extends BaseCloudService implements Pref
...
+  getPreferences(appName: string, key?: string): Observable<any> {
     if (appName) {
       const url = `${this.getBasePath(appName)}/preference/v1/preferences`;
-      return this.get(url);
+
+      return this.get(url).pipe(map((preferences: any) => {
+        if (key) {
+          const preferencesList = preferences?.list?.entries ?? [];
+          const searchedPreferences = preferencesList.find(preference => preference.entry.key === key);
+          return searchedPreferences ? JSON.parse(searchedPreferences.entry.value) : null;
```
VitoAlbano, line 49: This seems a little bit confusing, maybe wrap this information in a model would help getting what are we trying to populate here and will help having a single return to identify what is returned.

VitoAlbano (changes requested):

lib/process-services-cloud/src/lib/process-services-cloud.module.ts
BSekula: Hi @VitoAlbano, sorry I forgot to mention it here, I reverted this changes in next PR (I should revert it in this PR here though)

https://github.com/Alfresco/alfresco-ng2-components/pull/7569/files#r842588171

BSekula: I mentioned it in this PR :)

https://github.com/Alfresco/alfresco-ng2-components/pull/7568#discussion_r842608834

lib/process-services-cloud/src/lib/process/process-list/components/process-list-cloud.component.ts
BSekula: I though about it but could not find any good solution.

We need to wait with displaying table until we can create `columns schema` and after all `processes` are loaded.

`Columns schema` consist now from two factors:  `columns order` and `columns` itself.

`columns` are provided using `transclusion`, therefor creating schema should be done in `ngAfterContentInit` (and the same for loading columns order, since it is needed for creating schema)

From the other hand, processes are loaded after change of property in `ngChange` (`columns schema` is loaded only once).

In order to unify this flag I should run this both actions (load processes and create schema) in one place e.g. `ngChange` (but due to transclusion, creating schema should be done in `ngAfterContentInit`).

lib/process-services-cloud/src/lib/services/user-preference-cloud.service.ts
BSekula: It was my error (reverted it), unfortunately we are returning here `any` which does not help with inferring what we should return.

When I checked how this is implemented in `local-preference-cloud.service` (these both services implements the same interface `PreferenceCloudServiceInterface` and we should be able to use them interchangeably)

we are returning structure like:

```
{
  list: {
   entries: [ {
          entry: {
            key,
            value: 'MY VALUE'
          }
       }]
   }
};
```
Since we are using `PreferenceCloudServiceInterface` with method signature that looks like: `getPreferences(appName: string, key?: string): Observable<any>;`

I would thought that after passing the `key` e.g. `my-preferences` I am going to get preferences for that `key` and I do not have to perform search every time.

Thats why I created `searchPreferences` that is going to return exactly what we need: https://github.com/Alfresco/alfresco-ng2-components/pull/7569/files#diff-b04373b5164c2ea36b539499ce6aab959e51532f2b379cb6b65734739028a88bR27 (I will change `key: any` to list of keys if I need to get more preferences as  'key: string[]')

BSekula: I see that searchPreferences causing some confusion, therefore I removed it and used `getPreferences`.
We can think about updating the preferences service some other time :)

lib/process-services-cloud/src/lib/process/process-list/components/process-list-cloud.component.ts
VitoAlbano: You can catch the event n the latter and put all on Loading in the meantime, it makes sense to have it loading and showing the table once if finished to have its loading

lib/process-services-cloud/src/lib/services/user-preference-cloud.service.ts
VitoAlbano: If that helps the clarity we could define a model to wrap this entity

lib/process-services-cloud/src/lib/process/process-list/components/process-list-cloud.component.ts
BSekula: unified `isLoading` flag with `isColumnSchemaCreated$` observable

lib/process-services-cloud/src/lib/services/user-preference-cloud.service.ts
BSekula: looks like different services implements it in different way, so it should be generic, currently I will skip it since this is not in scope of this story (basically I am lack of time :confused:)

BSekula: I am getting some issues while applying this model, will leave it right now, there need to be more changes for type this properly.

PR #7630, opened 2022-05-12
[AAE-7856] Show process variables in table

BSekula:

> - [ ] Bugfix
> - [x] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://alfresco.atlassian.net/browse/AAE-7856

Allow to show process variables in datatable
Screenshot 2022-05-22 at 14 05 31

In order mark variable to be displayable in data-table, variable need to be marked in modelling app using specific flag:
Screenshot 2022-05-22 at 14 10 21

Processes are grouped by `display name` property, so if two processes have the same variable display name set in modelling app, these variables are going to be shown in the same column 

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/docs/core/components/data-column.component.md
+++ b/docs/core/components/data-column.component.md
@@ -21,6 +21,7 @@ Defines column properties for DataTable, Tasklist, Document List and other compo
     -   [Column Template](#column-template)
     -   [Styling Techniques](#styling-techniques)
     -   [Using the copyContent option](#using-the-copycontent-option)
+    -   [Exapmple of column customData](#example-of-column-customData)
 -   [See also](#see-also)
 
 ## Basic Usage
@@ -52,6 +53,7 @@ Defines column properties for DataTable, Tasklist, Document List and other compo
 | formatTooltip | `Function` |  | Custom tooltip formatter function. |
 | key | `string` |  | Data source key. Can be either a column/property key like `title`  or a property path like `createdBy.name`. |
 | sortable | `boolean` | true | Toggles ability to sort by this column, for example by clicking the column header. |
+| customData | `Generic` | any | Any feature specific data |
 | draggable | `boolean` | false | Toggles drag and drop for header column. |
 | isHidden | `boolean` | false | Hides columns |
 | sortingKey | `string` |  | When using server side sorting the column used by the api call where the sorting will be performed |
@@ -349,6 +351,35 @@ HTML `<data-column>` element example:
         ...
     </data-columns>
 </adf-tasklist>
+```
+
+### Example of column customData
+
+If you would like to pass any custom data related to your specific feature, you can use customData
+
+HTML `<data-column>` element example:
+
+```html
+<data-column [customData]="MyCustomData" key="id" title="Id"></data-column>
+```
+
+You can use generic type for `DataColumn` in order to get intellisense working e.g.
+
+```ts
+const dataColumn: DataColumn<{ shouldPerformActionIfDisplayed: boolean }> = {
+    ...
+    customData: { shouldPerformActionIfDisplayed: true }
+}
+
+// We should get proper types
+consol.log(dataColumn.customData.shouldPerformActionIfDisplayed);
+
+// Now we can use this data in our feature e.g.
+const shouldPerformAction = this.columns
+    .filter(column => column.isHidden)
+    .some(column => column.customData?.shouldPerformActionIfDisplayed === true);
+
+if (shouldPerformAction) { /* action */}
 ```
 
 ## See also
--- a/lib/content-services/src/lib/document-list/data/share-datatable-adapter.ts
+++ b/lib/content-services/src/lib/document-list/data/share-datatable-adapter.ts
@@ -70,6 +70,10 @@ export class ShareDataTableAdapter implements DataTableAdapter {
         this.allowDropFiles = allowDropFiles;
     }
 
+    getColumnType(_row: DataRow, col: DataColumn): string {
+        return col.type;
+    }
+
     getRows(): Array<DataRow> {
         return this.rows;
     }
--- a/lib/core/data-column/data-column.component.ts
+++ b/lib/core/data-column/data-column.component.ts
@@ -34,6 +34,10 @@ export class DataColumnComponent implements OnInit {
     @Input()
     key: string;
 
+    /** You can specify any custom data which can be used by any specific feature */
+    @Input()
+    customData: any;
+
     /** Value type for the column. Possible settings are 'text', 'image',
      * 'date', 'fileSize', 'location', and 'json'.
      */
@@ -51,6 +55,10 @@ export class DataColumnComponent implements OnInit {
     /* Enable drag and drop for header column */
     @Input()
     draggable: boolean = false;
+
+    /* Hide column */
+    @Input()
+    isHidden: boolean = false;
 
     /** Display title of the column, typically used for column headers. You can use the
      * i18n resource key to get it translated automatically.
--- a/lib/core/datatable/components/columns-selector/columns-selector.component.html
+++ b/lib/core/datatable/components/columns-selector/columns-selector.component.html
@@ -24,20 +24,22 @@
             [placeholder]='"ADF-DATATABLE.COLUMNS_SELECTOR.SEARCH" | translate'>
     </div>
 
-    <ng-container *ngFor="let column of columnItems">
-        <div
-            *ngIf="(column.title | translate | filterString:searchQuery) as translatedTitle"
-            class="adf-columns-selector-list-item-container">
-            <mat-checkbox
-                color="primary"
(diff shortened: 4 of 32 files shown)


docs/core/components/data-column.component.md
```diff
@@ -51,7 +52,9 @@ Defines column properties for DataTable, Tasklist, Document List and other compo
 | format | `string` |  | Value format (if supported by the parent component), for example format of the date. |
 | formatTooltip | `Function` |  | Custom tooltip formatter function. |
 | key | `string` |  | Data source key. Can be either a column/property key like `title`  or a property path like `createdBy.name`. |
+| keys | `MultipleDataColumnKey` |  | Multiple keys which can be used for searching specific vlaue (similar as for `key`). |
```
mauriziovitale, line 55: typo value
are we going to deprecate the key?

docs/core/components/data-column.component.md
```diff
@@ -351,6 +354,40 @@ HTML `<data-column>` element example:
...
+### Using the `keys` option
+
+`keys` options can be useful if your data object contains different properties and you would like to show them in one column:
+
+```json
+[
+    {"type": "text", "key": "name", "title": "name"},
+    {
+        "type": "text",
+        "key": "",
```
mauriziovitale, line 366: why do u need to pass an empty key? is it mandatory ?

lib/process-services-cloud/src/lib/process/process-list/components/process-list-cloud.component.ts
```diff
@@ -359,6 +360,7 @@ export class ProcessListCloudComponent extends DataTableSchema implements OnChan
         }, {});
 
         this.createColumns();
+        this.reload();
```
mauriziovitale, line 363: why are we calling a reload?

docs/core/components/data-column.component.md
BSekula: Yes, the `key` is mandatory right now:
Screenshot 2022-05-23 at 11 46 17

BSekula:
One more change, I added new var for column-data: `customData`  which is generic and can be used for specific feature, in this case I need to assign specific column with array of process variable ids

docs/core/components/data-column.component.md
eromano: looks like we have key and keys that are in some way similar responsibility should we deprecate one ? or just make key accept string or array?

lib/core/data-column/data-column.component.ts
```diff
@@ -52,6 +65,10 @@ export class DataColumnComponent implements OnInit {
     @Input()
     draggable: boolean = false;
 
+    /* Hide column */
+    @Input()
```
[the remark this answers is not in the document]
BSekula: Already added https://github.com/Alfresco/alfresco-ng2-components/pull/7630/files#diff-7902db12d9f0de2a8d431c599b63b1f93068c47ff608e08507fcf8fd5a3e17b5R60 (I split this feature in to few smaller, not sure why this input was not here)

docs/core/components/data-column.component.md
BSekula: Yes, I though about making the key `string | array`, the reason why I did not do it is that it could be possible breaking change (even I think it should be fine for e.g. alfresco-apps and only updates would be done in ng2)

Or for deprecation the `key`, when it will be time to remove it, we would have to update e.g. `lib/search/components/search-filter-container/search-filter-container.component.ts`

Screenshot 2022-05-24 at 09 20 41

to e.g. 
```
this.searchFilterQueryBuilder.getCategoryForColumn(this.col.keys[0]);
```

What do you think @mauriziovitale  @eromano? I am thinking more about the `string | array` approach, but do not have very strong opinion about it.

BSekula: from the other hand, maybe would be cleaner to have one type kor keys 🤔

BSekula: Will update the approach so we would use only `key` instead of `keys`

BSekula:
Updated the approach, currently we are using only `key` in datatable. 

Rows model (only for processes) contains specific values for variables e.g.:

```
rowViewModel = {
  ...,
  variablesMap: {
      myColumn: { value: 1 },
     [variableDisplayName]: { [variableData] }
  }
}
```

And column are going to have `key` like: `variablesMap.[variableDisplayName].value (variablesMap.myColumn.value)`

In order to create the `rowViewModel` with `variablesMap`, I am using the generic `customType` from column `DataColumn<{ assignedVariableDefinitionIds: ['variableId']}>` since I need to know which variable is assigned to which column.

lib/core/datatable/components/columns-selector/columns-selector.component.html
```diff
@@ -24,20 +24,22 @@
...
-                [checked]="!column.isHidden"
-                (change)="changeColumnVisibility(column)">
-                <div class="adf-columns-selector-list-content">{{translatedTitle}}</div>
-            </mat-checkbox>
-        </div>
-    </ng-container>
+    <div class="adf-columns-selector-list-container">
+        <ng-container *ngFor="let column of columnItems">
+            <div
+                *ngIf="(column.title | translate | filterString:searchQuery) as translatedTitle"
```
DenysVuika, line 30: I think there's too many angular pipes used for the "ngIf" here... are they all needed? note that those pipes are called inside the "ngFor" above, so might have performance hit

lib/content-services/src/lib/document-list/data/share-datatable-adapter.ts
```diff
@@ -70,6 +70,10 @@ export class ShareDataTableAdapter implements DataTableAdapter {
         this.allowDropFiles = allowDropFiles;
     }
 
+    getType(_row: DataRow, col: DataColumn): string {
```
DenysVuika, line 73: what's the reason behind this method? you provide a useless parameter `_row` just to get the column type? maybe worth calling the method `getColumnType` and using only the `col` param?

lib/core/datatable/data/data-column.model.ts
```diff
@@ -29,7 +29,7 @@ export interface DataColumnTypes {
 
 export type DataColumnType = keyof DataColumnTypes;
 
-export interface DataColumn {
+export interface DataColumn<CustomData = unknown> {
```
DenysVuika, line 32: traditionally this kind of generics is using `T`, also, not sure if that's not a breaking change for the apps that are using this interface...

lib/core/datatable/data/data-table.schema.ts
```diff
@@ -24,7 +24,7 @@ import { ObjectDataColumn } from './object-datacolumn.model';
 
 @Directive()
 // eslint-disable-next-line @angular-eslint/directive-class-suffix
-export abstract class DataTableSchema {
+export abstract class DataTableSchema<CustomColumnData = unknown> {
```
DenysVuika, line 27: should better be `T` here, also can you check if this is not a breaking change for existing apps?

lib/core/datatable/data/datatable-adapter.ts
```diff
@@ -29,6 +29,7 @@ export interface DataTableAdapter {
     getColumns(): Array<DataColumn>;
     setColumns(columns: Array<DataColumn>): void;
     getValue(row: DataRow, col: DataColumn, resolverFn?: (_row: DataRow, _col: DataColumn) => any): any;
+    getType(row: DataRow, col: DataColumn): string;
```
DenysVuika, line 32: I'm sure this is a breaking change, why this is needed compared to just using `col.type` in the HTML like before?

lib/core/datatable/data/object-datatable-adapter.ts
```diff
@@ -77,6 +77,10 @@ export class ObjectDataTableAdapter implements DataTableAdapter {
         this.rowsChanged = new Subject<Array<DataRow>>();
     }
 
+    getType(_row: DataRow, col: DataColumn): string {
```
DenysVuika, line 80: as above, does not really make sense introducing a method with parameters it does not even use

lib/process-services-cloud/src/lib/models/column-data-type.model.ts
```diff
@@ -0,0 +1,4 @@
+// eslint-disable-next-line no-shadow
+export enum ColumnDataType {
+    processVariableColumn = 'process-variable-column'
```
DenysVuika, line 3: enums with a single entry can be replaced by constants I presume

DenysVuika (changes requested):

DenysVuika (approved):

lib/core/datatable/data/data-table.schema.ts
BSekula: As discussed, updated to `T`.

PR #7632, opened 2022-05-12
[AAE-8639] Discovery OpenId - Load discovery and pass info to jsapi

mauriziovitale:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

> - [ ] Yes
> - [ ] No

**Other information**:

### DIFF
--- a/lib/core/i18n/en.json
+++ b/lib/core/i18n/en.json
@@ -186,7 +186,8 @@
       "SILENT": "Silent Login",
       "SCOPE": "Scope",
       "CLIENT": "Client ID",
-      "PUBLIC_URLS": "Public urls silent Login"
+      "PUBLIC_URLS": "Public urls silent Login",
+      "SECRET": "Secret"
     },
     "CARDVIEW": {
       "KEYVALUEPAIRS": {
--- a/lib/core/settings/host-settings.component.html
+++ b/lib/core/settings/host-settings.component.html
@@ -103,6 +103,15 @@
                         </mat-error>
                     </mat-form-field>
 
+                    <mat-form-field class="adf-full-width">
+                        <mat-label>{{ 'CORE.HOST_SETTINGS.SECRET'| translate }}</mat-label>
+                        <input matInput name="{{ 'CORE.HOST_SETTINGS.SECRET'| translate }}"
+                               formControlName="secret" placeholder="{{ 'CORE.HOST_SETTINGS.SECRET'| translate }}">
+                        <mat-error *ngIf="secret.hasError('required')">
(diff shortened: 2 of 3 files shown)


lib/core/services/alfresco-api.service.ts
```diff
@@ -79,15 +109,21 @@ export class AlfrescoApiService {
...
+    }
+
+    private initAlfrescoApiWithLastConfig() {
+        if (this.alfrescoApi && this.isDifferentConfig(this.lastConfig, this.currentAppConfig)) {
+            this.lastConfig = this.currentAppConfig;
+            this.alfrescoApi.setConfig(this.currentAppConfig);
         } else {
-            this.lastConfig = config;
-            this.alfrescoApi = new AlfrescoApi(config);
+            this.lastConfig = this.currentAppConfig;
```
arditdomi, line 124: you can put this code before the if else statement so you write it only once

lib/core/services/alfresco-api.service.ts
```diff
@@ -79,15 +109,21 @@ export class AlfrescoApiService {
...
-            this.lastConfig = config;
-            this.alfrescoApi.setConfig(config);
+    protected initAlfrescoApi() {
+        this.getCurrentAppConfig();
+        this.initAlfrescoApiWithLastConfig();
+    }
+
+    private initAlfrescoApiWithLastConfig() {
+        if (this.alfrescoApi && this.isDifferentConfig(this.lastConfig, this.currentAppConfig)) {
+            this.lastConfig = this.currentAppConfig;
```
arditdomi, line 121: the method name says initAlfrescoApiWithLastConfig but in practice is initialising with the current config because you are doing  `this.lastConfig = this.currentAppConfig;`

lib/core/services/alfresco-api.service.ts
```diff
@@ -36,6 +36,9 @@ export class AlfrescoApiService {
     protected alfrescoApi: AlfrescoApi;
 
     lastConfig: AlfrescoApiConfig;
+    currentAppConfig: AlfrescoApiConfig;
+
+    idpConfig: any;
```
arditdomi, line 41: can it be of type `OpenidConfiguration` instead of any ?

eromano (approved):

popovicsandras (approved):

lib/core/services/alfresco-api.service.ts
mauriziovitale: I cannot put it before the if otherwise, they will never be different. Let me check why we

arditdomi (approved):

PR #7641, opened 2022-05-18
[AAE-8641] Ability to use people widget as part of a form

tomgny:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

> - [ ] Yes
> - [ ] No

**Other information**:

### DIFF
--- a/lib/core/services/identity-user.service.interface.ts
+++ b/lib/core/services/identity-user.service.interface.ts
@@ -16,10 +16,13 @@
  */
 
 import { Pagination } from '@alfresco/js-api';
+import { InjectionToken } from '@angular/core';
 import { Observable } from 'rxjs';
 import { IdentityGroupModel } from '../models/identity-group.model';
 import { IdentityRoleModel } from '../models/identity-role.model';
 import { IdentityUserModel } from '../models/identity-user.model';
+
+export const IDENTITY_USER_TOKEN = new InjectionToken<IdentityUserServiceInterface>('identity-user-token');
 
 export interface IdentityUserQueryResponse {
 
--- a/lib/core/services/identity-user.service.interface.ts
+++ b/lib/core/services/identity-user.service.interface.ts
@@ -16,13 +16,10 @@
  */
 
 import { Pagination } from '@alfresco/js-api';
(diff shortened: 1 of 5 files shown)


lib/core/services/identity-user.service.interface.ts
```diff
@@ -16,11 +16,14 @@
  */
 
 import { Pagination } from '@alfresco/js-api';
+import { InjectionToken } from '@angular/core';
 import { Observable } from 'rxjs';
 import { IdentityGroupModel } from '../models/identity-group.model';
 import { IdentityRoleModel } from '../models/identity-role.model';
 import { IdentityUserModel } from '../models/identity-user.model';
 
+export const IDENTITY_USER_TOKEN = new InjectionToken<IdentityUserServiceInterface>('identity-user-token');
```
arditdomi, line 25: can you put this in a new file? identity-user-service.token.ts for example? Also the `identity-user-token` should be `identity-user-service-token` ?

lib/process-services-cloud/src/lib/people/components/people-cloud.component.ts
```diff
@@ -160,7 +161,8 @@ export class PeopleCloudComponent implements OnInit, OnChanges, OnDestroy {
     searchLoading = false;
 
     constructor(
-        private identityUserService: IdentityUserService,
+        @Inject(IDENTITY_USER_TOKEN)
```
arditdomi, line 164: I guess you can do the same in all the places where we inject the identityUserService. Nothing will change because you will provide IdentityUserService by default but if we need to overwrite it we don't have to do any additional change in ADF

lib/core/services/identity-user.service.interface.ts
tomgny: renamed and moved

lib/process-services-cloud/src/lib/people/components/people-cloud.component.ts
tomgny: As Maurizio suggested, we should do it 1 at time, so we can replace it in other places in the future.

arditdomi (approved):

PR #7682, opened 2022-06-21
Fetch the userInfo once loggedIn and expose the capability of admin

mauriziovitale:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

> - [ ] Yes
> - [ ] No

**Other information**:

### DIFF
--- a/lib/core/services/auth-guard-base.ts
+++ b/lib/core/services/auth-guard-base.ts
@@ -32,6 +32,7 @@ import { OauthConfigModel } from '../models/oauth-config.model';
 import { MatDialog } from '@angular/material/dialog';
 import { StorageService } from './storage.service';
 import { Observable } from 'rxjs';
+import { PeopleContentService } from './people-content.service';
 
 export abstract class AuthGuardBase implements CanActivate, CanActivateChild {
 
@@ -47,7 +48,8 @@ export abstract class AuthGuardBase implements CanActivate, CanActivateChild {
         protected router: Router,
         protected appConfigService: AppConfigService,
         protected dialog: MatDialog,
-        private storageService: StorageService
+        private storageService: StorageService,
+        protected userService: PeopleContentService
     ) {
     }
 
--- a/lib/core/services/auth-guard-base.ts
+++ b/lib/core/services/auth-guard-base.ts
@@ -32,7 +32,6 @@ import { OauthConfigModel } from '../models/oauth-config.model';
 import { MatDialog } from '@angular/material/dialog';
 import { StorageService } from './storage.service';
 import { Observable } from 'rxjs';
-import { UserInfoServiceInterface } from './user.service.interface';
 
 export abstract class AuthGuardBase implements CanActivate, CanActivateChild {
 
@@ -48,8 +47,7 @@ export abstract class AuthGuardBase implements CanActivate, CanActivateChild {
         protected router: Router,
         protected appConfigService: AppConfigService,
         protected dialog: MatDialog,
-        private storageService: StorageService,
-        protected userService: UserInfoServiceInterface
+        private storageService: StorageService
     ) {
     }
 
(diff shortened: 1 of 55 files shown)


lib/core/services/auth-guard-base.ts
```diff
@@ -47,7 +48,8 @@ export abstract class AuthGuardBase implements CanActivate, CanActivateChild {
         protected router: Router,
         protected appConfigService: AppConfigService,
         protected dialog: MatDialog,
-        private storageService: StorageService
+        private storageService: StorageService,
+        protected userService: PeopleContentService
```
arditdomi, line 52: Shall we name it peopleContentService instead of user service?

lib/core/userinfo/components/user-info.component.ts
```diff
@@ -70,7 +70,7 @@ export class UserInfoComponent implements OnInit, OnDestroy {
     selectedIndex: number;
     private destroy$ = new Subject();
 
-    constructor(private ecmUserService: EcmUserService,
+    constructor(private ecmUserService: PeopleContentService,
```
arditdomi, line 73: I suggest we name the service to match the service class name, in this case `peopleContentService`

arditdomi (commented):
You also need to update the docs about the services you changed

lib/core/services/people-content.service.ts
```diff
@@ -61,11 +61,11 @@ export class PeopleContentService {
...
         private apiService: AlfrescoApiService,
         authenticationService: AuthenticationService,
-        private logService: LogService
+        private logService: LogService,
+        private contentService: ContentService
     ) {
         authenticationService.onLogout.subscribe(() => {
-            this.hasCheckedIsContentAdmin = false;
-            this.hasContentAdminRole = false;
+            this.resetLocalCurrentUser();
```
arditdomi, line 68: is there a unit test about this?

arditdomi: oh yes it's covered, looks good

arditdomi (approved):
Good job 👏

lib/core/services/people-content.service.ts
```diff
@@ -16,13 +16,14 @@
...
-import { PersonEntry, PeopleApi, PersonBodyCreate, Pagination, PersonBodyUpdate } from '@alfresco/js-api';
+import { catchError, map, tap } from 'rxjs/operators';
+import { PeopleApi, PersonBodyCreate, Pagination, PersonBodyUpdate } from '@alfresco/js-api';
 import { EcmUserModel } from '../models/ecm-user.model';
 import { LogService } from './log.service';
 import { AuthenticationService } from './authentication.service';
+import { ContentService } from './content.service';
 
 // eslint-disable-next-line no-shadow
 export enum ContentGroups {
```
DenysVuika, line 29: can you just move the enum to a separate file? this will allow not using the eslit-disable hack

lib/core/services/user-info-resolver.service.ts
```diff
@@ -0,0 +1,41 @@
...
+import { PeopleContentService } from './people-content.service';
+
+@Injectable({
+  providedIn: 'root'
+})
+export class UserInfoResolverService implements Resolve<EcmUserModel> {
+  constructor(private peopleContentService: PeopleContentService) {}
+
+  resolve(
+    _route: ActivatedRouteSnapshot,
```
DenysVuika, line 35: you can remove the parameters instead of prefixing them, the JS/TS allow that

DenysVuika (changes requested):

DenysVuika (approved):

mauriziovitale:
@DenysVuika I only had to change the test because affected by the refactoring 
btw we should decide what a unit test of a service should test. 

In the above case, the error is making sure there is a catch and a throw and in my opinion, this should be checked because sometimes the developer forgets to add the catch. 
For the others the unit is mocking the js api an testing the result is correctly mapped (no entry is returned)

PR #7800, opened 2022-08-30
[AAE-10499] create angular based custom http client for alfresco js api

mserwicki:

> - [ ] Bugfix
> - [x] Feature
> - [ ] Build related changes
> - [ ] Documentation

Currently, we use superagent to make all requests inside alfersco/js-api, with prevents us from using angular-specific features, like interceptors that are going to be required to proceed with the next step of adding angular-oauth2-oidc for authentication, as we want to take advantage of them

This PR, do not change anything yet on its own, its creating an http client, that can be passed to AlfrescoApi and replace the superagent. but I did add the AuthBearerInterceptor inside core, with will hijack all requests, and add correct token for most requests, not only the ones for alfresco-js api

```
{ provide: HTTP_INTERCEPTORS, useClass: AuthBearerInterceptor, multi: true }
```

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/core/api/alfresco-api/alfresco-api.http-client.spec.ts
+++ b/lib/core/api/alfresco-api/alfresco-api.http-client.spec.ts
@@ -1,0 +1,313 @@
+/*!
+ * @license
+ * Copyright 2019 Alfresco Software, Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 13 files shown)


lib/core/api/alfresco-api/alfresco-api.http-client.spec.ts
```diff
@@ -0,0 +1,316 @@
...
+                contentType: 'multipart/form-data',
+                accept: 'application/json',
+                returnType: null
+            };
+
+            angularHttpClient.request('http://example.com', requestOptions, securityOptions, emitters);
+            const req = controller.expectOne('http://example.com?autoRename=true&include=allowableOperations');
+            expect(req.request.method).toEqual('POST');
+
+            // filedata: (binary)
```
DenysVuika, line 184: these comments can be removed from the code

DenysVuika (approved):

lib/core/api/alfresco-api/alfresco-api.http-client.spec.ts
mserwicki: oh, nice catch! thank you :)

PR #7814, opened 2022-09-06
[AAE-10457] Storybook User Info module

marcol13:

> - [ ] Bugfix
> - [x] Feature
> - [ ] Build related changes
> - [ ] Documentation

[https://alfresco.atlassian.net/browse/AAE-10457](https://alfresco.atlassian.net/browse/AAE-10457)

Five new stories for UserInfo module in UserInfo component:

- **Login with APS**

image

- **Login with APS and ACS**

image

- **Login with ACS**

image

- **Login with SSO**

image

- **Login with SSO and ACS**

image

Video: 

https://user-images.githubusercontent.com/56632321/189074332-4012d416-07f8-4027-91c8-26e8011c9fd6.mp4

> - [ ] Yes
> - [ ] No

**Other information**:

TODO: 

- Video
- Are pictures really necessary? 
- Move mock files into proper place


lib/core/userinfo/components/mocks/authentication.service.mock.ts
```diff
@@ -0,0 +1,76 @@
...
+ */
+
+import { Inject, Injectable } from '@angular/core';
+import { ReplaySubject } from 'rxjs';
+import { AlfrescoApiService } from './../../../services/alfresco-api.service';
+
+@Injectable({
+    providedIn: 'root'
+})
+export class AuthenticationServiceMock {
```
DudaRobert, line 25: Create a new interface that will have the same methods as the real service, and make both the real and the mock service implement this interface:

**Interface:**
interface

**Real service:**
real

**Mocked service:**
mock

lib/core/userinfo/components/user-info.component.stories.ts
```diff
@@ -0,0 +1,202 @@
...
+                    useClass: AuthenticationServiceMock
+                }
+            ]
+        }),
+        componentWrapperDecorator(
+            (story) =>
+                `<div style="height: 100vh; display: flex; align-items: center; justify-content: center;">
+                    ${story}
+                </div>`
+        )
```
DudaRobert, line 68: We decided not to use componentWrapperDecorators and similar solutions, unless necessary. Can you check if it can be done with `template`?

marcol13: I think that this is the best solution in this case - I need only to center a div to show how some properties are working. `Template` requires from me binding this properties and here I can use only `${story}`.

lib/core/userinfo/components/mocks/authentication.service.mock.ts
marcol13: This looks good, but once again I think that in this case it will not work that well. `AuthenticationServiceMock` here has one additional dependency - `MODE` which is not present in original `AuthenticationService`. Also, then I would have to mock a lot of unused functions and these functions/properties which I need are based on `MODE` variable so probably it will not be used in the future by another tests/stories/etc.

DudaRobert: Okay, let's leave it as is

lib/core/userinfo/components/user-info.component.stories.ts
DudaRobert: 👍

DudaRobert (approved):

tomgny (commented):
Can we avoid to use new images for users avatar? Can we use existing some ones, like normal Alfresco logo?

lib/core/src/lib/userinfo/components/mocks/user.service.mock.ts
```diff
@@ -0,0 +1,76 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {
+    BpmUserModel,
+    EcmUserModel,
+    IdentityUserModel
+} from '@alfresco/adf-core';
```
tomgny, line 22: Can we use relative paths to import those models? They are in the same lib as this file

tomgny (approved):

PR #7832, opened 2022-09-16
[AAE-10314] Fixed unhandled promises in ADF Core

Venrofi:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/core/src/lib/card-view/components/card-view-textitem/card-view-textitem.component.spec.ts
+++ b/lib/core/src/lib/card-view/components/card-view-textitem/card-view-textitem.component.spec.ts
@@ -476,122 +476,93 @@ describe('CardViewTextItemComponent', () => {
             fixture.detectChanges();
         });
 
-        it('should call the isValid method with the edited value', fakeAsync((done) => {
+        it('should call the isValid method with the edited value', fakeAsync(() => {
             fixture.detectChanges();
-            fixture.whenStable().then(() => {
-                fixture.detectChanges();
 
-                spyOn(component.property, 'isValid');
-                updateTextField(component.property.key, 'updated-value');
-                tick(600);
-                fixture.detectChanges();
-                fixture.whenStable().then(() => {
-                    expect(component.property.isValid).toHaveBeenCalledWith('updated-value');
-                    done();
-                });
-            });
+            spyOn(component.property, 'isValid');
+            updateTextField(component.property.key, 'updated-value');
+            tick(600);
+
+            fixture.detectChanges();
+            expect(component.property.isValid).toHaveBeenCalledWith('updated-value');
         }));
 
-        it('should trigger the update event if the editedValue is valid', fakeAsync((done) => {
+        it('should trigger the update event if the editedValue is valid', fakeAsync(() => {
             component.property.isValid = () => true;
-
             fixture.detectChanges();
-            fixture.whenStable().then(() => {
-                fixture.detectChanges();
-                const cardViewUpdateService = TestBed.inject(CardViewUpdateService);
-                const property = { ...component.property };
 
-                spyOn(cardViewUpdateService, 'update');
-                updateTextField(component.property.key, 'updated-value');
-                tick(600);
-                fixture.detectChanges();
-                fixture.whenStable().then(() => {
-                    expect(cardViewUpdateService.update).toHaveBeenCalledWith(property, 'updated-value');
-                    done();
-                });
-            });
+            const cardViewUpdateService = TestBed.inject(CardViewUpdateService);
+            const property = { ...component.property };
+
+            spyOn(cardViewUpdateService, 'update');
+            updateTextField(component.property.key, 'updated-value');
+            tick(600);
+            fixture.detectChanges();
+
+            expect(cardViewUpdateService.update).toHaveBeenCalledWith({
+                ...property,
+                value: 'updated-value'
+            }, 'updated-value');
         }));
 
-        it('should NOT trigger the update event if the editedValue is invalid', fakeAsync((done) => {
+        it('should NOT trigger the update event if the editedValue is invalid', fakeAsync(() => {
             component.property.isValid = () => false;
 
             fixture.detectChanges();
-            fixture.whenStable().then(() => {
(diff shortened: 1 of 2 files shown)


lib/core/src/lib/app-config/app-config.service.spec.ts
```diff
@@ -164,24 +164,23 @@ describe('AppConfigService', () => {
         expect(appConfigService.get('testUrl')).toBe('ftp://localhost:9090');
     });
 
-    it('should load external settings', () => {
-        appConfigService.load().then((config) => {
+    it('should load external settings', async () => {
+        await appConfigService.load().then((config) => {
```
tomgny, line 168: do we need to use `.then()` when we are using `await`?

```suggestion
        await appConfigService.load();
        expect(config).toEqual(mockResponse);
```

lib/core/src/lib/app-config/app-config.service.spec.ts
```diff
@@ -111,12 +111,12 @@ describe('AppConfigService', () => {
...
     });
 
-    it('should stream only the selected attribute changes when using select',  fakeAsync(() => {
+    it('should stream only the selected attribute changes when using select',  async () => {
         appConfigService.config.testProp = true;
-        appConfigService.select('testProp').subscribe((property) => {
-            expect(property).toBeTruthy();
-        });
-    }));
+        const propertyPromise = appConfigService.select('testProp').toPromise();
```
tomgny, line 116: can we try to use `lastValueFrom() / firstValueFrom()`?
[FE unit test best practices](https://alfresco.atlassian.net/wiki/spaces/TECH/pages/510919005/FE+unit+tests+best+practices#Adapting-a-test-written-with-done()-to-use-promises-instead)

Venrofi: Currently, we're using RxJS v6.6.6 and those functions are supported from v7 onwards 😄

arditdomi (commented):
Are u sure about this? Keeping fakeAsync with tick I think is fine, but we have to simplify and refactor the tests. For example I see there are nested `fixture.whenStable().then` and we can start by getting rid of those. Do the errors go away if we simply refactor and get rid of the `fixture.whenStable().then`

Venrofi:
In most cases, the root cause for the problems with Unhandled Promises was this `fixture.whenStable().then` with fakeAsync. It just didn't work correctly, I guess after the Angular upgrades and all the `expect()` was nested inside of those `.then()`. When I refactored it from `fakeAsync` to normal `async` and simplified the `fixture.whenStable().then` to `await fixture.whenStable()` and got rid of the nesting it worked as intended. Now that the `expect()` is appropriately checked, there are some more failures and plenty of tests with no expectations, but that's a topic for another issue that I've already created a ticket for. I mean you can have a combo of fakeAsync and a normal async inside of it as well because I saw one test that utilises that and it gives you the option to use await but that's complicating rather than simplifying to my mind. 
About the `tick()` - in order for some tests to pass the `tick()` was necessary and when I got rid of `fakeAsync` I tried different methods of substituting the `tick()` and this was the best solution. In some tests the `tick()` was unnecessary but I guess someone did it to be sure they'll pass. I've tested the ones I've touched if this await is really needed and in many cases, I could get away without it 😄

Venrofi:
With my changes, there are no more errors regarding Unhandled Promises apart from 3 backend-related tests. I don't know what to do about them really. I think Tom said something about a need to mock a service or something for them.

arditdomi:
> In most cases, the root cause for the problems with Unhandled Promises was this `fixture.whenStable().then` with fakeAsync. It just didn't work correctly, I guess after the Angular upgrades and all the `expect()` was nested inside of those `.then()`. When I refactored it from `fakeAsync` to normal `async` and simplified the `fixture.whenStable().then` to `await fixture.whenStable()` and got rid of the nesting it worked as intended. Now that the `expect()` is appropriately checked, there are some more failures and plenty of tests with no expectations, but that's a topic for another issue that I've already created a ticket for. I mean you can have a combo of fakeAsync and a normal async inside of it as well because I saw one test that utilises that and it gives you the option to use await but that's complicating rather than simplifying to my mind. About the `tick()` - in order for some tests to pass the `tick()` was necessary and when I got rid of `fakeAsync` I tried different methods of substituting the `tick()` and this was the best solution. In some tests the `tick()` was unnecessary but I guess someone did it to be sure they'll pass. I've tested the ones I've touched if this await is really needed and in many cases, I could get away without it 😄

https://github.com/Alfresco/alfresco-ng2-components/pull/7832/files#diff-e98fea480e52047f41b7acc01e473d01f6d07657841ecd8cfb29c3ae2580c6c7R417 I don't think this is a good practice. I suggest to keep fakeAsync in combination with tick(); and remove the whenStable() and let's see how many test cases fail without it.

Venrofi:
I'll try different scenarios and give you an update 😄

lib/core/src/lib/card-view/components/card-view-textitem/card-view-textitem.component.spec.ts
```diff
@@ -476,123 +474,112 @@ describe('CardViewTextItemComponent', () => {
...
-        }));
+            updateTextField(component.property.key, 'updated-value');
+
+            fixture.detectChanges();
+            await fixture.whenStable();
+            expect(component.errors).toBe(expectedErrorMessages);
+        });
 
-        it('should render the error when the editedValue is invalid', fakeAsync((done) => {
+        it('should render the error when the editedValue is invalid', async () => {
```
Venrofi, line 542: I've tested a few options on this test and only my solution works correctly. 
image
image
image
image
image

Venrofi:
I'm testing different solutions and for some, a remedy is just to delete `.then()`, in some cases using `fakeAsync( async () => ) {} )` gets the job done but I've found one test and I feel there will be more of them where only my solution with `await setTimeoutPromise() ` works. I'll check each test and give you the final update 😄@arditdomi

PR #7842, opened 2022-09-20
[AAE-10454] Stories for search-text-input component

DudaRobert:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [x] Documentation
> - [x] Other... Please describe: Storybook stories

[AAE-10454](https://alfresco.atlassian.net/browse/AAE-10454)

Added `search-text-input.component.stories` file with stories for the component

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/core/src/lib/search-text/search-text-input.component.stories.ts
+++ b/lib/core/src/lib/search-text/search-text-input.component.stories.ts
@@ -1,0 +1,195 @@
+/*!
+ * @license
+ * Copyright 2019 Alfresco Software, Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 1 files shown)


lib/core/src/lib/search-text/search-text-input.component.stories.ts
```diff
@@ -0,0 +1,195 @@
...
+            table: { category: 'Actions' }
+        }
+    }
+} as Meta;
+
+const template: Story<SearchTextInputComponent> = (args: SearchTextInputComponent) => ({
+    props: args
+});
+
+export const searchTextInput = template.bind({});
```
BSekula, line 195: just out of curiosity, is this needed? (I saw that also in different stories) Generally speaking we cannot rebind `this` in arrow function (arrow functions do not have its own `this`, it points to [lexical this](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/this))

If we would like to change the `this` I think we can have something like:

```
template = function(args) {...}
```

BSekula (approved):

lib/core/src/lib/search-text/search-text-input.component.stories.ts
DudaRobert: This is the Storybook's way of defining stories and inferring their names - we create a template and export functions binded to it.  [more here](https://storybook.js.org/docs/angular/writing-stories/introduction#using-args)

BSekula: 👌

`Template.bind({}) is a standard JavaScript technique for making a copy of a function. We copy the Template so each exported story can set its own properties on it.
`

PR #7851, opened 2022-09-22
MNT-23143 allow discoveryService withCredentials

nbarithel:
If withCredentials=true in app.config.json, the contentAuth is not used. Therefore, the filter isEcmLoggedIn() will be false. Adding a dedicated check for Kerberos (i.e. withCredentials=true)

### DIFF
--- a/lib/core/src/lib/services/discovery-api.service.ts
+++ b/lib/core/src/lib/services/discovery-api.service.ts
@@ -42,7 +42,9 @@ export class DiscoveryApiService {
     ) {
         this.authenticationService.onLogin
             .pipe(
-                filter(() => this.apiService.getInstance()?.isEcmLoggedIn() || this.authenticationService.isKerberosEnabled()),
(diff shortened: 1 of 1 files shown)


DenysVuika (approved):

DenysVuika (approved):

lib/core/src/lib/services/discovery-api.service.ts
```diff
@@ -42,7 +42,9 @@ export class DiscoveryApiService {
     ) {
         this.authenticationService.onLogin
             .pipe(
```
DenysVuika, line 44: this PR is for master, should be for develop branch

DenysVuika (changes requested):

lib/core/src/lib/services/discovery-api.service.ts
nbarithel: done!

MichalKinas (approved):

DenysVuika (approved):

PR #7900, opened 2022-10-17
AAE-10842: Modified properties viewer widget to handle file input

ehsan-2019:

> - [ ] Bugfix
> - [X] Feature
> - [ ] Build related changes
> - [ ] Documentation

MetaData viewer widget works with only field values of type string.

MetaData viewer widget works with field values of type file too.

> - [ ] Yes
> - [X] No

**Other information**:

### DIFF
--- a/lib/process-services-cloud/src/lib/form/components/widgets/properties-viewer/properties-viewer.widget.ts
+++ b/lib/process-services-cloud/src/lib/form/components/widgets/properties-viewer/properties-viewer.widget.ts
@@ -15,7 +15,7 @@
  * limitations under the License.
  */
 
-import { Component, EventEmitter, Output, ViewEncapsulation } from '@angular/core';
+import { Component, EventEmitter, OnInit, Output, ViewEncapsulation } from '@angular/core';
 import { FormService, WidgetComponent } from '@alfresco/adf-core';
 import { Node } from '@alfresco/js-api';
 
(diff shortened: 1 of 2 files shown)


lib/process-services-cloud/src/lib/form/components/widgets/properties-viewer/properties-viewer.widget.ts
```diff
@@ -47,6 +47,16 @@ export class PropertiesViewerWidgetComponent extends WidgetComponent {
         super(formService);
     }
 
+    ngOnInit(): void {
+        if (this.field &&
```
VitoAlbano, line 51: Why not adding the input property to pass the file id / node id we want to open up? this seems that we are forcing a little bit our way through . We already merged a similar PR but please apply the same changes also to the other one

VitoAlbano (changes requested):

lib/process-services-cloud/src/lib/form/components/widgets/properties-viewer/properties-viewer.widget.ts
ehsan-2019: After performing some investigations and having discussions between us, we decided that the currently implemented approach will fulfill our needs without any problem. 

_P.S._ by design the value coming to this component is an array of files not a single file. any changes to this will require a big code refactoring.

VitoAlbano (approved):

PR #7916, opened 2022-10-21
ACS-3640 a 11 y aca 881740 snackbar messages disappear without option to adjust timing

AleksanderSklorz:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

Some snackbars were autoclosing very quickly and there was not way to close it manually which caused accessibility issue. 

Time of autoclosing of snackbars have been increased to 10 seconds. There is also possibility to close snackbars using X manually. 

> - [x] Yes
> - [ ] No

There was accessibility issue because of missing option to closing notification manually. We added X icon to allow that by default and added showAction boolean flag to control visibility of action. By default it is set to true. If anyone had earlier notification without any action and still does not want to have any action visible then depending on type of notification we need to pass showAction parameter to open notification function or pass showAction field as part of config object parameter of opening function. 
**Other information**:

ACA PR related to that jira: https://github.com/Alfresco/alfresco-content-app/pull/2734

### DIFF
--- a/docs/core/components/snackbar-content.component.md
+++ b/docs/core/components/snackbar-content.component.md
@@ -1,8 +1,8 @@
 ---
-Title: Sorting Picker Component
+Title: Snackbar Content Component
 Added: v5.1.0
 Status: Active
-Last reviewed: 2022-10-31
+Last reviewed: 2022-11-08
 ---
 
 # [Snackbar Content Component](lib/core/src/lib/snackbar-content/snackbar-content.component.ts "Defined in snackbar-content.component.ts")
--- a/docs/core/components/snackbar-content.component.md
+++ b/docs/core/components/snackbar-content.component.md
@@ -1,8 +1,8 @@
 ---
 Title: Sorting Picker Component
-Added: v5.0.0
+Added: v5.1.0
 Status: Active
-Last reviewed: 2022-10-21
+Last reviewed: 2022-10-31
 ---
 
 # [Snackbar Content Component](lib/core/src/lib/snackbar-content/snackbar-content.component.ts "Defined in snackbar-content.component.ts")
@@ -35,6 +35,7 @@ snackBar.openFromComponent(SnackbarContentComponent, {
 |-------------|-----------|---------------|-------------------------------------------------------------------------|
 | actionLabel | `string`  | false         | Displayed action as a text.                                             |
 | actionIcon  | `string`  | false         | Displayed action as an material icon.                                   |
+| actionIconAriaLabel  | `string`  | false         | Sets aria-label attribute for button with icon action.                  |
 | message     | `string`  | false         | Visible snackbar's message for user.                                    |
 | showAction     | `boolean` | false         | True if action should be visible, false in other case.                  |
 | callActionOnIconClick     | `boolean` | false         | True if clicking on icon should to trigger action, false in other case. |
--- a/lib/core/src/lib/snackbar-content/snack-bar-data.ts
+++ b/lib/core/src/lib/snackbar-content/snack-bar-data.ts
@@ -1,6 +1,6 @@
 /*!
  * @license
- * Copyright 2022 Alfresco Software, Ltd.
+ * Copyright 2019 Alfresco Software, Ltd.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/lib/core/src/lib/snackbar-content/snack-bar-data.ts
+++ b/lib/core/src/lib/snackbar-content/snack-bar-data.ts
@@ -18,6 +18,7 @@
 export interface SnackBarData {
     actionLabel?: string;
     actionIcon?: string;
+    actionIconAriaLabel?: string;
     message: string;
     showAction?: boolean;
     callActionOnIconClick?: boolean;
--- a/lib/core/src/lib/snackbar-content/snack-bar-data.ts
+++ b/lib/core/src/lib/snackbar-content/snack-bar-data.ts
@@ -19,4 +19,6 @@ export interface SnackBarData {
     actionLabel?: string;
     actionIcon?: string;
     message: string;
+    showAction?: boolean;
+    callActionOnIconClick?: boolean;
 }
--- a/lib/core/src/lib/snackbar-content/snack-bar-data.ts
+++ b/lib/core/src/lib/snackbar-content/snack-bar-data.ts
@@ -1,0 +1,22 @@
+/*!
+ * @license
+ * Copyright 2022 Alfresco Software, Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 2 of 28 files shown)


lib/core/src/lib/snackbar-content/snackbar-content.component.scss
```diff
@@ -0,0 +1,16 @@
...
+    display: flex;
+    align-items: center;
+    justify-content: space-between;
+
+    .snackbar-message {
+        margin: 0;
+    }
+
+    .mat-button {
+        //padding: 0;
```
DenysVuika, line 11: commented out code should not be committed to repository

lib/core/src/lib/snackbar-content/snackbar-content.component.html
```diff
@@ -0,0 +1,4 @@
+<p class="snackbar-message">{{ data.message }}</p>
+<button mat-button (click)="snackBarRef.dismissWithAction()">
```
DenysVuika, line 2: we should probably have some flag to show/hide this button if necessary?

lib/core/src/lib/snackbar-content/snackbar-content.component.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+        fixture = TestBed.createComponent(SnackbarContentComponent);
+        component = fixture.componentInstance;
+    });
+
+    it('should create', () => {
+        expect(component).toBeTruthy();
+    });
+
+    it('should display message if message in data is set', () => {
+        // given
```
DenysVuika, line 59: no need using comments in the tests as the flow is already obvious, we don't use code comments

DenysVuika (changes requested):

lib/core/src/lib/snackbar-content/snackbar-content.component.scss
AleksanderSklorz: thanks for noticing that @DenysVuika. I were looking for any commented code but did not notice this single line ;)

AleksanderSklorz: done

lib/core/src/lib/snackbar-content/snackbar-content.component.html
AleksanderSklorz: done

lib/core/src/lib/snackbar-content/snackbar-content.component.spec.ts
AleksanderSklorz: done

lib/core/src/lib/snackbar-content/snack-bar-data.ts
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+export interface SnackBarData {
+    actionLabel?: string;
```
MichalKinas, line 19: I think we should make sure that actionLabel is required when actionIcon is passed to mitigate potential future accessibility problems. What do you think?

MichalKinas (changes requested):

lib/core/src/lib/snackbar-content/snack-bar-data.ts
AleksanderSklorz: Sometimes actionLabel is not the same what actionIcon. For example if you remove file and you have notification with Undo action. Then you have Undo label which revert removing file and X icon which just closes notification if there is no need to revert removing. So there is no "relation" between both fields. Also sometimes you may not need label for example when you want just closing button. What do you think? Also we can ask @DenysVuika about opinion because I'm not sure too what approach is better :).

AleksanderSklorz: Could you please explain it a bit more what you exactly mean because I'm not sure if I understand it correctly :) ?
Actual html looks like: 
`<p class="snackbar-message">{{ data.message }}</p>
<div *ngIf="data.showAction" class="action">
    <button mat-button (click)="snackBarRef.dismissWithAction()" *ngIf="data.actionLabel" class="action-label">
        {{data.actionLabel}}
    </button>
    <button mat-button *ngIf="data.actionIcon" (click)="onIconClicked()" class="action-icon">
        <mat-icon>{{ data.actionIcon }}</mat-icon>
    </button>
</div>`

So do you mean to add extra label to second button for icon? Like 
`{{somePassedLabelForIcon}}<mat-icon>{{ data.actionIcon }}</mat-icon>`

Then what if somebody will pass also label for first button? Won't we have too much content in action part? 

Or maybe do you mean to add label visible only for screen readers like aria-label? Please explain deeper what do you mean :)

MichalKinas: Oh sorry I meant the aria-label for screen readers :)

AleksanderSklorz: Ah got it. Sure I think we can add aria-label for button with icon. Will add it in next commit in that PR.

AleksanderSklorz: done

MichalKinas (approved):

docs/core/components/snackbar-content.component.md
```diff
@@ -0,0 +1,46 @@
+---
+Title: Sorting Picker Component
```
DenysVuika, line 2: the title seems to be wrong

lib/core/src/lib/snackbar-content/snackbar-content.component.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+                useValue: {}
+            }]
+        })
+            .compileComponents();
+
+        fixture = TestBed.createComponent(SnackbarContentComponent);
+        component = fixture.componentInstance;
+    });
+
+    it('should create', () => {
```
DenysVuika, line 56: this test is useless as you already test component in other ones

DenysVuika (changes requested):

docs/core/components/snackbar-content.component.md
AleksanderSklorz: corrected

lib/core/src/lib/snackbar-content/snackbar-content.component.spec.ts
AleksanderSklorz: Removed. I thought it may be good to leave it because it says directly on which "level" it has failed. When somebody notices that "should create" is failed then it means that component is not able to be even created but yeah I see your point so removed because somebody can just check stacktrace  :)

DenysVuika (approved):

PR #7952, opened 2022-11-04
[AAE-10767] Make CardView Reusable

DiogoABastos:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

Core `CardViewUpdateService` is coupled with content-services logic.
https://alfresco.atlassian.net/browse/AAE-10767

breaking-changes.md present.

Core `CardViewUpdateService` is no longer coupled with content-services logic. Instead a service was created in content-services to reflect this logic.

breaking-changes.md was removed and logic was moved to upgrade50-60.md.

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/core/src/lib/card-view/services/card-view-update.service.ts
+++ b/lib/core/src/lib/card-view/services/card-view-update.service.ts
@@ -15,19 +15,12 @@
  * limitations under the License.
  */
 
-import { MinimalNode } from '@alfresco/js-api';
 import { Injectable } from '@angular/core';
 import { Subject } from 'rxjs';
+import { BaseCardViewUpdate } from '../interfaces/base-card-view-update.interface';
+import { ClickNotification } from '../interfaces/click-notification.interface';
+import { UpdateNotification } from '../interfaces/update-notification.interface';
 import { CardViewBaseItemModel } from '../models/card-view-baseitem.model';
-
-export interface UpdateNotification {
-    target: CardViewBaseItemModel;
-    changed: any;
-}
-
-export interface ClickNotification {
-    target: any;
-}
 
 export const transformKeyToObject = (key: string, value): any => {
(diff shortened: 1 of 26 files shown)


lib/core/src/lib/card-view/services/card-view-update.service.ts
```diff
@@ -29,6 +28,10 @@ export interface ClickNotification {
     target: any;
 }
 
+export interface MinimalNode {
```
popovicsandras, line 31: This one seems ok for now, just please don't forget about it later. It seems like a very generic interface, which might be used later in many places. It would be premature to move it to a common place, but if/when the time comes, it is important to not duplicate really the same interfaces (if they are indeed distinct).

Also about the naming... I'm not sure the MinimalNode name is necessary anymore. From the perspective of this component, we are saying: "give men an entity which has an id". Interface name for an object containing id can be something more generic (which is easier and more reasonable to reuse), like:
- `HasId`
- `IdAble`
- etc...

Please feel free to change the interfaces' name to something which is meaningful for their new context. Just because we borrow the already existing interfaces from js-api, their name can be adjusted, most of the times we are only interested in (part of) their structure.

An interface like HasId might be better, because having small interfaces like this let us conform easier to Interface Segregation Pattern, since classes can easily implement multiple interfaces.

popovicsandras (approved):

lib/core/src/lib/card-view/services/card-view-update.service.ts
DiogoABastos: I will update the interface's name

DenysVuika (approved):

eromano (approved):

mauriziovitale (approved):

PR #7961, opened 2022-11-08
ACS-3550 Fixed accessibility issues related with move and copy dialogs

AleksanderSklorz:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

There were multiple accessibility issues for move and copy dialogs related with missing labels, headings, not conveyed information to screen reader and missing roles. 

Fixed accessibility issues for move and copy dialogs related with missing labels, headings, not conveyed information to screen reader and missing roles. 

> - [ ] Yes
> - [x] No

**Other information**:


DenysVuika (approved):

lib/content-services/src/lib/content-node-selector/content-node-selector-panel.component.html
```diff
@@ -45,7 +45,7 @@
     <adf-toolbar>
         <adf-toolbar-title>
             <ng-container *ngIf="!showBreadcrumbs()">
-                <span role="heading" aria-level="3" class="adf-search-results-label">{{ 'NODE_SELECTOR.SEARCH_RESULTS' | translate }}</span>
+                <h2 class="adf-search-results-label">{{ 'NODE_SELECTOR.SEARCH_RESULTS' | translate }}</h2>
```
MichalKinas, line 48: Is there any h1 tag above this one? Otherwise this will raise new accessibility issue, h tags should be placed in order

lib/content-services/src/lib/content-node-selector/content-node-selector-panel.component.scss
```diff
@@ -1,7 +1,7 @@
 /* stylelint-disable no-descending-specificity */
 $content-node-selector-thumbnail-width: 35px !default;
 
-.adf-search-results-label {
+h2.adf-search-results-label {
```
MichalKinas, line 4: Do we need to specify h2 selector here? Wouldn't class be enough?

lib/content-services/src/lib/content-node-selector/content-node-selector-panel.component.ts
```diff
@@ -567,6 +572,9 @@ export class ContentNodeSelectorPanelComponent implements OnInit, OnDestroy {
      * Attempts to set the currently loaded node
      */
     onFolderLoaded(nodePaging: NodePaging): void {
+        setTimeout(() => {
```
MichalKinas, line 575: Is `setTimeout` required here?

lib/core/src/lib/datatable/components/datatable-cell/datatable-cell.component.ts
```diff
@@ -41,12 +41,14 @@ import { takeUntil } from 'rxjs/operators';
                 [attr.aria-label]="value$ | async"
                 [title]="tooltip"
                 class="adf-datatable-cell-value"
+                role="button"
```
MichalKinas, line 44: Do we want to set button role to each table cell? It is quite multipurpose component in general. @DenysVuika what do you think?

DenysVuika: hmm, cells are not buttons, correct

AleksanderSklorz: Well, that is interesting point. I was not sure as well so thank you for raising that here. I needed to add button role because of accessibility request from CSV file:
image
https://axeauditor.dequecloud.com/api/v1/file/91e90b78-f1f3-11ec-b66c-0748e7218481

You see that mentions that we need to add that role to these elements. 
Do you have any suggestion how should I fix it instead?

lib/content-services/src/lib/content-node-selector/content-node-selector-panel.component.ts
AleksanderSklorz: Unfortunately yes, I added setTimeout because it was giving me following error: 
"expressionChangedAfterItHasBeenCheckedError: Expression has changed after it was checked. Previous value: 'false'. Current value: 'true'."

I guess it is caused because asynchronous action because onFolderLoaded is called when data loading is finished. I set that timeout without miliseconds because it is enough to fix that error and there is no visible delay thanks that. Other option which I can see is calling change detection manually here so if you prefer that approach I can change it and instead of timeout - call change detection manually. Let me know if you prefer that approach

lib/content-services/src/lib/content-node-selector/content-node-selector-panel.component.scss
AleksanderSklorz: Yes, adding h2 is required. Please have a look at below screenshot: 
image
If you check at CSS panel then you see that when I removed h2 from first rule for adf-search-results-label then all css are greyed out and css for that header are broken. 
The reason is the second rule which is: 
```css
.mat-toolbar, .mat-toolbar h1, .mat-toolbar h2, .mat-toolbar h3, .mat-toolbar h4, .mat-toolbar h5, .mat-toolbar h6 {
    font: 500 20px / 32px Muli, Roboto, "Helvetica Neue", sans-serif;
    letter-spacing: normal;
    margin: 0;
}
```
You see that second rule includes h2 together with class name so to make equal/higher priority for my rule I needed to include h2 tag in my rule.

MichalKinas: Ok sure fine for me in this case :)

lib/content-services/src/lib/content-node-selector/content-node-selector-panel.component.html
AleksanderSklorz: I see content-node-selector-panel is used in three places: 
1.
Screen Shot 2022-11-08 at 15 37 52
2.
Screen Shot 2022-11-08 at 15 38 10

Both have h1 tag in template itself. 

I see it is used also in third place which is AttachFileWidgetDialogComponent but can't find usage of that component to ensure there is any h1 tag. Do you have idea where in application it is used?

lib/core/src/lib/datatable/components/datatable-cell/datatable-cell.component.ts
MichalKinas: Honestly nothing good comes to my mind right now. @DenysVuika any ideas for that?

lib/content-services/src/lib/content-node-selector/content-node-selector-panel.component.html
AleksanderSklorz: ok I will try to check that one place too to ensure it is working there as well. Rest places have h1 tag like you can see on screenshots :) Will let you know about third place when I check. If you know any other edge cases where we use that component then let me know but what I can see except these 3 places there is no more usage.

lib/core/src/lib/datatable/components/datatable-cell/datatable-cell.component.ts
DenysVuika: Sometimes a11y tools suggest completely stupid solutions, so we should be still evaluating scenarios with the common sense. Table cells can never be buttons

AleksanderSklorz: I think the reason why it could think it is button by mistake is that you can focus cell using tab and "click" on that cell using enter so then you enter into folder: 
image

But sure we can skip that point. I will remove these roles in next commit in that PR

AleksanderSklorz: Done

lib/content-services/src/lib/content-node-selector/content-node-selector-panel.component.ts
TomikaArome: Adding an immediate `setTimeout` is my usual fix for `ExpressionChangedAfterItHasBeenCheckedError` too. You _could_ manually trigger a change detection (I believe it's with `ChangeDetectorRef`), but I generally prefer this solution as it doesn't require an additional injection. `setTimeout` with 0 timeout should just cause the callback to be ran on the next tick, which appears as immediate.

lib/content-services/src/lib/content-node-selector/content-node-selector-panel.component.scss
```diff
@@ -194,3 +194,16 @@ $content-node-selector-thumbnail-width: 35px !default;
...
+.adf-content-node-selector-content-list {
+    position: relative;
+
+    .adf-content-node-selector-content-list-empty {
+        & + adf-infinite-pagination {
+            position: absolute;
+            bottom: 0;
+            width: 100%;
+        }
+    }
```
TomikaArome, line 208: Minor, but this could be contracted a little:
```suggestion
    &-empty + adf-infinite-pagination {
        position: absolute;
        bottom: 0;
        width: 100%;
    }
```

TomikaArome (approved):

lib/content-services/src/lib/content-node-selector/content-node-selector-panel.component.ts
MichalKinas: Sure guys thanks for explanation :D

MichalKinas (approved):

PR #7976, opened 2022-11-15
[MNT-23143] Allow discoveryService withCredentials

MichalKinas:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://alfresco.atlassian.net/browse/MNT-23143

Allowed discovery service withCredentials. 

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/core/src/lib/services/discovery-api.service.spec.ts
+++ b/lib/core/src/lib/services/discovery-api.service.spec.ts
@@ -25,7 +25,7 @@ import { ApiClientsService } from '../../../api/api-clients.service';
 import { RepositoryInfo } from '@alfresco/js-api';
 import { Observable } from 'rxjs';
 
-fdescribe('DiscoveryApiService', () => {
+describe('DiscoveryApiService', () => {
 
     let service: DiscoveryApiService;
     let authService: AuthenticationService;
--- a/lib/core/src/lib/services/discovery-api.service.spec.ts
+++ b/lib/core/src/lib/services/discovery-api.service.spec.ts
@@ -1,0 +1,76 @@
+/*!
+ * @license
+ * Copyright 2019 Alfresco Software, Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { TestBed } from '@angular/core/testing';
+import { CoreTestingModule } from '../testing/core.testing.module';
+import { TranslateModule } from '@ngx-translate/core';
+import { AuthenticationService } from './authentication.service';
+import { DiscoveryApiService } from './discovery-api.service';
+import { AlfrescoApiService } from './alfresco-api.service';
+import { ApiClientsService } from '../../../api/api-clients.service';
+import { RepositoryInfo } from '@alfresco/js-api';
+import { Observable } from 'rxjs';
+
+fdescribe('DiscoveryApiService', () => {
+
+    let service: DiscoveryApiService;
+    let authService: AuthenticationService;
+    let alfrescoApiService: AlfrescoApiService;
+    let apiClientsService: ApiClientsService;
+
+    beforeEach(() => {
+        TestBed.configureTestingModule({
+            imports: [
+                TranslateModule.forRoot(),
+                CoreTestingModule
+            ]
+        });
+
+        authService = TestBed.inject(AuthenticationService);
+        alfrescoApiService = TestBed.inject(AlfrescoApiService);
+        apiClientsService = TestBed.inject(ApiClientsService);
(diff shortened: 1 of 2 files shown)


lib/core/src/lib/services/discovery-api.service.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+import { CoreTestingModule } from '../testing/core.testing.module';
+import { TranslateModule } from '@ngx-translate/core';
+import { AuthenticationService } from './authentication.service';
+import { DiscoveryApiService } from './discovery-api.service';
+import { AlfrescoApiService } from './alfresco-api.service';
+import { ApiClientsService } from '../../../api/api-clients.service';
+import { RepositoryInfo } from '@alfresco/js-api';
+import { Observable } from 'rxjs';
+
+fdescribe('DiscoveryApiService', () => {
```
AleksanderSklorz, line 28: I think should be describe?

MichalKinas: Exactly, forgot to remove it, fixed now

lib/core/src/lib/services/discovery-api.service.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+    it('should filter out events when ECM provider and all provider are both not enabled',  () => {
+        spyOn(authService, 'isECMProvider').and.returnValue(false);
+        spyOn(authService, 'isALLProvider').and.returnValue(false);
+        spyOn(authService, 'isKerberosEnabled').and.returnValue(true);
+        service = new DiscoveryApiService(alfrescoApiService, authService, apiClientsService);
+        authService.onLogin.next(new RepositoryInfo());
+        expect(service.getEcmProductInfo).not.toHaveBeenCalled()
+    });
+
+    it('should pass events when Kerberos and ECM or all provider are enabled',  () => {
```
AleksanderSklorz, line 68: I'm wondering if that part in title is correct: "and ECM or all provider are enabled". In test ECM returns false which means it is not enabled so that test does not check case for enabled ECM so we can't be sure that it works for that case too so maybe better to rename title?

MichalKinas: Basically the condition for this case is `this.authenticationService.isECMProvider() || this.authenticationService.isALLProvider()` so one of them enabled is enough, second test case in this example would look exactly the same just with other provider enabled, I can add it if you think it should be covered as well :)

lib/core/src/lib/services/discovery-api.service.ts
```diff
@@ -42,7 +42,10 @@ export class DiscoveryApiService {
     ) {
         this.authenticationService.onLogin
             .pipe(
-                filter(() => this.apiService.getInstance()?.isEcmLoggedIn()),
+                filter(() => this.apiService.getInstance()?.isEcmLoggedIn() ||
+                ((this.authenticationService.isECMProvider() ||
```
eromano, line 46: This condition seems to be redundant

eromano (changes requested):

lib/core/src/lib/services/discovery-api.service.ts
MichalKinas: Why do you think so? In case of `isEcmLoggedIn` returns `false` we check for ecm or all provider but in this case Kerberos must be enabled as well.

eromano: have you check isEcmLoggedIn inside? does some of the thing you are adding here.

probably will make more sense move this methods in the authentication 

https://github.com/Alfresco/alfresco-ng2-components/blob/147fb5672448dbdb51296c005fcde8b292749411/lib/core/src/lib/userinfo/components/user-info.component.ts#L139 

and use it?

or would be better implement this logic before to do the next in the onLogin ?

https://github.com/Alfresco/alfresco-ng2-components/blob/4410557d8dbf6573d430c3cda7d35330d5975c00/lib/core/src/lib/services/authentication.service.ts#L73

Otherwise all the call that use  needs onLogin will need to add those checks

eromano (changes requested):

lib/core/src/lib/services/discovery-api.service.spec.ts
AleksanderSklorz: ok, fine, you can leave it :)

lib/core/src/lib/services/discovery-api.service.ts
MichalKinas: @eromano I did, but this is different case I think, when using Kerberos, `isEcmLoggedIn()` always returns false and those checks are added for discoveryService to be able to retrieve the ECM product info with Kerberos enabled. Also I don't think that we need to modify each subscription to `onLogin` but I don't know this area that well. If you're unsure about this solution or if you think that it should be added in some other areas please talk to Nicolas Barithel. He is the author of the change but I recreated his PR so we can unit test the changes, the original PR -> https://github.com/Alfresco/alfresco-ng2-components/pull/7851. In case of any changes that has to be made please let me know.

MichalKinas: @eromano could you please check out my comment?

eromano:
What we need to modify is actually the alfresco JS-API https://github.com/Alfresco/alfresco-js-api/blob/2103ded662921b411288e72bf064fae732487257/src/alfrescoApi.ts#L475
isEcmLoggedIn to return true when  WithCredentials is true .

So this PR I guess is not needed what is needed to change is in the JS-API

MichalKinas:
Changes moved to the JS-API, closing this one.

PR #8034, opened 2022-12-06
[ACS-3757] returning focus to element from which they were opened

AleksanderSklorz:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

Modals which were opened from personal files/libraries list or toolbar or more menu, after closing, were not returning focus to element from which they were opened.

After closing modals opened from personal files/libraries or toolbar or more menu, focus is returned to correct element.

> - [ ] Yes
> - [x] No

**Other information**:

ACA PR: https://github.com/Alfresco/alfresco-content-app/pull/2837 
APPS PR: https://github.com/Alfresco/alfresco-apps/pull/4630

### DIFF
--- a/lib/content-services/src/lib/aspect-list/services/dialog-aspect-list.service.spec.ts
+++ b/lib/content-services/src/lib/aspect-list/services/dialog-aspect-list.service.spec.ts
@@ -1,0 +1,88 @@
+/*!
+ * @license
+ * Copyright 2019 Alfresco Software, Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { setupTestBed } from '@alfresco/adf-core';
+import { TranslateModule } from '@ngx-translate/core';
+import { ContentTestingModule } from '../../testing/content.testing.module';
+import { DialogAspectListService } from '@alfresco/adf-content-services';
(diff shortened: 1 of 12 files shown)


MichalKinas:
Can we add some unit tests to make sure everything will work as expected. when new param is provided?

AleksanderSklorz:
@MichalKinas I added comments

lib/content-services/src/lib/aspect-list/services/dialog-aspect-list.service.spec.ts
```diff
@@ -0,0 +1,88 @@
...
+            const afterClosed$ = new Subject<void>();
+            spyOn(dialog, 'open').and.returnValue({
+                afterClosed: () => afterClosed$.asObservable()
+            } as MatDialogRef<any>);
+            const elementToFocus = document.createElement(elementToFocusSelector);
+            spyOn(elementToFocus, 'focus');
+            spyOn(document, 'querySelector').withArgs(elementToFocusSelector).and.returnValue(elementToFocus);
+            dialogAspectListService.openAspectListDialog('some-node-id', elementToFocusSelector);
+            afterClosed$.next();
+            expect(elementToFocus.focus).toHaveBeenCalledWith();
```
nikita-web-ua, line 55: why not `toHaveBeenCalled()`?

nikita-web-ua (approved):
LGTM

lib/content-services/src/lib/aspect-list/services/dialog-aspect-list.service.spec.ts
AleksanderSklorz: The reason why I decided to use toHaveBeenCalledWith is that focus function takes some optional parameters. Difference between these two functions is that toHaveBeenCalled will check only if that function has been called but toHaveBeenCalledWith checks also parameters. 

If anyone will pass value to focus function then that value will affect on behavior of focus function right? Focus function will behave depending on arguments which it takes. 

But I get your point that it may be too strict so I will change it as you suggested.

AleksanderSklorz: done :)

DenysVuika (approved):

PR #8036, opened 2022-12-06
[AAE-10777] Make the core services truly reusable Part 1

alep85:
…e calling api client service get method

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

> - [ ] Yes
> - [ ] No

**Other information**:


popovicsandras (changes requested):

lib/content-services/src/lib/services/deleted-nodes-api.service.ts
```diff
@@ -17,10 +17,8 @@
```
kathrine0, line 17: General question - do we want to aggregate the services in one `services` folder as before? Maybe it would be a better idea to create area-dedicated directories and place the services (and other components, directives etc) where they belong logically rather than functionally (so in this case, in a directory called something like `deleted-nodes`)

@popovicsandras what are your thoughts about it?

lib/core/api/clients/activiti/activiti-client.module.ts
```diff
@@ -15,7 +15,7 @@
  * limitations under the License.
  */
 
-import { AboutApi, SystemPropertiesApi } from '@alfresco/js-api';
+import { AboutApi, SystemPropertiesApi, UserProfileApi } from '@alfresco/js-api';
```
kathrine0, line 18: If I understood correctly, the ultimate goal is to get rid of the js-api dependency from the core? Is this going to be done in the next step for this module here (or maybe whole module moved somewhere else?)

lib/process-services/src/lib/services/apps-process.service.ts
```diff
@@ -18,8 +18,7 @@
 import { Injectable } from '@angular/core';
```
kathrine0, line 18: Similar comment as above- I'd avoid stacking up all of the services in one `services` folder but rather organize them by domain/area on what they actually do. If possible of course :) If not, let me know why :)

DiogoABastos: In this specific case, this service was moved to process services but kept in a general services folder because it is used across multiple modules across the package.

lib/core/api/clients/activiti/activiti-client.module.ts
alep85: Yeah, for now is ok to get it from the `js-api`, we'll get rid of it in the next step

DenysVuika:
is  this PR still actual?

DiogoABastos:
I believe the changes from this pr are being handled by this one, https://github.com/Alfresco/alfresco-ng2-components/pull/8242

alep85:
Closed because a new PR https://github.com/Alfresco/alfresco-ng2-components/pull/8242 has been created to clean up the services

PR #8041, opened 2022-12-09
AAE-11739 removed unnecessary api call when cancelling not started pr…

tomaszhanaj:
…ocess

> - [ ] Bugfix
> - [x] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://alfresco.atlassian.net/browse/AAE-11739

When user press cancel no api call is made, only redirect.

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/process-services-cloud/src/lib/process/start-process/components/start-process-cloud.component.ts
+++ b/lib/process-services-cloud/src/lib/process/start-process/components/start-process-cloud.component.ts
@@ -117,8 +117,8 @@ export class StartProcessCloudComponent implements OnChanges, OnInit, OnDestroy 
     loading$ = new BehaviorSubject<boolean>(!this.processDefinitionLoaded);
 
     constructor(private startProcessCloudService: StartProcessCloudService,
-                private formBuilder: UntypedFormBuilder,
-                private processNameCloudPipe: ProcessNameCloudPipe) {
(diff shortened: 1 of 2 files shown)


lib/process-services-cloud/src/lib/process/start-process/components/start-process-cloud.component.ts
```diff
@@ -334,11 +334,7 @@ export class StartProcessCloudComponent implements OnChanges, OnInit, OnDestroy
             );
     }
 
-    async cancelStartProcess() {
-        if (this.currentCreatedProcess) {
-            await this.startProcessCloudService.deleteProcess(this.appName, this.currentCreatedProcess.id);
-        }
-
+    cancelStartProcess() {
```
mauriziovitale, line 337: don't we need unit test? 

I think also the name of the methods does't reflect the current behaviour. As far as I remember it was because of the separation we did ( and reverted ) between started process but not initialized? cc @VitoAlbano

VitoAlbano: This is a two parts task, i expected to see some unit test removed though (or added in case we don't have. For the method name i'm fine with the actual one, makes enough sense

VitoAlbano (approved):

PR #8045, opened 2022-12-13
[AAE-10769] remove alfresco deps from datatable 

eromano:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

remove alfresco deps from datatable 

> - [ ] Yes
> - [ ] No

**Other information**:


lib/core/src/lib/datatable/components/datatable-cell/datatable-cell.component.ts
```diff
@@ -82,21 +80,10 @@ export class DataTableCellComponent implements OnInit, OnDestroy {
 
     protected onDestroy$ = new Subject<boolean>();
 
-    constructor(protected alfrescoApiService: AlfrescoApiService) {}
+    constructor() {}
 
     ngOnInit() {
         this.updateValue();
-        this.alfrescoApiService.nodeUpdated
```
DenysVuika, line 89: what will be the alternative for this? this feature was to update the document list when node is renamed via the metadata card

eromano: i think we need to trigger events from that component for ask a refresh or we can add this nodeupdate there if is a content level

eromano: my solution actually refresh the whole datatable we may need a generic approach to be less invasice I am open to suggestions

eromano: @DenysVuika check now

popovicsandras (approved):

lib/core/src/lib/datatable/services/datatable.service.ts
```diff
@@ -0,0 +1,31 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { Injectable } from '@angular/core';
+import { Subject } from 'rxjs';
+import { DataRowUpdateModel } from '../data/data-row-update.model';
+
+@Injectable({
+    providedIn: 'root'
```
popovicsandras, line 23: Actually, I've just realised, this auto provision should not be here, since this way there would be a default singleton instance, which:
- makes the Optional() flag meaningless
- makes it very difficult to troubleshoot in case of multiple datatables

popovicsandras (changes requested):

popovicsandras (approved):

PR #8052, opened 2022-12-14
[AAE-11740] replaced api for starting process

tomaszhanaj:

> - [ ] Bugfix
> - [x] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://alfresco.atlassian.net/browse/AAE-11740

We are starting process when user clicks button start (without generating instance of process in ngOnInit). 

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/process-services-cloud/src/lib/process/start-process/components/start-process-cloud.component.spec.ts
+++ b/lib/process-services-cloud/src/lib/process/start-process/components/start-process-cloud.component.spec.ts
@@ -664,22 +664,40 @@ describe('StartProcessCloudComponent', () => {
             component.ngOnChanges({});
         });
 
-        it('should call service to start process if required fields provided', () => {
-            component.processForm.controls['processInstanceName'].setValue('My Process 1');
-            component.processPayloadCloud.processDefinitionKey = 'my test definition key';
-            component.startProcess();
-            expect(startProcessSpy).toHaveBeenCalled();
+        it('should see start button', async () => {
+            component.ngOnChanges({ appName: firstChange });
+            fixture.detectChanges();
+            await fixture.whenStable();
+
+            const startButton = fixture.debugElement.query(By.css('#button-start'));
+            expect(startButton).toBeDefined();
+            expect(startButton).not.toBeNull();
         });
 
-        it('should call service to start process with the correct parameters', () => {
+        it('should call service with the correct parameters when button is clicked', async () => {
+            component.ngOnChanges({ appName: firstChange });
+            component.processForm.controls['processInstanceName'].setValue('My Process 1');
             component.appName = 'test app name';
             const payload: ProcessPayloadCloud = new ProcessPayloadCloud({
                 name: component.processInstanceName.value,
                 ProcessDefinitionKey: component.processPayloadCloud.processDefinitionKey
             });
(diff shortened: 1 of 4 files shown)


lib/process-services-cloud/src/lib/process/start-process/services/start-process-cloud.service.ts
```diff
@@ -96,7 +96,9 @@ export class StartProcessCloudService extends BaseCloudService {
         const url = `${this.getBasePath(appName)}/rb/v1/process-instances`;
         payload.payloadType = 'StartProcessPayload';
 
-        return this.post(url, payload);
+        return this.post(url, payload).pipe(
+            map((result: any) => result.entry)
```
VitoAlbano, line 100: This might be potentially problematic on apps, as maybe someone is using it and it doesn't get the entry or already maps outside for entry.

VitoAlbano (changes requested):

lib/process-services-cloud/src/lib/process/start-process/services/start-process-cloud.service.ts
VitoAlbano: As the fact that this would change in a more invasive way i would say that's ok

VitoAlbano (approved):

lib/process-services-cloud/src/lib/process/start-process/components/start-process-cloud.component.spec.ts
```diff
@@ -694,43 +665,32 @@ describe('StartProcessCloudComponent', () => {
         });
 
         it('should call service to start process if required fields provided', () => {
-            component.currentCreatedProcess = fakeProcessInstance;
+            component.processForm.controls['processInstanceName'].setValue('My Process 1');
```
mauriziovitale, line 668: I think we should have just a test from the UI where the user provides the Process instance name, the process definition key and when clicking on submit we can spy on the startProcessSpy but making sure the payload is the expected not just toHaveBeenCalled

lib/process-services-cloud/src/lib/process/start-process/components/start-process-cloud.component.spec.ts
```diff
@@ -939,6 +896,13 @@ describe('StartProcessCloudComponent', () => {
 
             expect(card).toBeTruthy();
         });
+
+        it('should emit value when process started', () => {
+            component.success.pipe(first()).subscribe((data: ProcessInstanceCloud) => {
+                expect(data).not.toBeNull();
```
mauriziovitale, line 902: please let's be more specific pass a define data and expect a defined data. 

Again, in my opinion we could merge some different unit into 1

Fill out the start form / press submit / expect the component emits the correct payload

mauriziovitale (commented):
check comments

lib/process-services-cloud/src/lib/process/start-process/components/start-process-cloud.component.spec.ts
tomaszhanaj: Done, test in: "should call service with the correct parameters when button is clicked"

DenysVuika (approved):

PR #8021, opened 2022-11-30
[AAE-11480] - Fix required error is displayed for a non required drop…

arditdomi:
…down

**Please check if the PR fulfills these requirements**

> - [X] The commit message follows our [guidelines](https://github.com/Alfresco/alfresco-ng2-components/wiki/Commit-format)
> - [X] Tests for the changes have been added (for bug fixes / features)
> - [ ] Docs have been added / updated (for bug fixes / features)

**What kind of change does this PR introduce?** (check one with "x")

> - [X] Bugfix
> - [ ] Feature
> - [ ] Code style update (formatting, local variables)
> - [ ] Refactoring (no functional changes, no api changes)
> - [ ] Build related changes
> - [ ] Documentation
> - [ ] Other... Please describe:

**What is the current behaviour?** (You can also link to an open issue here)
https://alfresco.atlassian.net/browse/AAE-11480

**What is the new behaviour?**

**Does this PR introduce a breaking change?** (check one with "x")

> - [ ] Yes
> - [X] No

If this PR contains a breaking change, please describe the impact and migration path for existing applications: ...

**Other information**:

PR #8107, opened 2023-01-06
MNT-23288 - Removed required field when starting a process

antoniojfelix:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://alfresco.atlassian.net/browse/MNT-23288

> - [ ] Yes
> - [x] No

**Other information**:

This change was based on the PR https://github.com/Alfresco/alfresco-ng2-components/pull/8021/ since it was the same fix.

### DIFF
--- a/lib/process-services/src/lib/form/widgets/dropdown/dropdown.widget.spec.ts
+++ b/lib/process-services/src/lib/form/widgets/dropdown/dropdown.widget.spec.ts
@@ -126,6 +126,82 @@ describe('DropdownWidgetComponent', () => {
         expect(widget.field.options[1]).toBe(restFieldValue);
     });
 
+
+
+    it('should not display required error for a non required dropdown when selecting the none option', async () => {
+        widget.field.options = [
+            { id: 'empty', name: 'Choose empty' },
+            ...fakeOptionList
+        ];
+
+        widget.ngOnInit();
+        fixture.detectChanges();
+        await openSelect();
+
+        const defaultOption: any = fixture.debugElement.query(By.css('[id="empty"]'));
+        widget.touched = true;
+        defaultOption.triggerEventHandler('click', null);
+        fixture.detectChanges();
+
+        const requiredErrorElement = fixture.debugElement.query(By.css('.adf-dropdown-required-message .adf-error-text'));
+        expect(requiredErrorElement).toBeFalsy();
+    });
+
+    it('should not display required error when selecting a valid option for a required dropdown', async () => {
+        widget.field.required = true;
+        widget.field.options = [
+            { id: 'empty', name: 'Choose empty' },
+            ...fakeOptionList
+        ];
+
+        widget.ngOnInit();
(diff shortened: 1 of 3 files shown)


lib/process-services/src/lib/form/widgets/dropdown/dropdown.widget.ts
```diff
@@ -119,7 +124,15 @@ export class DropdownWidgetComponent extends WidgetComponent implements OnInit {
...
+    private isNoneValueSelected(value: string): boolean {
+        return value === undefined;
+    }
+
     showRequiredMessage(): boolean {
-        return (this.isInvalidFieldRequired() || this.field.value === 'empty') && this.isTouched();
+        return (this.isInvalidFieldRequired() || (this.isNoneValueSelected(this.field.value) && this.isRequired())) && this.isTouched() ;
+    }
+
+    getDefaultOption(options: FormFieldOption[]): FormFieldOption {
```
arditdomi, line 135: are you using this new method anywhere?

lib/process-services/src/lib/form/widgets/dropdown/dropdown.widget.ts
```diff
@@ -119,7 +124,15 @@ export class DropdownWidgetComponent extends WidgetComponent implements OnInit {
         return this.field.type === 'readonly';
     }
 
+    private isNoneValueSelected(value: string): boolean {
+        return value === undefined;
```
arditdomi, line 128: are you sure this will work to show the required error when the none option is selected? In my fix for APA I did that because we changed the way the dropdown widgets work to match the Angular Material approach https://material.angular.io/components/select/overview#select-hint-error . 

With that approach when the none option is selected the value equals undefined. As far as I can understand from seeing this code I don't see a similar implementation for that, so your none value still contains the `id: 'empty'`. Considering that I don't think this method is what you need.

Have you tested it and it works?

lib/process-services/src/lib/form/widgets/dropdown/dropdown.widget.spec.ts
```diff
@@ -126,6 +126,80 @@ describe('DropdownWidgetComponent', () => {
         expect(widget.field.options[1]).toBe(restFieldValue);
     });
 
+    it('should not display required error for a non required dropdown when selecting the none option', async () => {
```
mauriziovitale, line 129: do we have a test for the opposite case? make sure we have an error when is required?

antoniojfelix:
Hello, I've opened this PR because I believed that this issue was partially fixed with the fix made on the https://alfresco.atlassian.net/browse/AAE-11480, plus another fix that was made on the APS side. After further testing and investigation, I realized that the APS fix resolved the problem altogether without any change on the front end side. That's why that in my tests it worked, not because of this fix but because of the APS fix.
Therefore I'll close this PR. Thanks for your input.

PR #8128, opened 2023-01-11
[AAE-11992] Fix processInstanceId filter empty

alep85:
…ut to check for the processInstanceId changes to set the filter

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://alfresco.atlassian.net/browse/AAE-11992

https://user-images.githubusercontent.com/23027333/212028151-9abc5590-f50f-4b06-a057-5efda617360d.mp4

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/process-services-cloud/src/lib/task/task-filters/components/edit-task-filters/base-edit-task-filter-cloud.component.ts
+++ b/lib/process-services-cloud/src/lib/task/task-filters/components/edit-task-filters/base-edit-task-filter-cloud.component.ts
@@ -117,7 +117,9 @@ export abstract class BaseEditTaskFilterCloudComponent<T> implements OnInit, OnC
         label: 'ADF_CLOUD_TASK_FILTERS.STATUS.ALL'
     };
 
+    @Input()
     taskFilter: T;
+
     changedTaskFilter: T;
 
     /** Emitted when a task filter property changes. */
@@ -145,8 +147,12 @@ export abstract class BaseEditTaskFilterCloudComponent<T> implements OnInit, OnC
     }
 
     ngOnChanges(changes: SimpleChanges) {
-        const id = changes['id'];
-        if (id && id.currentValue !== id.previousValue) {
+        const { id, taskFilter } = changes;
+        if(taskFilter && taskFilter.currentValue?.processInstanceId !== taskFilter.previousValue?.processInstanceId){
+            this.taskFilterProperties = this.createAndFilterProperties();
+            this.taskFilterActions = this.createAndFilterActions();
+            this.buildForm(this.taskFilterProperties);
+        } else if (id && id.currentValue !== id.previousValue) {
             this.retrieveTaskFilterAndBuildForm();
         }
     }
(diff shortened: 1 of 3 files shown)


eromano (approved):

VitoAlbano (approved):

lib/process-services-cloud/src/lib/task/task-filters/components/edit-task-filters/base-edit-task-filter-cloud.component.ts
```diff
@@ -117,7 +117,9 @@ export abstract class BaseEditTaskFilterCloudComponent<T> implements OnInit, OnC
         label: 'ADF_CLOUD_TASK_FILTERS.STATUS.ALL'
     };
 
+    @Input()
```
arditdomi, line 120: Are u sure about this? The idea before was that u pass the id and the filter gets fetched by the component calling the API. Instead now you are changing the strategy completely by passing the whole filter as input

lib/process-services-cloud/src/lib/task/task-filters/components/edit-task-filters/base-edit-task-filter-cloud.component.ts
```diff
@@ -145,8 +147,12 @@ export abstract class BaseEditTaskFilterCloudComponent<T> implements OnInit, OnC
...
 
     ngOnChanges(changes: SimpleChanges) {
-        const id = changes['id'];
-        if (id && id.currentValue !== id.previousValue) {
+        const { id, taskFilter } = changes;
+        if(taskFilter && taskFilter.currentValue?.processInstanceId !== taskFilter.previousValue?.processInstanceId){
+            this.taskFilterProperties = this.createAndFilterProperties();
+            this.taskFilterActions = this.createAndFilterActions();
+            this.buildForm(this.taskFilterProperties);
+        } else if (id && id.currentValue !== id.previousValue) {
```
arditdomi, line 155: Are you sure this is not breaking anything? Before the flow was

if (id && id.currentValue !== id.previousValue) {
            this.retrieveTaskFilterAndBuildForm();
}

and now the same condition is within an else if. It changes the flow and for that condition to be there I suppose there was a reason.

Have u tested a scenario were process instance id changes but also the filter id? Is there such scenario where the filter id would change but the else if condition (line 55) not reached because it is inside an else if instead of a plain if that was before?

lib/process-services-cloud/src/lib/task/task-filters/components/edit-task-filters/base-edit-task-filter-cloud.component.ts
alep85: The problem passing only the `id` is that the preferences API always return the `processInstanceId` `null` and the filter is always empty, passing the taskFilter as input we can detect `processInstanceId` changes and update the filter value

arditdomi: and why don't we pass the processInstanceId as input same way the appName is passed or the sortingProperties or the filterProperties but we want to pass the whole filter object?

PR #8143, opened 2023-01-13
[ACS-4118] create a tag from tags list

AleksanderSklorz:

> - [ ] Bugfix
> - [x] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://alfresco.atlassian.net/browse/ACS-4118

Added functions to create tags without node and to search tags which name contains specified name.

> - [ ] Yes
> - [x] No

**Other information**:
JS-API PR https://github.com/Alfresco/alfresco-js-api/pull/1492
APPS PR https://github.com/Alfresco/alfresco-apps/pull/4759

### DIFF
--- a/lib/content-services/src/lib/tag/services/tag.service.spec.ts
+++ b/lib/content-services/src/lib/tag/services/tag.service.spec.ts
@@ -25,7 +25,8 @@ import {
     RequestQuery,
     RequestSortDefinitionInner,
     ResultSetPaging,
-    SearchApi, SearchRequest,
+    SearchApi,
+    SearchRequest,
     TagBody,
     TagEntry
 } from '@alfresco/js-api';
--- a/lib/content-services/src/lib/tag/services/tag.service.spec.ts
+++ b/lib/content-services/src/lib/tag/services/tag.service.spec.ts
@@ -21,7 +21,14 @@ import { fakeAsync, TestBed, tick } from '@angular/core/testing';
 import { ContentTestingModule } from '../../testing/content.testing.module';
 import { TranslateModule } from '@ngx-translate/core';
 import { throwError } from 'rxjs';
-import { TagBody, TagEntry } from '@alfresco/js-api';
+import {
+    RequestQuery,
+    RequestSortDefinitionInner,
+    ResultSetPaging,
+    SearchApi, SearchRequest,
+    TagBody,
+    TagEntry
+} from '@alfresco/js-api';
 
 describe('TagService', () => {
 
@@ -74,11 +81,11 @@ describe('TagService', () => {
         describe('createTags', () => {
             it('should call createTags on tagsApi', () => {
                 spyOn(service.tagsApi, 'createTags').and.returnValue(Promise.resolve([]));
-                const tag1 = new TagBody();
+                const tag1: TagBody = new TagBody();
                 tag1.tag = 'Some tag 1';
-                const tag2 = new TagBody();
+                const tag2: TagBody = new TagBody();
                 tag2.tag = 'Some tag 2';
-                const tags = [tag1, tag2];
+                const tags: TagBody[] = [tag1, tag2];
                 service.createTags(tags);
                 expect(service.tagsApi.createTags).toHaveBeenCalledWith(tags);
             });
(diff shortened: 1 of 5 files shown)


lib/content-services/src/lib/tag/services/tag.service.spec.ts
```diff
@@ -15,13 +15,20 @@
...
+import { fakeAsync, TestBed, tick } from '@angular/core/testing';
 import { ContentTestingModule } from '../../testing/content.testing.module';
 import { TranslateModule } from '@ngx-translate/core';
 import { throwError } from 'rxjs';
-import { TagEntry } from '@alfresco/js-api';
+import {
+    RequestQuery,
+    RequestSortDefinitionInner,
+    ResultSetPaging,
+    SearchApi, SearchRequest,
```
jatin2008, line 28: can we add SearchRequest import in next line.

AleksanderSklorz: done

lib/content-services/src/lib/tag/services/tag.service.ts
```diff
@@ -105,6 +116,46 @@ export class TagService {
...
+        sortingByName.ascending = true;
+        sortingByName.type = RequestSortDefinitionInner.TypeEnum.FIELD;
+        return from(this.searchApi.search({
+            query: {
+                language: RequestQuery.LanguageEnum.Afts,
+                query: `PATH:"/cm:categoryRoot/cm:taggable/*" AND cm:name:"${name}*"`
+            },
+            paging: {
+                skipCount,
+                maxItems: 100
```
MichalKinas, line 153: Do we always want 100 entries to be fetched? Or should we maybe use `UserPreferencesService` as we do for pagination? @AleksanderSklorz What do you think?

MichalKinas (approved):

lib/content-services/src/lib/tag/services/tag.service.ts
AleksanderSklorz: Let me test it how it behaves

AleksanderSklorz: Works fine. Done. Thank you for suggestion :)

MichalKinas: Oh by the way we can also consider providing `maxItems` as optional method parameter, so then whoever wants to use it can decide if he wants to pass some value there, or leave it empty and then 100 is the default that API provides when `maxItems` is empty

AleksanderSklorz: Sure I can add maxItems parameter to function. If anyone sets maxItems parameters then this value will be taken. In other case value from UserPreferencesService will be used.

AleksanderSklorz: @MichalKinas done

jatin2008 (approved):

lib/content-services/src/lib/tag/services/tag.service.spec.ts
```diff
@@ -70,5 +78,95 @@ describe('TagService', () => {
...
+        describe('searchTags', () => {
+            let result: ResultSetPaging;
+
+            beforeEach(() => {
+                result = new ResultSetPaging();
+            });
+
+            it('should call search on searchApi with correct parameters', () => {
+                const searchSpy: jasmine.Spy<(queryBody: SearchRequest) => Promise<ResultSetPaging>> =
+                    spyOn(SearchApi.prototype, 'search').and.returnValue(Promise.resolve(result));
```
DenysVuika, line 128: we do not spy on prototypes, it's Angular, and there are means to inject and spy properly

AleksanderSklorz: @DenysVuika I spied prototype because SearchAPI is not injectable so it is created in that way in service: 
private searchApi: SearchApi = new SearchApi(this.apiService.getInstance());

It is created directly by constructor. I can't spy instance of searchApi field because it is private. Can I leave it because of that as it is?

PR #8165, opened 2023-01-18
[ACS-4051] Copy to clipboard button is now accessible through the keyboard enter earlier which was only accessible through mouse click

jatin2008:
… earlier which was only accessible through mouse click

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

Copy to clipboard button was not accessible through the keyboard enter button

Copy to clipboard button is now accessible through the keyboard enter earlier which was only accessible through mouse click

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/core/src/lib/clipboard/clipboard.directive.spec.ts
+++ b/lib/core/src/lib/clipboard/clipboard.directive.spec.ts
@@ -64,6 +64,15 @@ describe('ClipboardDirective', () => {
 
         expect(clipboardService.copyToClipboard).toHaveBeenCalled();
     });
+
+    it('should notify copy target value on mouse enter event', () => {
+        spyOn(clipboardService, 'copyToClipboard');
+        fixture.nativeElement.querySelector('input').value = 'some value';
+        fixture.nativeElement.querySelector('button').dispatchEvent(new KeyboardEvent('keydown', {code: 'Enter', key: 'Enter'}));
+
+        expect(clipboardService.copyToClipboard).toHaveBeenCalled();
+    });
+
 });
 
 describe('CopyClipboardDirective', () => {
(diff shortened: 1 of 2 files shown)


lib/core/src/lib/clipboard/clipboard.directive.ts
```diff
@@ -47,6 +47,13 @@ export class ClipboardDirective {
         this.copyToClipboard();
     }
 
+    @HostListener('keydown.enter', ['$event'])
+    handleKeyDown(event: KeyboardEvent){
```
AleksanderSklorz, line 51: This function looks like that function above so maybe worth to attach two HostListeners to single function instead of repeating whole logic twice

lib/core/src/lib/clipboard/clipboard.directive.spec.ts
```diff
@@ -128,6 +137,16 @@ describe('CopyClipboardDirective', () => {
         expect(navigator.clipboard.writeText).toHaveBeenCalledWith('text to copy');
     }));
 
+    it('should copy the content of element on mouse enter', fakeAsync(() => {
```
AleksanderSklorz, line 140: I think title is incorrect. It says: "mouse enter" but your event is different.

lib/core/src/lib/clipboard/clipboard.directive.spec.ts
```diff
@@ -64,6 +64,15 @@ describe('ClipboardDirective', () => {
 
         expect(clipboardService.copyToClipboard).toHaveBeenCalled();
     });
+
+    it('should notify copy target value on mouse enter event', () => {
```
AleksanderSklorz, line 68: Again, seems like incorrect title - it says about mouse enter but you have different event.

lib/core/src/lib/clipboard/clipboard.directive.ts
jatin2008: attached 2 hostlisteners to a single function.

lib/core/src/lib/clipboard/clipboard.directive.spec.ts
jatin2008: title changed and added keydown instead of mouse enter.

lib/core/src/lib/clipboard/clipboard.directive.spec.ts
jatin2008: title changed and added keydown instead of mouse enter.

AleksanderSklorz (approved):

PR #8167, opened 2023-01-19
[AAE-12146] Update getFeature typings

BSekula:

Added possibility to subscribe on extension config file loading.

> - [ ] Bugfix
> - [x] Feature
> - [ ] Build related changes
> - [ ] Documentation

Allow to subscribe on extension loaded

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/extensions/src/lib/services/extension.service.ts
+++ b/lib/extensions/src/lib/services/extension.service.ts
@@ -118,9 +118,9 @@ export class ExtensionService {
      * Gets features by key.
      *
      * @param key Key string, using dot notation
-     * @returns Features array found by key
+     * @returns Feature found by key
      */
-    getFeature(key: string): any[] {
+    getFeature<T = any[]>(key: string): T {
         const properties: string[] = Array.isArray(key) ? [key] : key.split('.');
         return properties.reduce((prev, curr) => prev && prev[curr], this.features) || [];
     }
--- a/lib/extensions/src/lib/services/extension.service.ts
+++ b/lib/extensions/src/lib/services/extension.service.ts
@@ -117,12 +117,13 @@ export class ExtensionService {
     /**
      * Gets features by key.
      *
-     * @param key Key string, using dot notation
+     * @param key Key string using dot notation or array of strings
+     * @param defaultValue Default value returned if feature is not found, default is empty array
      * @returns Feature found by key
      */
-    getFeature<T = any[]>(key: string): T {
-        const properties: string[] = Array.isArray(key) ? [key] : key.split('.');
-        return properties.reduce((prev, curr) => prev && prev[curr], this.features) || [];
+    getFeature<T = any[]>(key: string | string[], defaultValue: any = []): T {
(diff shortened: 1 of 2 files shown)


lib/core/src/lib/app-config/app-config.service.spec.ts
```diff
@@ -188,4 +188,10 @@ describe('AppConfigService', () => {
 
         expect(appConfigService.get('files.excluded')[0]).toBe('excluded');
     });
+
+    it('should emit on extension load', () => {
+        appConfigService.onExtensionSetup$.subscribe((config) => {
```
mauriziovitale, line 193: have you tested that the unit is red first?

DudaRobert (approved):

lib/core/src/lib/app-config/app-config.service.spec.ts
DenysVuika: the test is async, so probably never reaches the "expect"

BSekula: hmm the test is not async, `it('should emit on extension load', () => {..}` no `async` there 

Nevertheless I reverted this changes, since I will have different approach after discussion with @mauriziovitale 

Just needs to update typings for [this](https://github.com/Alfresco/alfresco-apps/pull/4773#discussion_r1087739879)

lib/extensions/src/lib/services/extension.service.ts
```diff
@@ -118,9 +118,9 @@ export class ExtensionService {
      * Gets features by key.
      *
      * @param key Key string, using dot notation
-     * @returns Features array found by key
+     * @returns Feature found by key
      */
-    getFeature(key: string): any[] {
+    getFeature<T = any[]>(key: string): T {
         const properties: string[] = Array.isArray(key) ? [key] : key.split('.');
         return properties.reduce((prev, curr) => prev && prev[curr], this.features) || [];
```
DenysVuika, line 125: note the ending of the line here, it returns an empty array while your `T` does not state that

BSekula: hmm the default is array of any, so in my opinion this is Ok. 

In case if the searched `features` is something different and we know that e.g. the searched feature is an object, then we can use generic for it.

will add a test for it as well

BSekula: ps. there is something wrong with the typings e.g. 

```
getFeature<T = any[]>(key: string): T {
        const properties: string[] = Array.isArray(key) ? [key] : key.split('.');
```
The `key` can only be `string`, but we are checking if it is an `array` and if it is an array, then we are adding it to another array, so we ended-up with type for `properties` set to nested array - `string[][]` 

i.e. 
Screenshot 2023-01-30 at 13 02 50

DenysVuika (approved):

lib/extensions/src/lib/services/extension.service.ts
DenysVuika: probably was some sort of the bug

PR #8168, opened 2023-01-19
[ACS-4257] - Resolved a11y issue around jumping focus

swapnil-verma-gl:
Clicking on filter buttons on search results no longer switches focus from button to first element

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

Clicking on any of the filter button chips on the search results page would instantly switch focus from the button to the first element in the filter layover

Focus no longer switches to the first element. Instead, it remains on the button. Pressing tab moves and then traps the focus within the layover.

> - [ ] Yes
> - [x] No

**Other information**:


lib/content-services/src/lib/search/components/search-filter-chips/search-facet-chip/search-facet-chip.component.ts
```diff
@@ -46,7 +46,6 @@ export class SearchFacetChipComponent {
     onMenuOpen() {
         if (this.menuContainer && !this.focusTrap) {
             this.focusTrap = this.focusTrapFactory.create(this.menuContainer.nativeElement);
-            this.focusTrap.focusInitialElement();
```
DenysVuika, line 49: if you remove this line, the whole sense in having a focusTrab attribute gets lost... the only thing that is left is assigning this property, but not using it...

lib/content-services/src/lib/search/components/search-filter-chips/search-widget-chip/search-widget-chip.component.ts
```diff
@@ -47,7 +47,6 @@ export class SearchWidgetChipComponent  {
     onMenuOpen() {
         if (this.menuContainer && !this.focusTrap) {
             this.focusTrap = this.focusTrapFactory.create(this.menuContainer.nativeElement);
-            this.focusTrap.focusInitialElement();
```
DenysVuika, line 50: same as above, you are essentially disabling this whole feature

lib/content-services/src/lib/search/components/search-filter-chips/search-facet-chip/search-facet-chip.component.ts
swapnil-verma-gl: The focusTrap does seem to work even after removing this line. I suspect that is because we are passing it the native element on which the focusTrap is to be initialised (see line num 48, right above this change). The user experience after this change came out to be - 

1. User performs a search (same as before, no change here)
2. Once search results are received, he clicks on one of these (search filters?) chips (same as before, no change here)
3. On clicking on the chip, the focus remains on the chip itself (changed from before, where the focus was getting jumped to the first element in the panel/popover)
4. User presses Tab. The focus is switched to the first element in the popover (changed from before, this focus was received as soon as step#3 was performed)
5. Now the user can keep on pressing 'Tab'. The focus remains within the fields in the popover (same as before, no change here)

Also, we do need to initialise the focusTrap and store it in a const reference. This is because we are using that reference when the popover is closed, to unregister the focusTrap (see [line num 52](https://github.com/Alfresco/alfresco-ng2-components/blob/b0883788338c9534f60c3a08b9cd3cdbc8984b70/lib/content-services/src/lib/search/components/search-filter-chips/search-facet-chip/search-facet-chip.component.ts#L52) onwards)

MichalKinas (approved):

DenysVuika (approved):

PR #8171, opened 2023-01-19
[ACS-4260] Added aria-hidden=false to user informational icons on file uploading list row

swapnil-verma-gl:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

Icons depicting upload status had aria-hidden flag set to true, hiding them from screen readers.

Set aria-hidden=false, so that screen readers can see the icons, and relay information appropriately

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/content-services/src/lib/upload/components/file-uploading-list-row.component.html
+++ b/lib/content-services/src/lib/upload/components/file-uploading-list-row.component.html
@@ -37,6 +37,7 @@
         </span>
 
         <mat-icon *ngIf="toggleIcon.isToggled"
+            aria-hidden="false"
             class="adf-file-uploading-row__action adf-file-uploading-row__action--cancel">
             clear
         </mat-icon>
@@ -51,6 +52,7 @@
         title="{{ 'ADF_FILE_UPLOAD.BUTTON.UPLOAD_SUCCESSFUL' | translate }}">
 
         <mat-icon
+            aria-hidden="false"
             class="adf-file-uploading-row__status adf-file-uploading-row__status--done">
             check_circle
         </mat-icon>
@@ -64,6 +66,7 @@
         >
         <mat-icon
             mat-list-icon
+            aria-hidden="false"
             class="adf-file-uploading-row__status--done">
             check_circle
         </mat-icon>
@@ -81,12 +84,14 @@
         [attr.aria-label]="'ADF_FILE_UPLOAD.ARIA-LABEL.STOP_FILE' | translate: { file: file.name }">
         <mat-icon
             *ngIf="!toggleIconCancel.isToggled"
+            aria-hidden="false"
             class="adf-file-uploading-row__status adf-file-uploading-row__status--pending">
             schedule
         </mat-icon>
 
         <mat-icon
             *ngIf="toggleIconCancel.isToggled"
+            aria-hidden="false"
             class="adf-file-uploading-row__action adf-file-uploading-row__action--remove">
             remove_circle
         </mat-icon>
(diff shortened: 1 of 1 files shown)


lib/content-services/src/lib/upload/components/file-uploading-list-row.component.html
```diff
@@ -37,6 +37,7 @@
         </span>
 
         <mat-icon *ngIf="toggleIcon.isToggled"
+            aria-hidden="false"
```
DenysVuika, line 40: what's the point in that? it's not hidden by default

DenysVuika (changes requested):

lib/content-services/src/lib/upload/components/file-uploading-list-row.component.html
swapnil-verma-gl: The aria-* attributes are used by screen readers to convey information. The aria-hidden attribute sets whether an element would be visible to a screen reader or not. It has no effect on whether it is visible on the UI. By default, i.e. without the 'aria-hidden=false', focusing on the successful/failure/etc icon would not give any information about it. However, after adding the 'aria-hidden=false' statement, it does say that it is an image when you focus on it.

DenysVuika: According to the docs: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-hidden

```
Using aria-hidden="false" will not re-expose the element to assistive technology if any of its parents specify aria-hidden="true".
```

DenysVuika: are you sure that fixes the original problem?

swapnil-verma-gl: There are no elements on this element's "parent-chain" (i.e. from <html> tag to this element) where aria-hidden=true has been set. Also, I've verified this with a screen reader extension, setting the focus to this status icon does cause the screen reader to say the word 'Image'. With aria-hidden=true setting, it does not say anything. The complete statement given when setting focus on the icon (via click) was - 

"Uploaded successfully file TestDoc 2.txt > Button > Image"

With aria-hidden=true, the statement was - 

"Uploaded successfully file TestDoc 2.txt > Button"

Here, TestDoc 2.txt is the name of the file that I was uploading. 

P.S. Note the fact that the screen reader says 'Button'. This is because this icon is set inside a 'Button' element (see [here](https://github.com/Alfresco/alfresco-ng2-components/blob/947c286dd2437bc4f05a84acc0d1511c8dcda909/lib/content-services/src/lib/upload/components/file-uploading-list-row.component.html#L46)). I'm not sure if that is the correct implementation here, since this button is not interact-able anymore and can mislead users. It stays interact-able only while the upload is in progress, where the user can click on it to cancel the upload. It seems correct when uploading a new version, where instead its wrapped inside a div.

MichalKinas: Hmm for me that looks a bit strange that we have to set explicitly set `aria-hidden` to 'false', the issues says that somewhere it is applied as `true` so maybe instead of adding it as `false` in so many places we can try checking where is it set to `true` and handle this from that place?

swapnil-verma-gl: The reason why it was applied at so many places is to ensure that all scenarios are taken care of. Based on the type of upload (file upload, folder upload, new version upload) as well as the outcome of upload (successful vs failure vs cancelled) , different HTML elements for icons are used. 
Additionally, looking at the [documentation](https://github.com/Alfresco/alfresco-ng2-components/pull/8222) for mat-icon, `aria-hidden=true` is set as default for mat-icons, since they are rarely used to convey information vital to screen readers. We can decide on if these icons convey useful information or not (Right now, screen readers only say  'Image' on this icon)

swapnil-verma-gl:
As the proposed solution is not solving the original problem satisfactorily, we are closing this PR. A new PR with a different approach will be raised in the future.

PR #8183, opened 2023-01-23
[ACS-4263] Add 'aria-pressed' attribute for reset button on content viewer page

rbahirsheth:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://alfresco.atlassian.net/browse/ACS-4263

Now screen reader reads the specific button as pressed.

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/core/src/lib/viewer/components/pdf-viewer.component.html
+++ b/lib/core/src/lib/viewer/components/pdf-viewer.component.html
@@ -106,6 +106,7 @@
 
         <button
             id="viewer-scale-page-button"
+            role="button" aria-pressed="true"
             title="{{ 'ADF_VIEWER.ARIA.FIT_PAGE' | translate }}"
             attr.aria-label="{{ 'ADF_VIEWER.ARIA.FIT_PAGE' | translate }}"
             mat-icon-button


lib/core/src/lib/viewer/components/pdf-viewer.component.html
```diff
@@ -106,6 +106,7 @@
 
         <button
             id="viewer-scale-page-button"
+            role="button" aria-pressed="true"
```
DenysVuika, line 109: why pressed should be set to true?

rbahirsheth: This reset button is not act as toggle button. This button will reset that PDF size back to 100% if size has changed(by zoom in or zoom out). I set this aria-pressed="true" because screen reader will announce "button has pressed". When you re leave  focus & click again reset button it will announce again.

DenysVuika: check the docs: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-pressed

```
The aria-pressed attribute indicates the current "pressed" state of a toggle button.
```

you are making this pressed state true by default for all the states, which is not correct

DenysVuika: You need to evaluate the state of the pressed attribute based on the actual state

DenysVuika (approved):

PR #8192, opened 2023-01-25
[ACS-4274] Add role heading and aria level in header template

rbahirsheth:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://alfresco.atlassian.net/browse/ACS-4274

Title has marked as header with aria level

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/core/src/lib/layout/components/header/header.component.html
+++ b/lib/core/src/lib/layout/components/header/header.component.html
@@ -1,8 +1,6 @@
 <mat-toolbar
     [color]="color"
-    [style.background-color]="color"
-    role="heading"
-    aria-level="1">
+    [style.background-color]="color">
     <button
         *ngIf="showSidenavToggle && position === 'start'"
         id="adf-sidebar-toggle-start"
@@ -27,7 +25,8 @@
     </a>
 
     <span
-        role="link"
+        role="heading"
+        aria-level="1"
         [attr.aria-label]="title | translate"
         [routerLink]="redirectUrl"
         fxFlex="1 1 auto"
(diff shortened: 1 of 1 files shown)


lib/core/src/lib/layout/components/header/header.component.html
```diff
@@ -27,7 +25,8 @@
     </a>
 
     <span
-        role="link"
```
MichalKinas, line 30: Is removing the `link` role correct? It was here before for some reason

rbahirsheth: Yes that is a title with link and that need to provide heading role.

MichalKinas: But then the title won't have `link` role, can this cause new accessibility issue? Since it's a router link

rbahirsheth: I think this link is working without link role too and i checked this is not introducing new accessibility issue.

MichalKinas: Link itself works but it won't be correctly announced by assistive technology, if link is not contain within `a` element it should have `link` role. Please check https://dequeuniversity.com/checklists/web/links, Axe Devtool is just a automated tool and it isn't able to detect all issues. You should always think yourself if your change will introduce new accessibility issue. Fixing one while creating new one on the way is not the right way...

rbahirsheth: added the link role back.

MichalKinas: Okay now what about `heading` role? As far as I understood from the linked Jira it should be applied to the title right?

MichalKinas (approved):

AleksanderSklorz (approved):

PR #8202, opened 2023-01-27
fix the private access to the service variables

DenysVuika:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

fix the accessors for the private service variables

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/content-services/src/lib/tag/services/tag.service.ts
+++ b/lib/content-services/src/lib/tag/services/tag.service.ts
@@ -43,7 +43,7 @@ export class TagService {
 
     private _searchApi: SearchApi;
     get searchApi(): SearchApi {
-        this._searchApi = this.searchApi ?? new SearchApi(this.apiService.getInstance());
+        this._searchApi = this._searchApi ?? new SearchApi(this.apiService.getInstance());
         return this._searchApi;
     }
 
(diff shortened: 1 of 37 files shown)


eromano (approved):

lib/content-services/src/lib/tag/services/tag.service.ts
```diff
@@ -33,16 +33,19 @@ import {
...
 
-    _tagsApi: TagsApi;
+    private _tagsApi: TagsApi;
     get tagsApi(): TagsApi {
         this._tagsApi = this._tagsApi ?? new TagsApi(this.apiService.getInstance());
         return this._tagsApi;
     }
 
-    private searchApi: SearchApi = new SearchApi(this.apiService.getInstance());
+    private _searchApi: SearchApi;
```
AleksanderSklorz, line 44: I know it is merged PR but for future - I'm wondering if that is proper approach. SearchAPI is used only by TagService internally. Additionally why somebody would like to refer to searchApi from TagService object? Getter for that seems useless because it will be used only by unit tests (we probably should not modify code only because of unit tests). It can be still mocked out in unit tests for example by spying prototype.

PR #8213, opened 2023-01-30
[ACS-4307] Label is not persistent as placeholder is being used as the only visual label for a text field

jatin2008:
…al label for a text field so added title in from and to form fields

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

Label is not persistent as placeholder is being used as the only visual label for a text field.

Label is not persistent as placeholder is being used as the only visual label for a text field so added mat-label in the form field

> - [ ] Yes
> - [x] No

**Other information**:


lib/content-services/src/lib/search/components/search-date-range/search-date-range.component.html
```diff
@@ -6,6 +6,7 @@
                 [errorStateMatcher]="matcher"
                 placeholder="{{ 'SEARCH.FILTER.RANGE.FROM-DATE' | translate }}"
                 [attr.aria-label]="'SEARCH.FILTER.RANGE.FROM-DATE' | translate"
+                [attr.title]="'SEARCH.FILTER.RANGE.FROM-DATE' | translate"
```
MichalKinas, line 9: I don't think this fix resolves the issue, as described here https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title -> "While title can be used to provide a programmatically associated label for an input element, this is not good practice. Use a label instead." So in my opinion this isn't a valid fix for this issue

lib/content-services/src/lib/search/components/search-date-range/search-date-range.component.html
```diff
@@ -1,5 +1,6 @@
 <form [formGroup]="form" novalidate (ngSubmit)="apply(form.value, form.valid)">
     <mat-form-field>
+        <mat-label>{{ 'SEARCH.FILTER.RANGE.SELECT-FROM-DATE' | translate }}</mat-label>
```
MichalKinas, line 3: I think we should use floating label as described in material documentation https://material.angular.io/components/form-field/overview#floating-label to not disrupt the UI in the form. Otherwise please consult designer or PM if non-floating label is fine here.

MichalKinas (changes requested):

lib/content-services/src/lib/search/components/search-date-range/search-date-range.component.html
MichalKinas: Btw, you didn't specify `for` attribute, without it `MatFormField` wouldn't know which label is associated with which field which will cause another accessibility issue in the future... Please read accessibility section in the documentation https://material.angular.io/components/form-field/overview#accessibility

MichalKinas (approved):

PR #8247, opened 2023-02-09
[ACS-4124] Display only as much tags as fits container in tag node list

MichalKinas:

> - [ ] Bugfix
> - [x] Feature
> - [ ] Build related changes
> - [ ] Documentation

All tags are always displayed no matter if they're fit into the container or not.

If tags display is limited, only the tags that fits are displayed together with view more button. After button is clicked all tags are displayed. https://alfresco.atlassian.net/browse/ACS-4124

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/content-services/src/lib/tag/tag-node-list.component.html
+++ b/lib/content-services/src/lib/tag/tag-node-list.component.html
@@ -13,7 +13,7 @@
         *ngIf="limitTagsDisplayed"
         class="adf-view-more-button"
         [class.adf-hidden-btn]="!calculationsDone"
-        (click)="displayAllTags()"
+        (click)="displayAllTags($event)"
         >{{ 'TAG_NODE_LIST.VIEW_MORE' | translate: { count: undisplayedTagsCount} }}
     </button>
 </div>
--- a/lib/content-services/src/lib/tag/tag-node-list.component.html
+++ b/lib/content-services/src/lib/tag/tag-node-list.component.html
@@ -1,9 +1,19 @@
-<mat-chip-list>
-    <mat-chip class="adf-tag-chips"
-              *ngFor="let currentEntry of tagsEntries; let idx = index" (removed)="removeTag(currentEntry.entry.id)">
-        <span id="tag_name_{{idx}}">{{currentEntry.entry.tag}}</span>
-        <mat-icon *ngIf="showDelete" id="tag_chips_delete_{{currentEntry.entry.tag}}"
-                  class="adf-tag-chips-delete-icon" matChipRemove>cancel
-        </mat-icon>
-    </mat-chip>
-</mat-chip-list>
+<div class="adf-tag-node-list" [class.adf-flex-column]="limitTagsDisplayed && (!calculationsDone || columnFlexDirection)" #nodeListContainer>
+    <mat-chip-list [class.adf-full-width]="limitTagsDisplayed && !calculationsDone">
+        <mat-chip class="adf-tag-chips"
+                  *ngFor="let currentEntry of tagsEntries; let idx = index"
+                  (removed)="removeTag(currentEntry.entry.id)">
+            <span id="tag_name_{{idx}}">{{currentEntry.entry.tag}}</span>
+            <mat-icon *ngIf="showDelete" id="tag_chips_delete_{{currentEntry.entry.tag}}"
+                      class="adf-tag-chips-delete-icon" matChipRemove>cancel
+            </mat-icon>
+        </mat-chip>
+    </mat-chip-list>
+    <button mat-button
+        *ngIf="limitTagsDisplayed"
+        class="adf-view-more-button"
+        [class.adf-hidden-btn]="!calculationsDone"
+        (click)="displayAllTags()"
+        >{{ 'TAG_NODE_LIST.VIEW_MORE' | translate: { count: undisplayedTagsCount} }}
+    </button>
+</div>
--- a/lib/content-services/src/lib/tag/tag-node-list.component.scss
+++ b/lib/content-services/src/lib/tag/tag-node-list.component.scss
@@ -18,7 +18,7 @@
     }
 
     .adf-hidden-btn {
-        color: transparent;
+        visibility: hidden;
     }
 
     .adf-tag-chips {
--- a/lib/content-services/src/lib/tag/tag-node-list.component.scss
+++ b/lib/content-services/src/lib/tag/tag-node-list.component.scss
@@ -1,7 +1,31 @@
 .adf-tag-node-list {
+    display: flex;
+    align-items: center;
+    flex-direction: row;
+    width: inherit;
+
+    &.adf-flex-column {
+        flex-direction: column !important;
+    }
+
+    .adf-view-more-button {
+        margin: 4px;
+        color: var(--theme-text-color);
+    }
+
+    .adf-full-width {
+        width: 100%;
+    }
+
+    .adf-hidden-btn {
+        color: transparent;
+    }
+
     .adf-tag-chips {
         color: var(--theme-primary-color-default-contrast);
         background-color: var(--theme-primary-color);
+        height: auto;
+        word-break: break-word;
     }
 
     .adf-tag-chips-delete {
--- a/lib/content-services/src/lib/tag/tag-node-list.component.scss
+++ b/lib/content-services/src/lib/tag/tag-node-list.component.scss
@@ -5,7 +5,7 @@
     width: inherit;
 
     &.adf-flex-column {
-        flex-direction: column !important;
+        flex-direction: column;
     }
 
     .adf-view-more-button {
--- a/lib/content-services/src/lib/tag/tag-node-list.component.spec.ts
+++ b/lib/content-services/src/lib/tag/tag-node-list.component.spec.ts
@@ -34,11 +34,17 @@ describe('TagNodeList', () => {
                 skipCount: 0,
                 maxItems: 100
             },
-            entries: [{
-                entry: {tag: 'test1', id: '0ee933fa-57fc-4587-8a77-b787e814f1d2'}
-            }, {entry: {tag: 'test2', id: 'fcb92659-1f10-41b4-9b17-851b72a3b597'}}, {
-                entry: {tag: 'test3', id: 'fb4213c0-729d-466c-9a6c-ee2e937273bf'}
-            }]
+            entries: [
+                {
+                    entry: {tag: 'test1', id: '0ee933fa-57fc-4587-8a77-b787e814f1d2'}
+                },
+                {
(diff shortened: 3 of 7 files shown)


lib/content-services/src/lib/tag/tag-node-list.component.html
```diff
@@ -1,9 +1,19 @@
...
-                  class="adf-tag-chips-delete-icon" matChipRemove>cancel
-        </mat-icon>
-    </mat-chip>
-</mat-chip-list>
+<div class="adf-tag-node-list" [class.adf-flex-column]="limitTagsDisplayed && (!calculationsDone || columnFlexDirection)" #nodeListContainer>
+    <mat-chip-list [class.adf-full-width]="limitTagsDisplayed && !calculationsDone">
+        <mat-chip class="adf-tag-chips"
+                  *ngFor="let currentEntry of tagsEntries; let idx = index"
+                  (removed)="removeTag(currentEntry.entry.id)">
+            <span id="tag_name_{{idx}}">{{currentEntry.entry.tag}}</span>
```
AleksanderSklorz, line 6: Do we use somewhere _ for ids? I think more ofter we use -.

lib/content-services/src/lib/tag/tag-node-list.component.scss
```diff
@@ -1,7 +1,31 @@
 .adf-tag-node-list {
+    display: flex;
+    align-items: center;
+    flex-direction: row;
+    width: inherit;
+
+    &.adf-flex-column {
+        flex-direction: column !important;
```
AleksanderSklorz, line 8: Is that !important required? Can't we do it without it?

lib/content-services/src/lib/tag/tag-node-list.component.ts
```diff
@@ -42,7 +42,20 @@ export class TagNodeListComponent implements OnChanges, OnDestroy, OnInit {
...
+    limitTagsDisplayed = false;
+
+    @ViewChild('nodeListContainer')
+    containerView: ElementRef;
+
+    @ViewChildren(MatChip)
+    tagChips: QueryList<MatChip>;
+
+    tagsEntries: TagEntry[] = [];
+    calculationsDone = false;
```
jatin2008, line 56: can we specify data type for this variable.

lib/content-services/src/lib/tag/tag-node-list.component.ts
```diff
@@ -42,7 +42,20 @@ export class TagNodeListComponent implements OnChanges, OnDestroy, OnInit {
...
+    columnFlexDirection = false;
```
jatin2008, line 57: same for this variable, data type not present.

lib/content-services/src/lib/tag/tag-node-list.component.ts
```diff
@@ -42,7 +42,20 @@ export class TagNodeListComponent implements OnChanges, OnDestroy, OnInit {
...
+    undisplayedTagsCount = 0;
```
jatin2008, line 58: same goes here

lib/content-services/src/lib/tag/tag-node-list.component.ts
```diff
@@ -87,4 +108,35 @@ export class TagNodeListComponent implements OnChanges, OnDestroy, OnInit {
             this.refreshTag();
         });
     }
+
+    displayAllTags(): void {
+        this.limitTagsDisplayed = false;
+        this.refreshTag();
+    }
+
+    private calculateTagsToDisplay() {
```
AleksanderSklorz, line 117: please add void

AleksanderSklorz:
Question: What happens if you have many short tags and many long tags? Long tag takes more than one line to fit whole text, right? So what if you have multiple short tags and long tags and order of them is mixed? Is layout correct then?

lib/content-services/src/lib/tag/tag-node-list.component.ts
DenysVuika: guys, usually we don't need explicit types in cases we provide the value, the linters even suggest removing types for Booleans and Strings (like you can see in many other places)

lib/content-services/src/lib/tag/tag-node-list.component.ts
DenysVuika: we need to decide on the return types and enable that in the linter... adding void everywhere was the "strict" mode of the Typescript, not sure if they suggest that these or not

AleksanderSklorz: So not sure if we should follow that for now or not :)

lib/content-services/src/lib/tag/tag-node-list.component.ts
MichalKinas: See above comment from Denys

lib/content-services/src/lib/tag/tag-node-list.component.ts
MichalKinas: See above comment from Denys

lib/content-services/src/lib/tag/tag-node-list.component.html
MichalKinas: Yes we do, the id was already defined when I started changing the component and I'm wondering if this would be a breaking change for someone who already started using this component, if we can change without breaking anything than I would do it, @DenysVuika what do you think?

lib/content-services/src/lib/tag/tag-node-list.component.ts
MichalKinas: That's good question, I saw that Denys added some comment to that Jira but nobody replied yet

MichalKinas:
> Question: What happens if you have many short tags and many long tags? Long tag takes more than one line to fit whole text, right? So what if you have multiple short tags and long tags and order of them is mixed? Is layout correct then?

Yes it is, if short tag comes first it is aligned with view more button in the first line and after showing all they are displayed in several lines depending on how much fits into one line, if tag is longer than one line it is collapsed into two lines I confirmed this behaviour with Lilian. When longer tag comes first it is displayed in one line (or collapsed into two) and view more button is displayed under it as we should always display at least one tag.

lib/content-services/src/lib/tag/tag-node-list.component.scss
MichalKinas: I just checked and we can remove it, thanks!

lib/content-services/src/lib/tag/tag-node-list.component.html
DenysVuika: I don't think we have any special requirements for the ID format, except it should be unique

lib/content-services/src/lib/tag/tag-node-list.component.ts
DenysVuika: for now I suggest we allow "void" based on personal preferences, otherwise it should go to linter rules to be enforced

AleksanderSklorz (approved):

lib/content-services/src/lib/tag/tag-node-list.component.scss
```diff
@@ -1,7 +1,31 @@
...
+        margin: 4px;
+        color: var(--theme-text-color);
+    }
+
+    .adf-full-width {
+        width: 100%;
+    }
+
+    .adf-hidden-btn {
+        color: transparent;
```
DenysVuika, line 21: why not just using "ngIf" to hide the button, with transparent color it might still be visible in the layout

lib/content-services/src/lib/tag/tag-node-list.component.scss
```diff
@@ -1,7 +1,31 @@
 .adf-tag-node-list {
+    display: flex;
+    align-items: center;
+    flex-direction: row;
+    width: inherit;
+
+    &.adf-flex-column {
+        flex-direction: column;
```
DenysVuika, line 8: you can just put that on the parent adf-tag-node-list class? is it's used there by default anyway

lib/content-services/src/lib/tag/tag-node-list.component.spec.ts
```diff
@@ -121,5 +122,66 @@ describe('TagNodeList', () => {
             const deleteButton: any = element.querySelector('#tag_chips_delete_test1');
             expect(deleteButton).not.toBeNull();
         });
+
+        it('should not render view more button by default', async () => {
+            component.ngOnChanges();
```
DenysVuika, line 127: I think this one is running automatically during the "detecChanges" call below

lib/content-services/src/lib/tag/tag-node-list.component.scss
MichalKinas: This is done on purpose, intially the button is rendered but is "hidden" so I can get its exact width to calculate how many tags can fit into container

AleksanderSklorz: @MichalKinas @DenysVuika maybe as replacement for transparent we can use visibility:hidden?

lib/content-services/src/lib/tag/tag-node-list.component.scss
MichalKinas: By default it is set to `row` and `column` is used only at the beginning for calculations and in case of first tag being too wide to fit into one line with button, in other cases `row` should remain

lib/content-services/src/lib/tag/tag-node-list.component.spec.ts
MichalKinas: It is but then without calling it tags are not loaded

lib/content-services/src/lib/tag/tag-node-list.component.scss
MichalKinas: The problem with visibility changes is the fact that it then shows 0 as width of the button so I can't perform calculations in this case

AleksanderSklorz: hmm how you apply hidden? Please check here: https://developer.mozilla.org/en-US/docs/Web/CSS/visibility. It still has correct dimensions: 
image

MichalKinas: I think in the same way as described in the docs, I'll check once again maybe I missed something

AleksanderSklorz: Ok, I'm asking because there is also angular's [hidden] so in case you used Angular's hidden then you may try pure css visibility: hidden.

MichalKinas: I've just checked and `visibility: hidden` works as well, must have missed something earlier, @DenysVuika which solution would you prefer?

AleksanderSklorz: thank you for verifying @MichalKinas :)

DenysVuika (approved):

PR #8344, opened 2023-03-07
[AAE-12501] Align JS API

alep85:
…get rid of the js-api oauth init

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://alfresco.atlassian.net/browse/AAE-12501

> - [ ] Yes
> - [ ] No

**Other information**:
To merge it after this [PR](https://github.com/Alfresco/alfresco-js-api/pull/1545) has been merged


lib/core/src/lib/auth/oidc/oidc-authentication.service.ts
```diff
@@ -46,11 +46,6 @@ export class OIDCAuthenticationService extends BaseAuthenticationService {
         private readonly auth: AuthService
     ) {
         super(alfrescoApi, appConfig, cookie, logService);
-        this.alfrescoApi.alfrescoApiInitialized.subscribe(() => {
```
popovicsandras, line 49: don't we need it at all?

lib/process-services-cloud/src/lib/process/process-list/services/process-list-cloud.service.ts
```diff
@@ -16,20 +16,21 @@
...
 import { ProcessListCloudSortingModel } from '../models/process-list-sorting.model';
 import { BaseCloudService } from '../../../services/base-cloud.service';
 import { map } from 'rxjs/operators';
+import { AdfHttpClient } from '@alfresco/adf-core/api';
 
 @Injectable({ providedIn: 'root' })
 export class ProcessListCloudService extends BaseCloudService {
 
-    constructor(apiService: AlfrescoApiService,
+    constructor(adfHttpClient: AdfHttpClient,
```
popovicsandras, line 30: Don't we use it only in case of OIDC auth? Completely swapping the old AlfrescoApiService to this service, will it work with the current auth implementation?

popovicsandras (approved):

lib/core/src/lib/auth/oidc/oidc-authentication.service.ts
alep85: I removed it because the `'logged-in'` event is emitted by the js-api authentication that we want get rid of, but we can't removed it at all, we need to trigger the onLogin because is needed to handle the file version, I'll change that logic using the new auth instead using the alfresco api

PR #8359, opened 2023-03-13
[ACS-4746] Returning focus to trigger element after restore option is…

arohilaGL:
… selected and manage version modal is closed

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

Once the manage version modal is opened and restore option is selected. After the restore option is performed, when you close the dialog, the focus does not return to the trigger element.

Once the manage version modal is opened and restore option is selected. After the restore option is performed, when you close the dialog, the focus returns to the trigger element.

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/core/src/lib/datatable/components/datatable/datatable.component.ts
+++ b/lib/core/src/lib/datatable/components/datatable/datatable.component.ts
@@ -224,6 +224,7 @@ export class DataTableComponent implements OnInit, AfterContentInit, OnChanges, 
     isSelectAllIndeterminate: boolean = false;
     isSelectAllChecked: boolean = false;
     selection = new Array<DataRow>();
+    selectedRowId: string = '';
 
     isDraggingHeaderColumn = false;
     hoveredHeaderColumnIndex = -1;
@@ -285,6 +286,7 @@ export class DataTableComponent implements OnInit, AfterContentInit, OnChanges, 
 
     ngOnChanges(changes: SimpleChanges) {
         this.initAndSubscribeClickStream();
+        this.setRowAsContextSource();
 
         const dataChanges = changes['data'];
         const rowChanges = changes['rows'];
@@ -780,8 +782,13 @@ export class DataTableComponent implements OnInit, AfterContentInit, OnChanges, 
     }
 
     markRowAsContextMenuSource(selectedRow: DataRow): void {
+        this.selectedRowId = selectedRow.id ? selectedRow.id : '';
         this.data.getRows().forEach((row) => row.isContextMenuSource = false);
         selectedRow.isContextMenuSource = true;
+    }
+
+    setRowAsContextSource(): void {
+        this.data.getRows().forEach((row) => row.isContextMenuSource = this.selectedRowId === row.id ? true : false);
     }
 
     getSortingKey(): string | null {


lib/core/src/lib/datatable/components/datatable/datatable.component.ts
```diff
@@ -780,10 +782,15 @@ export class DataTableComponent implements OnInit, AfterContentInit, OnChanges,
     }
 
     markRowAsContextMenuSource(selectedRow: DataRow): void {
+        this.selectedRowId = selectedRow.id ? selectedRow.id : '';
         this.data.getRows().forEach((row) => row.isContextMenuSource = false);
         selectedRow.isContextMenuSource = true;
     }
 
+    setRowAsContextSource(): void {
+        this.data.getRows().forEach((row) => row.isContextMenuSource = this.selectedRowId === row.id ? true : false);
```
AleksanderSklorz, line 791: can be a bit shorter: 
```ts
this.data.getRows().forEach((row) => row.isContextMenuSource = this.selectedRowId === row.id);
```

Or maybe even better: 
```ts
this.data.getRows().find((row) => this.selectedRowId = row.id)?.isContextMenuSource = true;
```

Second is even better because it does not iterate over all rows. It is just looking for first row which match condition and then loop stopped

AleksanderSklorz: or even better if you use find: 
```ts
this.data.getRows().find((row) => this.selectedRowId === row.id)?.isContextMenuSource = true;
```

because then it is just looking for first row which meets condition during iteration and stopped loop.

arohilaGL: Indeed, it makes sense, but we need to mark 'isContextMenuSource' for other rows as false, as well. The test case that you have written check that for rest of the rows 'isContextMenuSource' must be false.

AleksanderSklorz: I guess you set it to true because rows are reloaded right? So is not isContextMenuSource just undefined in that case? If yes then it is fine, it does not need to be false. It is js so undefined is falsy similar to false. 

So what is value of isContextMenuSource? If it is undefined then you don't need to set it to false because it is falsy too, unless it is true then yes you need set it to false.

arohilaGL: Okay, in that case, 'this.data.getRows().find((row) => this.selectedRowId === row.id)?.isContextMenuSource = true;' should work. Thanks for the comment.

AleksanderSklorz (approved):

lib/core/src/lib/datatable/components/datatable/datatable.component.ts
```diff
@@ -285,6 +286,9 @@ export class DataTableComponent implements OnInit, AfterContentInit, OnChanges,
 
     ngOnChanges(changes: SimpleChanges) {
         this.initAndSubscribeClickStream();
+        if(this.selectedRowId) {
```
DenysVuika, line 289: missing unit tests for all the added code

DenysVuika (changes requested):

lib/core/src/lib/datatable/components/datatable/datatable.component.ts
arohilaGL: working on it.

arohilaGL: unit test added for the change.

lib/core/src/lib/datatable/components/datatable/datatable.component.ts
```diff
@@ -780,10 +784,18 @@ export class DataTableComponent implements OnInit, AfterContentInit, OnChanges,
     }
 
     markRowAsContextMenuSource(selectedRow: DataRow): void {
+        this.selectedRowId = selectedRow.id ? selectedRow.id : '';
         this.data.getRows().forEach((row) => row.isContextMenuSource = false);
         selectedRow.isContextMenuSource = true;
     }
 
+    private setRowAsContextSource(): void {
```
DenysVuika, line 792: you are running this on `ngOnChanges`, that means hundreds of times per second potentially. you can call this from the `markRowAsContextMenuSource` as this looks like a one-time operation. You already set the `selectedRowId` there, so why not updating the row?

DenysVuika (changes requested):

lib/core/src/lib/datatable/components/datatable/datatable.component.ts
DenysVuika: or use the get/set for the `selectedRowId` to perform this 1 time only when needed. Please avoid polluting ngOnChanges with stuff that does not need to run a lot of times

arohilaGL: Its running only two times. And it needs to be executed in ngOnChanges and not in markRowAsContextMenuSource because the data gets refreshed after the operation of restore in manage versions is performed, markRowAsContextMenuSource is called before the said operation.

DenysVuika: I understand that it needs to be run multiple times, but in your case you run it on *each* property change, so the performance will have a hit. Check the ngOnChanges docs, each time one of the properties will change it state, you will be looping through the data and performing the same task again.

PR #8382, opened 2023-03-15
[ACS-4117] Added possibility to create multiple categories

nikita-web-ua:

> - [ ] Bugfix
> - [X] Feature
> - [ ] Build related changes
> - [ ] Documentation

Currently it is possible to create only one category at the same time.

Now it is possible to create multiple categories/ subcategories with one request.

> - [ ] Yes
> - [X] No

**Other information**:
https://alfresco.atlassian.net/browse/ACS-4117


lib/content-services/src/lib/category/services/category.service.ts
```diff
@@ -58,14 +58,14 @@ export class CategoryService {
...
      *
      * @param parentCategoryId The identifier of a parent category.
-     * @param payload Created category body
-     * @return Observable<CategoryEntry>
+     * @param payload List of categories to be created.
+     * @return Observable<CategoryPaging | CategoryEntry>
      */
-    createSubcategory(parentCategoryId: string, payload: CategoryBody): Observable<CategoryEntry> {
-        return from(this.categoriesApi.createSubcategory(parentCategoryId, [payload], {}));
+    createSubcategories(parentCategoryId: string, payload: CategoryBody[]): Observable<CategoryPaging | CategoryEntry> {
```
AleksanderSklorz, line 67: Something is wrong with type I think - you say CategoryPaging | CategoryEntry but if you check what that function does it just returns following line: 
from(this.categoriesApi.createSubcategory(parentCategoryId, payload, {}));

where this.categoriesApi.createSubcategory returns Promise<CategoryEntry> type so from where CateegoryPaging comes?

I think you should not use createSubcategory because it is for creating single category and it is why it returns CategoryEntry. You need probably write new function in js-api to handle multiple categories.

lib/content-services/src/lib/category/services/category.service.ts
```diff
@@ -58,14 +58,14 @@ export class CategoryService {
...
+        return from(this.categoriesApi.createSubcategory(parentCategoryId, payload, {}));
```
AleksanderSklorz, line 68: As mentioned in previous comment that does not look correct because your function is "createSubcategories" so why createSubcategories calls createSubcategory which suggests creating single category and it returns even type for single result?

nikita-web-ua: `createSubcategory` in jsAPI receives a list of category bodies as second parameter, so maybe I should change name and return type in jsAPI for this function to `createSubcategories`? What do you think @AleksanderSklorz ?

AleksanderSklorz: sounds like good idea

lib/content-services/src/lib/category/services/category.service.ts
nikita-web-ua: updated jsAPI here https://github.com/Alfresco/alfresco-js-api/pull/1549 to make it more clear

AleksanderSklorz (approved):

MichalKinas (approved):

PR #8390, opened 2023-03-16
[ACS-4125] Add support for the tags search facet

TomikaArome:

> - [ ] Bugfix
> - [x] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://alfresco.atlassian.net/browse/ACS-4125

Added three new configuration options: `facetOrder`, `bucketSortBy` and `bucketSortDirection` to allow sorting of the facets and buckets

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/content-services/src/lib/search/models/facet-field.interface.ts
+++ b/lib/content-services/src/lib/search/models/facet-field.interface.ts
@@ -40,11 +40,22 @@ export interface FacetFieldSettings {
     allowUpdateOnChange?: boolean;
     /* allow the user show/hide default search actions */
     hideDefaultAction?: boolean;
-    displayAfterField?: string;
-    sorting?: FacetBucketOrder;
+    /* a number to compare to other facets to determine the order in which they will appear */
+    facetOrder?: number;
+    /* the field used to sort the buckets */
+    bucketSortBy?: FacetBucketSortBy;
+    /* the direction in which the buckets are ordered */
+    bucketSortDirection?: FacetBucketSortDirection;
 }
 
-export enum FacetBucketOrder {
+// eslint-disable-next-line no-shadow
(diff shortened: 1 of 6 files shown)


lib/content-services/src/lib/search/models/facet-field.interface.ts
```diff
@@ -40,4 +40,22 @@ export interface FacetFieldSettings {
...
+    /* a number to compare to other facets to determine the order in which they will appear */
+    facetOrder?: number;
+    /* the field used to sort the buckets */
+    bucketSortBy?: FacetBucketSortBy;
+    /* the direction in which the buckets are ordered */
+    bucketSortDirection?: FacetBucketSortDirection;
+}
+
+// eslint-disable-next-line no-shadow
+export enum FacetBucketSortBy {
```
DenysVuika, line 52: the shadow issue is due to having more than one interface/enum/etc per file. you can fix this issue by moving enums to separate files

DenysVuika (approved):

MichalKinas (approved):

AleksanderSklorz (approved):

lib/content-services/src/lib/search/models/facet-field.interface.ts
TomikaArome: I'll make a mental note of this for the future. I did check the use of this `no-shadow` rule in the rest of the project, and disabling it seems to appear numerous times. It seems like the rule fails due to a false positive with enums: https://github.com/typescript-eslint/typescript-eslint/issues/2483

Given that the PR is green and I've not received any other review comments, I'll merge it as is for now. If we want to improve this in the future, perhaps we could improve it to either use the other rule suggested in the above link, or separate every enum to their own file as you suggest.

PR #8414, opened 2023-03-25
Update branch for JS-API PR#756

alfresco-build:


AleksanderSklorz:
@eromano @DenysVuika not sure if anyone was working on fixing that but seems like now versions for js-api are updated instead of ""

eromano (approved):

PR #8417, opened 2023-03-27
[ACS-4126] viewer allow assigning tags to content basic

AleksanderSklorz:
**Please check if the PR fulfills these requirements**

> - [x] The commit message follows our [guidelines](https://github.com/Alfresco/alfresco-ng2-components/wiki/Commit-format)
> - [x] Tests for the changes have been added (for bug fixes / features)
> - [x] Docs have been added / updated (for bug fixes / features)

**What kind of change does this PR introduce?** (check one with "x")

> - [ ] Bugfix
> - [x] Feature
> - [ ] Code style update (formatting, local variables)
> - [ ] Refactoring (no functional changes, no api changes)
> - [ ] Build related changes
> - [ ] Documentation
> - [ ] Other... Please describe:

**What is the current behaviour?** (You can also link to an open issue here)
https://alfresco.atlassian.net/browse/ACS-4126

**What is the new behaviour?**
Moved TagsCreator component from alfresco-applications as we need to reuse that component in additional places. Displaying tags creation in Viewer for files and folders. 

**Does this PR introduce a breaking change?** (check one with "x")

> - [ ] Yes
> - [x] No

If this PR contains a breaking change, please describe the impact and migration path for existing applications: ...

**Other information**:
JS-API PR: https://github.com/Alfresco/alfresco-js-api/pull/1557
alfresco-applications PR: https://github.com/Alfresco/alfresco-applications/pull/100
ACA PR: https://github.com/Alfresco/alfresco-content-app/pull/3093

PR #8451, opened 2023-04-07
[ACS-4523] Assigning content to categories

MichalKinas:

> - [ ] Bugfix
> - [x] Feature
> - [ ] Build related changes
> - [ ] Documentation

Assigning content to categories is impossible from ACA. https://alfresco.atlassian.net/browse/ACS-4523

There is a new component that allows both create and manage categories, depending on selected mode. Assign mode will be used in ACA to allow assigning content to categories and create mode will be used in control center to allow creating multiple categories.

> - [ ] Yes
> - [x] No

**Other information**:


lib/content-services/src/lib/category/categories-management/categories-management.component.html
```diff
@@ -0,0 +1,76 @@
...
+                data-automation-id="categories-remove-category-button"
+                mat-icon-button
+                (click)="removeCategory(category)"
+                [attr.title]="removeCategoryTitle | translate"
+                [disabled]="disableRemoval">
+                <mat-icon>remove</mat-icon>
+            </button>
+        </span>
+    </div>
+    <div *ngIf="((!categoryNameControlVisible && categories.length)) ||categoryNameControlVisible"
```
AleksanderSklorz, line 22: can you add space after || ?

lib/content-services/src/lib/category/categories-management/categories-management.component.html
```diff
@@ -0,0 +1,76 @@
...
+            <mat-label id="adf-category-name-input-label">
+                {{'CATEGORIES_MANAGEMENT.NAME' | translate}}
+            </mat-label>
+            <input
+                matInput
+                autocomplete="off"
+                [formControl]="categoryNameControl"
+                (keyup.enter)="addCategory()"
+                aria-labelledby="adf-category-name-input-label"
+                adf-auto-focus
```
AleksanderSklorz, line 36: Don't you need to import ContentDirectiveModule to use that directive?

docs/content-services/components/categories-management.component.md
```diff
@@ -0,0 +1,44 @@
...
+</adf-categories-management>
+```
+
+## Class members
+
+### Properties
+
+| Name | Type | Default value | Description |
+| ---- | ---- | ------------- | ----------- |
+| categories | [`Category`](https://github.com/Alfresco/alfresco-js-api/blob/develop/src/api/content-rest-api/docs/Category.md) | [] | List of categories to assign/create. |
```
nikita-web-ua, line 32:
```suggestion
| categories | [`Category`](https://github.com/Alfresco/alfresco-js-api/blob/develop/src/api/content-rest-api/docs/Category.md) [] | [] | List of categories to assign/create. |
```

lib/content-services/src/lib/category/categories-management/categories-management.component.html
```diff
@@ -0,0 +1,76 @@
...
+            </button>
+        </span>
+    </div>
+    <div *ngIf="((!categoryNameControlVisible && categories.length)) ||categoryNameControlVisible"
+        [hidden]="!categoryNameControlVisible"
+        class="adf-category-name-field">
+        <mat-form-field>
+            <mat-icon matPrefix>search</mat-icon>
+            <mat-label id="adf-category-name-input-label">
+                {{'CATEGORIES_MANAGEMENT.NAME' | translate}}
```
nikita-web-ua, line 28:
```suggestion
                {{ 'CATEGORIES_MANAGEMENT.NAME' | translate }}
```

lib/content-services/src/lib/category/categories-management/categories-management.component.ts
```diff
@@ -0,0 +1,287 @@
...
+            .pipe(takeUntil(this.onDestroy$))
+            .subscribe(() => this.setCategoryNameControlErrorMessageKey());
+
+        this.setCategoryNameControlErrorMessageKey();
+        this.noCategoriesTitle = this.isCRUDMode ?
+            'CATEGORIES_MANAGEMENT.NO_CATEGORIES_CREATED':
+            'CATEGORIES_MANAGEMENT.NO_CATEGORIES_ASSIGNED';
+        this.removeCategoryTitle = this.isCRUDMode ?
+            'CATEGORIES_MANAGEMENT.DELETE_CATEGORY':
+            'CATEGORIES_MANAGEMENT.UNASSIGN_CATEGORY';
```
AleksanderSklorz, line 124: Small thing but I'm wondering if we need unassign label because in theory if you have categories which you want to assign then they are not yet assigned so technically you don't "unassign" them but you "remove" them from top list. But that's small thing, if you prefer that then that's fine to me :)

lib/content-services/src/lib/category/categories-management/categories-management.component.spec.ts
```diff
@@ -0,0 +1,487 @@
...
+
+            expect(component.categories.length).toBe(2);
+            expect(component.categories[1].name).toBe('testCat4');
+            expect(categoriesChangeSpy).toHaveBeenCalledOnceWith(component.categories);
+            expect(component.existingCategories.length).toBe(2);
+            discardPeriodicTasks();
+            flush();
+        }));
+
+        it('should not add back to existing categories when category was assigned intially', fakeAsync(() => {
```
AleksanderSklorz, line 338: typo, should be initially

lib/content-services/src/lib/category/categories-management/categories-management.component.spec.ts
```diff
@@ -0,0 +1,487 @@
...
+            ],
+            teardown: { destroyAfterEach: true }
+        });
+
+        fixture = TestBed.createComponent(CategoriesManagementComponent);
+        component = fixture.componentInstance;
+        categoryService = TestBed.inject(CategoryService);
+    });
+
+    function getNoCategoriesMessage(): HTMLParagraphElement {
```
AleksanderSklorz, line 70: All usages looks like: getNoCategoriesMessage().textContent.trim() so you can just return string directly to don't repeat whole chain

lib/content-services/src/lib/category/services/category.service.ts
AleksanderSklorz, line 1: Could you please update documentation for that file?

AleksanderSklorz:
@MichalKinas I see that clicking at + button closes input for Tags if input for tags is already visible. Is that intended change? I think Agnieszka told that we can see both actually because there is no common scrollbar. Can you verify with her?

lib/content-services/src/lib/category/categories-management-mode.ts
AleksanderSklorz, line 1: Just a thought - maybe this file should be inside categories-management as it is mode of that component so it is highly linked with that component?

lib/content-services/src/lib/category/category.module.ts
```diff
@@ -0,0 +1,40 @@
...
+import { MaterialModule } from '../material.module';
+import { CategoriesManagementComponent } from './categories-management/categories-management.component';
+
+@NgModule({
+    imports: [
+        CommonModule,
+        CoreModule,
+        MaterialModule,
+        TranslateModule,
+        ContextMenuModule
```
AleksanderSklorz, line 31: That module seems redundant because you don't use context menu

lib/content-services/src/lib/category/services/category.service.ts
AleksanderSklorz, line 1: Missing unit tests for changes in that file

lib/content-services/src/lib/content-metadata/components/content-metadata/content-metadata.component.scss
```diff
@@ -1,3 +1,7 @@
+[hidden] {
```
AleksanderSklorz, line 1: Moving it to top of scss file will cause probably that all places will be affected with these styles because we have none encapsulation for that component. 

In theory maybe it has sense that hidden causes that element is hidden but then if we really want to make it global style then it should be in some file with global styles. In other case better to have it nested because it may be strange that styles for metadata component affect on all places.

lib/content-services/src/lib/content-metadata/components/content-metadata/content-metadata.component.ts
```diff
@@ -282,43 +339,62 @@ export class ContentMetadataComponent implements OnChanges, OnInit, OnDestroy {
...
+        return zip(properties$, contentTypeProperty$)
+            .pipe(map(([properties, contentTypeProperty]) => {
+                const filteredProperties = contentTypeProperty.filter((property) => properties.findIndex((baseProperty) => baseProperty.key === property.key) === -1);
+                return [...properties, ...filteredProperties];
+            }));
     }
 
-    public canExpandTheCard(group: CardViewGroup): boolean {
-        return group.title === this.displayAspect;
+    private isEmpty(value: any): boolean {
```
AleksanderSklorz, line 367: can you set type?

lib/content-services/src/lib/content-metadata/components/content-metadata/content-metadata.component.ts
```diff
@@ -99,30 +101,37 @@ export class ContentMetadataComponent implements OnChanges, OnInit, OnDestroy {
     @Input()
     displayTags = false;
```
AleksanderSklorz, line 102: You may need similar input for categories. Notice that metadata component is used in couple places, not only in viewer which we change but also for example in some places in outside content-services like process-services-cloud and we don't want probably affect other places

lib/content-services/src/lib/category/categories-management/categories-management.component.ts
```diff
@@ -0,0 +1,287 @@
...
+    private _existingCategoriesLoading = false;
+    private _typing = false;
+    private _existingCategoriesPanelVisible: boolean;
+    existingCategoriesListLimit = 15;
+    noCategoriesTitle = '';
+    removeCategoryTitle = '';
+    initialCategories: Category[] = [];
+
+    @Input()
+    categories: Category[] = [];
```
AleksanderSklorz, line 70: Could you add jsdoc to whole public API? It was commented by Denys too here https://github.com/Alfresco/alfresco-ng2-components/pull/8417#discussion_r1151149349. 

So please add doc to inputs, outputs, public functions etc.

lib/content-services/src/lib/content-metadata/components/content-metadata/content-metadata.component.ts
```diff
@@ -282,43 +339,62 @@ export class ContentMetadataComponent implements OnChanges, OnInit, OnDestroy {
...
+    private saveCategories(): { [key: string]: Observable<CategoryPaging | CategoryEntry | void> } {
+        const observables: { [key: string]: Observable<CategoryPaging | CategoryEntry | void> } = {};
+        if (this.categories) {
+            this.assignedCategories.forEach((assignedCategory) => {
+                if (this.categories.every((category) => category.name !== assignedCategory.name)) {
+                    observables[`Removing ${assignedCategory.id}`] = this.categoryService.unlinkNodeFromCategory(this.node.id, assignedCategory.id);
+                }
+            });
+            const categoryLinkBodies: CategoryLinkBody[] = [];
+            this.categories.forEach((category) => {
```
AleksanderSklorz, line 388: This can be shorter, something like:

```ts
const categoryLinkBodies = this.categories.map((category) => {
    const categoryLinkBody = new CategoryLinkBody();
    categoryLinkBody.categoryId = category.id;
    return category;
});
```

Thanks that you can initialize and declare categoryLinkBodies in the same line (you don't need even initialize that with empty array) and you don't need push categories manually to array.

AleksanderSklorz:
I think I found a bug. There is no "Create" button as we can't create category during assigning it but there is still possibility to confirm new category by hitting "Enter". You should disable Enter too, not only hide Create button because that causes user can still create categories.

lib/content-services/src/lib/category/categories-management/categories-management.component.ts
AleksanderSklorz, line 1: Agnieszka asked me some time ago about scrolling to input in case top list is long so I added following line of code: 
image

I see that line is missing in your code. You can't create new categories but you can have long list of already assigned categories so it may be still helpful to scroll to input after hitting + icon. 
Thanks that in case long top list user does not need to scroll into input because it is scrolled manually

AleksanderSklorz:
I noticed an issue where when you type something, remove it and type again - you can see some results but after moment they are gone. If there is no matched results then there should not be visible list for moment, there should be just message 

https://user-images.githubusercontent.com/115619721/230669344-ef725e93-7a32-4ed7-9980-d66bc9f4c938.mov

AleksanderSklorz:
For tags we were asked to display exact match as first row. For categories it does not work like that because exact match is not in first row: 
image

AleksanderSklorz:
There is issue saying that category already exists but I guess it does not make sense for ACA as we can't create category?

image

AleksanderSklorz:
Should not "No existing Categories" be hidden when spinner is shown?

AleksanderSklorz:
I found issue with removing already assigned category. If search matches to category's name and you unlink it then it should return to bottom list so user can pick it again in case he unlinked it by accident but actually it is just gone. 

https://user-images.githubusercontent.com/115619721/230673169-4578958b-9f47-4063-b10d-e51eb7bf14cb.mov

AleksanderSklorz:
If user types something in search and there are some results and then he hides input and show input again - input's value is still present (what is ok because user does not need retype value in case it is hidden by accident) but list of existing categories is disappeared even if previously it was visible 

https://user-images.githubusercontent.com/115619721/230673726-2168ea22-12b7-498a-a0a7-f35cde145db4.mov

AleksanderSklorz:
plus and minus icons are aligned incorrectly in comparison with the same icons from tags: 
image

AleksanderSklorz:
Looks like ACA env is broken a bit for our epic but I found one issue which occurs only on your branch but on ACA works fine. If you do following actions like on video then you can't edit tags and categories anymore, even after clicking at Edit mode icon. 


https://user-images.githubusercontent.com/115619721/230674902-d4bf0ff7-f956-4be2-9ff0-b54351bed275.mov

AleksanderSklorz:
I see also error when you type text which is equal name of category from top list. It is redundant I guess as you can't create category: 
image

AleksanderSklorz:
Do we need that message above input? In tags we don't show any message when input is visible above that input: 
image

AleksanderSklorz:
It is displayed a bit inconsistent if you do specific steps like on video. If you select category then it has path. Save it - still has path. Do any action and discard changes - path is lost. Won't that be surprising for user that he lost path and his list is different? When he clicks at discard changes then user can expect that he gets exactly the same results like earlier. What do you think?

https://user-images.githubusercontent.com/115619721/230676286-f8c8e94b-86a3-481c-8142-3e15ac54db7f.mov

AleksanderSklorz:
I found issue during review which occurs on both - tags and categories. I raised already jira for that and fix for tags, could you please do the same for categories? Here jira for tags, there is also linked PR to that fix https://alfresco.atlassian.net/browse/ACS-5021

We should fix that because that was working before our changes.

AleksanderSklorz:
I found issue - if you select existing category, then deselect it, then you can't see error that that category exists so you can create duplication: 

https://user-images.githubusercontent.com/115619721/230713015-86744cd4-4ac4-41c5-bb11-0bc2094738b3.mov

AleksanderSklorz:
Small thing but it is inconsistent: 
Remove icon hides for tags but for categories it becomes disabled during saving.

lib/content-services/src/lib/category/categories-management/categories-management.component.html
MichalKinas: Added

lib/content-services/src/lib/category/categories-management/categories-management.component.html
MichalKinas: Right, imported

lib/content-services/src/lib/category/categories-management/categories-management.component.html
MichalKinas: Fixed

lib/content-services/src/lib/category/categories-management/categories-management.component.ts
MichalKinas: Yes but it also depends on the case, since you can also start from having some categories assigned and then technically you're unassigning them so I think it would be more consistent to still call it unassign, we can discuss this with Agnieszka if you like

lib/content-services/src/lib/category/categories-management/categories-management.component.spec.ts
MichalKinas: Fixed

lib/content-services/src/lib/category/services/category.service.ts
MichalKinas: Updated

MichalKinas:
> Not sure if that is only my local issue but when I'm typing something then spinner is not centered horizontally for me. It is centered horizontally in tags

Right, fixed now

MichalKinas:
> @MichalKinas I see that clicking at + button closes input for Tags if input for tags is already visible. Is that intended change? I think Agnieszka told that we can see both actually because there is no common scrollbar. Can you verify with her?

You're right, I forgot to change it, fixed now

lib/content-services/src/lib/category/category.module.ts
MichalKinas: Removed

lib/content-services/src/lib/category/services/category.service.ts
MichalKinas: Added

lib/content-services/src/lib/content-metadata/components/content-metadata/content-metadata.component.scss
MichalKinas: Yeah that might interfere sometimes, for now nested it locally

lib/content-services/src/lib/content-metadata/components/content-metadata/content-metadata.component.ts
MichalKinas: `isEmpty` is then used to check if `CardViewItem` display value is not empty and it's set to any as well, I just moved the method to group public and private methods. To change this any we would have to change `CardViewItem` display value type and this would require some additional refactoring. We can create a Jira to investigate this

lib/content-services/src/lib/content-metadata/components/content-metadata/content-metadata.component.ts
MichalKinas: Added

lib/content-services/src/lib/category/categories-management/categories-management.component.ts
MichalKinas: Added

lib/content-services/src/lib/content-metadata/components/content-metadata/content-metadata.component.ts
MichalKinas: Shortened

lib/content-services/src/lib/category/categories-management/categories-management.component.ts
MichalKinas: Added

MichalKinas:
> I noticed an issue where when you type something, remove it and type again - you can see some results but after moment they are gone. If there is no matched results then there should not be visible list for moment, there should be just message

Actually I'm not sure why it's shown for a moment, I need to investigate this

MichalKinas:
> For tags we were asked to display exact match as first row. For categories it does not work like that because exact match is not in first row. I guess it can be caused because of added path. image

For categories it has to work differently since we can have categories with the same name in multiple places so we agreed to display the 15 relevant outcomes from server in sorted order

MichalKinas:
> There is issue saying that category already exists but I guess it does not make sense for ACA as we can't create category?
> 
> image

Right, now it validated only in CRUD mode

MichalKinas:
> Should not "No existing Categories" be hidden when spinner is shown?

Hidden now

MichalKinas:
> If user types something in search and there are some results and then he hides input and show input again - input's value is still present (what is ok because user does not need retype value in case it is hidden by accident) but list of existing categories is disappeared even if previously it was visible

Right fixed now

MichalKinas:
> plus and minus icons are aligned incorrectly in comparison with the same icons from tags: image

Aligned now

MichalKinas:
> Do we need that message above input? In tags we don't show any message when input is visible above that input: image

Fixed

MichalKinas:
> It is displayed a bit inconsistent if you do specific steps like on video. If you select category then it has path. Save it - still has path. Do any action and discard changes - path is lost. Won't that be surprising for user that he lost path and his list is different? When he clicks at discard changes then user can expect that he gets exactly the same results like earlier. What do you think?

As I mentioned this is related to the BE feature I did, it will be fixed as soon as it will be merged

MichalKinas:
> I found issue during review which occurs on both - tags and categories. I raised already jira for that and fix for tags, could you please do the same for categories? Here jira for tags, there is also linked PR to that fix https://alfresco.atlassian.net/browse/ACS-5021
> 
> We should fix that because that was working before our changes.

Sure, fixed for categories

lib/content-services/src/lib/category/categories-management/categories-management.component.ts
AleksanderSklorz: As I said - it's small thing and I'm not sure too what is better, just wanted to highlight it :) We can keep that for now

MichalKinas:
> I found issue - if you select existing category, then deselect it, then you can't see error that that category exists so you can create duplication:

Wait but in CRUD mode you can't select categories you can only create them, existing categories in this case are only pulled to check what's already created. And in ASSIGN mode you can't create categories

lib/content-services/src/lib/category/categories-management/categories-management.component.spec.ts
```diff
@@ -0,0 +1,487 @@
...
```
AleksanderSklorz, line 70: Incorrect type, you return now string instead of whole element so type can be string as well

MichalKinas:
> Small thing but it is inconsistent: Remove icon hides for tags but for categories it becomes disabled during saving. It may hide or be disabled, both solutions are fine to me but it should behave in the same way for each place to keep consistency. Probably less changes are required to make it just hidden.

I asked Agnieszka to confirm which path is better, I think disabling make layout more stable since icon don't disappear on save but we should wait for her decision

docs/content-services/services/category.service.md
```diff
@@ -50,6 +50,20 @@ Manages categories in Content Services.
     -   _categoryId:_ `string`  - The identifier of a category.
     -   _payload:_ `CategoryBody`  - Updated category body
     -   **Returns** [`Observable`](http://reactivex.io/documentation/observable.html)`<CategoryEntry>` - [`Observable`](http://reactivex.io/documentation/observable.html)&lt;CategoryEntry>
+-   **getCategoryLinksForNode**(nodeId: `string`): [`Observable`](http://reactivex.io/documentation/observable.html)`<`[`CategoryPaging`](https://github.com/Alfresco/alfresco-js-api/blob/master/src/api/content-rest-api/docs/CategoryPaging.md)`>`<br/>
+    Searches categories by their name.
```
AleksanderSklorz, line 54: Incorrect description, looks like description for searchCategories

lib/content-services/src/lib/category/categories-management/categories-management.component.spec.ts
MichalKinas: Fixed

docs/content-services/services/category.service.md
MichalKinas: Right, fixed now

MichalKinas:
> 

Should be fixed now

MichalKinas:
> > Looks like ACA env is broken a bit for our epic but I found one issue which occurs only on your branch but on ACA works fine. If you do following actions like on video then you can't edit tags and categories anymore, even after clicking at Edit mode icon.
> 
> This one is not fixed still

Should be fixed now

AleksanderSklorz (approved):

AleksanderSklorz:
Sorry I noticed it just now - after the newest changes there is infinite spinner visible in Viewer for tags. It is visible without end: 
image

MichalKinas:
> @MichalKinas sorry I noticed it just now - after the newest changes there is infinite spinner visible in Viewer for tags. It is visible without end: image

Fixed now

nikita-web-ua (approved):

MateuszHY (approved):

DenysVuika (approved):

PR #8513, opened 2023-04-26
[ACA-4690] To limit width of Create Tag in properties while adding tag

arohilaGL:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

The width of the Create tag keeps increasing indefinitely.

Limiting the width and breaking the word so that the whole name is visible but the width isn't indefinite.

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/content-services/src/lib/tag/tags-creator/tags-creator.component.scss
+++ b/lib/content-services/src/lib/tag/tags-creator/tags-creator.component.scss
@@ -38,7 +38,8 @@ adf-tags-creator {
         display: inline-block;
         padding-right: 12px;
         overflow: auto;
-        max-height: 7vh;
+        width: 75%;
+        word-break: break-all;
     }
 
     .adf-tags-list {


DenysVuika (approved):

lib/content-services/src/lib/tag/tags-creator/tags-creator.component.scss
```diff
@@ -38,7 +38,8 @@ adf-tags-creator {
         display: inline-block;
         padding-right: 12px;
         overflow: auto;
-        max-height: 7vh;
+        width: 75%;
```
MichalKinas, line 41: I don't think this should be the solution for the error mentioned in the Jira, this should be fixed in ACA as the new layout caused the error. The problem has nothing to do with T&C which has been working in ACA and control-center before layout changes. Also this might break dialog for creating tags in control-center. @AleksanderSklorz do you agree?

MichalKinas (changes requested):

lib/content-services/src/lib/tag/tags-creator/tags-creator.component.scss
AleksanderSklorz: Yes I completely agree with you @MichalKinas

AleksanderSklorz: We have demoed tags and categories earlier in old UI and it was working fine so it has nothing related with Tags and Categories.

arohilaGL: I looked up again, found a class missing in ACA which had happened during layout changes, I have added the class, raised another PR for it. But still suggesting to remove max-height from adf-tags-creator as it is causing a second scroll when the tag name is large. You can remove it later. https://github.com/Alfresco/alfresco-content-app/pull/3150
image

AleksanderSklorz: I think this scrollbar is to avoid situation that somebody will type really very long tag name so user would need to scroll down to see existing tags. Not sure if for categories we have scrollbar here @MichalKinas? We can remove scrollbar if the same is for categories, that's fine for me. Please also check if that does not break tags modal in control center. 
In case we want to remove that scrollbar then only change in that PR would be removing max-height I guess

MichalKinas: I think Categories doesn't have it since it would be really rare case so removing max-height is fine for me, but for sure we shouldn't modify the width and check if nothing breaks in control-center

PR #8534, opened 2023-05-05
[ACS-5137] Fixed navigation between images

nikita-web-ua:

> - [X] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://alfresco.atlassian.net/browse/ACS-5137

Navigation between images has been fixed and works as expected.

> - [ ] Yes
> - [X] No

**Other information**:

### DIFF
--- a/lib/core/src/lib/viewer/components/img-viewer.component.spec.ts
+++ b/lib/core/src/lib/viewer/components/img-viewer.component.spec.ts
@@ -354,17 +354,21 @@ describe('Test Img viewer component ', () => {
             component.readOnly = false;
             component.isEditing = true;
 
+            const canvasMock = document.createElement("CANVAS");
             spyOn(component.isSaving, 'emit');
             spyOn(component, 'save').and.callThrough();
+            spyOn(component.cropper, 'getCroppedCanvas').and.returnValue(canvasMock as HTMLCanvasElement);
+            spyOn(component.cropper.getCroppedCanvas(), 'toBlob').and.callFake(() => component.isSaving.emit(false));
+
             fixture.detectChanges();
             const saveButtonElement = fixture.debugElement.query(By.css('#viewer-save-button'));
             saveButtonElement.triggerEventHandler('click', null);
             tick();
 
             expect(component.save).toHaveBeenCalled();
-            expect(component.isSaving.emit).toHaveBeenCalled();
+            expect(component.isSaving.emit).toHaveBeenCalledWith(true);
+            expect(component.isSaving.emit).toHaveBeenCalledWith(false);
         }));
-
     });
 
 });
--- a/lib/core/src/lib/viewer/components/img-viewer.component.spec.ts
+++ b/lib/core/src/lib/viewer/components/img-viewer.component.spec.ts
@@ -354,10 +354,10 @@ describe('Test Img viewer component ', () => {
             component.readOnly = false;
             component.isEditing = true;
 
-            const canvasMock = document.createElement("CANVAS");
+            const canvasMock = document.createElement('canvas');
             spyOn(component.isSaving, 'emit');
             spyOn(component, 'save').and.callThrough();
-            spyOn(component.cropper, 'getCroppedCanvas').and.returnValue(canvasMock as HTMLCanvasElement);
+            spyOn(component.cropper, 'getCroppedCanvas').and.returnValue(canvasMock);
             spyOn(component.cropper.getCroppedCanvas(), 'toBlob').and.callFake(() => component.isSaving.emit(false));
 
             fixture.detectChanges();
--- a/lib/core/src/lib/viewer/components/img-viewer.component.spec.ts
+++ b/lib/core/src/lib/viewer/components/img-viewer.component.spec.ts
@@ -354,13 +354,15 @@ describe('Test Img viewer component ', () => {
             component.readOnly = false;
             component.isEditing = true;
 
-            spyOn(component, 'save');
+            spyOn(component.isSaving, 'emit');
+            spyOn(component, 'save').and.callThrough();
             fixture.detectChanges();
             const saveButtonElement = fixture.debugElement.query(By.css('#viewer-save-button'));
             saveButtonElement.triggerEventHandler('click', null);
             tick();
 
             expect(component.save).toHaveBeenCalled();
+            expect(component.isSaving.emit).toHaveBeenCalled();
         }));
 
     });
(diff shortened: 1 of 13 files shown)


lib/core/src/lib/viewer/components/img-viewer.component.ts
```diff
@@ -143,6 +143,13 @@ export class ImgViewerComponent implements AfterViewInit, OnChanges, OnDestroy {
             this.urlFile = this.urlService.createTrustedUrl(this.blobFile);
             return;
         }
+
+        if (!changes['urlFile'].firstChange && changes['fileName']) {
```
AleksanderSklorz, line 147: Small thing but you can combine these two 'if' into single one using &&

AleksanderSklorz:
@nikita-web-ua I can sometimes still reproduce that bug. If I edit file then navigation does not work. Sometimes I need to edit it more than once to reproduce (sometimes it works when I edit one time, sometimes not and I need edit more times).


https://user-images.githubusercontent.com/115619721/236405513-f2edb480-7dfa-4d52-b73d-0482f7f70141.mov

nikita-web-ua:
> @nikita-web-ua I can sometimes still reproduce that bug. If I edit file then navigation does not work. Sometimes I need to edit it more than once to reproduce (sometimes it works when I edit one time, sometimes not and I need edit more times). Please also note that it even exits from view and breaks whole list of files and displays some error
> 
>  Screen.Recording.2023-05-05.at.09.56.45.mov

Fixed, this happened when the file was not saved before moving on to the next one. Navigation is now disabled while the file is being saved

lib/core/src/lib/viewer/components/img-viewer.component.spec.ts
```diff
@@ -354,13 +354,15 @@ describe('Test Img viewer component ', () => {
...
-            spyOn(component, 'save');
+            spyOn(component.isSaving, 'emit');
+            spyOn(component, 'save').and.callThrough();
             fixture.detectChanges();
             const saveButtonElement = fixture.debugElement.query(By.css('#viewer-save-button'));
             saveButtonElement.triggerEventHandler('click', null);
             tick();
 
             expect(component.save).toHaveBeenCalled();
+            expect(component.isSaving.emit).toHaveBeenCalled();
```
AleksanderSklorz, line 365: Can you also test if that emits true or false? Also can you add test for second case as you have two usages of that in component?

lib/core/src/lib/viewer/components/viewer-render.component.ts
```diff
@@ -175,6 +179,10 @@ export class ViewerRenderComponent implements OnChanges, OnInit, OnDestroy {
         }
     }
 
+    onSavingFile(isSaving: boolean) {
```
AleksanderSklorz, line 182: This function is redundant probably as it only wraps this.isSaving.emit which can be called directly in html like: 
(isSaving)="isSaving.emit($event)"

lib/core/src/lib/viewer/components/viewer.component.ts
```diff
@@ -344,6 +344,10 @@ export class ViewerComponent<T> implements OnDestroy, OnInit, OnChanges {
         this.submitFile.emit(newImageBlob);
     }
 
+    onSavingFile(isSaving: boolean) {
```
AleksanderSklorz, line 347: Also here you can do that assigning in html directly probably as that function has nothing more, like: 
(isSaving)="allowNavigate = !$event"

AleksanderSklorz:
@nikita-web-ua I found new bug. If you have new loaded docx file and new loaded images (this is very important to have new loaded docx and images in your library and you should not open that docx earlier). Then you can't navigate from image to docx actually. See video: 

https://github.com/Alfresco/alfresco-ng2-components/assets/115619721/30124aa7-7793-493c-80a5-e4aa0561e421

AleksanderSklorz:
> > @nikita-web-ua I can sometimes still reproduce that bug. If I edit file then navigation does not work. Sometimes I need to edit it more than once to reproduce (sometimes it works when I edit one time, sometimes not and I need edit more times). Please also note that it even exits from view and breaks whole list of files and displays some error
> > Screen.Recording.2023-05-05.at.09.56.45.mov
> 
> Fixed, this happened when the file was not saved before moving on to the next one. Navigation is now disabled while the file is being saved

Thank you for resolving this one

lib/core/src/lib/viewer/components/viewer.component.spec.ts
```diff
@@ -31,106 +31,22 @@ import {
...
+import { ViewerWithCustomToolbarActionsComponent } from './mock/adf-viewer-container-toolbar-actions.component.mock';
+import { Component } from '@angular/core';
+import { By } from '@angular/platform-browser';
 
 @Component({
     selector: 'adf-dialog-dummy',
     template: ``
 })
-class DummyDialogComponent {
+export class DummyDialogComponent {
```
AleksanderSklorz, line 47: Any reason why you need export this?

AleksanderSklorz:
> @nikita-web-ua I found new bug. If you have new loaded docx file and new loaded images (this is very important to have new loaded docx and images in your library and you should not open that docx earlier). Then you can't navigate from image to docx actually. See video:
> 
>  Screen.Recording.2023-05-10.at.08.53.05.mov

Regarding to this issue - we decided to raise separate jira for this.

lib/core/src/lib/viewer/components/img-viewer.component.ts
```diff
@@ -172,13 +183,15 @@ export class ImgViewerComponent implements AfterViewInit, OnChanges, OnDestroy {
     }
 
     save() {
+        this.isSaving.emit(true);
         this.isEditing = false;
         this.cropper.setDragMode('move');
-
         this.cropper.getCroppedCanvas().toBlob((blob) => {
             this.submit.emit(blob);
+            console.log('msdg')
```
AleksanderSklorz, line 191: Please remove console log

lib/core/src/lib/viewer/components/img-viewer.component.spec.ts
```diff
@@ -354,17 +354,21 @@ describe('Test Img viewer component ', () => {
             component.readOnly = false;
             component.isEditing = true;
 
+            const canvasMock = document.createElement("CANVAS");
             spyOn(component.isSaving, 'emit');
             spyOn(component, 'save').and.callThrough();
+            spyOn(component.cropper, 'getCroppedCanvas').and.returnValue(canvasMock as HTMLCanvasElement);
```
AleksanderSklorz, line 360: I think you don't need this 'as HTMLCanvasElement' here. The reason why canvasMock has incorrect type is because you typed incorrect name in document.createElement. Should be 'canvas' not 'CANVAS'. If you rename it to 'canvas' then you will get proper type, see below screenshot: 
image

AleksanderSklorz (approved):

MichalKinas (approved):

jatin2008 (approved):

DenysVuika (approved):

PR #8584, opened 2023-05-23
[ACS-5088] Added cast pipe to allow for typecasting in templates

swapnil-verma-gl:

> - [ ] Bugfix
> - [x] Feature
> - [ ] Build related changes
> - [ ] Documentation

There is no pipe to allow casting variables on templates

Added a new pipe, to allow casting a variable to a different type on html

> - [ ] Yes
> - [x] No

**Other information**:
Added a 'type-cast.pipe.ts' that allows casting of variables from one type to another. This is useful if an input data property is of different type than what is available in the component.

Usage notes - In order to use it, create a reference (preferably readonly) of the constructor for the data type to which to cast to, in the component.ts file.

`protected readonly NodeEntity = NodeEntity`

You should then be able to use this reference in the template to cast a variable to this type. 

`<my-comp [inputProp]="someVariable | cast : NodeEntity"></my-comp>`

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,4 +1,3 @@
-
 ---
 Title: Component docs
 Github only: true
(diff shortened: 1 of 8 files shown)


docs/README.md
```diff
@@ -1,3 +1,4 @@
+
```
AleksanderSklorz, line 1: Do we need this empty line? It probably causes that change. Remove if not needed.  
image

lib/core/src/lib/pipes/type-cast.pipe.ts
```diff
@@ -0,0 +1,30 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { Pipe, PipeTransform } from '@angular/core';
+
+@Pipe({
+    name: 'cast',
+    pure: true
```
AleksanderSklorz, line 22: Pipes are pure by default, redundant: 
image

docs/README.md
swapnil-verma-gl: Removed unneeded spaces.

lib/core/src/lib/pipes/type-cast.pipe.ts
swapnil-verma-gl: Removed pure: true config from type-cast.pipe.ts

lib/core/src/lib/pipes/type-cast.pipe.ts
```diff
@@ -0,0 +1,30 @@
...
+})
+export class TypeCastPipe implements PipeTransform {
+
+    transform<T>(value: any, _type: (new (...args: any[]) => T) | T): T {
+        return value as T;
```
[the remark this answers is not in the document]
DenysVuika: I don't think it's a correct solution.. you are essentially hacking the Typescript to stop raising issues on types. Instead, the components should have proper types

docs/core/pipes/type-cast.pipe.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Basic Usage
+
+<!-- {% raw %} -->
+```TS
+    protected readonly MyCustomTypeConstructor = MyCustomTypeConstructor
+```
+```HTML
+<div>
+    <my-comp [inputProp]="someVar | cast : MyCustomTypeConstructor"></my-comp>
```
DenysVuika, line 20: this does not fix the issue of the component not using proper typing in the properties. Angular already provides "$any" cast for cases when one needs to suppress the warning:
```
[node]="$any(site)"
```

DenysVuika (changes requested):

docs/core/pipes/type-cast.pipe.md
swapnil-verma-gl: The function call of $any() is the reason why we created this pipe. The benefit of this pipe is that it allows us to assert an object to a specifc type, without requiring the use of the $any function in the template. 

If we are fine with using $any(), then we don't need this new pipe at all.

PR #8597, opened 2023-05-26
[AAE-14699] Improved card view text item component update

ehsan-2019:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

Currently we emit update event of card view text item only when the value is valid. in order to better management of show/hide error message and enable/disable save buttons on apps we need to receive recent value regardless of it being valid/invalid.  

> - [ ] Yes
> - [X] No

**Other information**:

### DIFF
--- a/lib/core/src/lib/card-view/components/card-view-textitem/card-view-textitem.component.spec.ts
+++ b/lib/core/src/lib/card-view/components/card-view-textitem/card-view-textitem.component.spec.ts
@@ -509,11 +509,23 @@ describe('CardViewTextItemComponent', () => {
             expect(cardViewUpdateService.update).toHaveBeenCalledWith(property, 'updated-value');
         });
 
-        it('should trigger the update event if the editedValue', async () => {
+        it('should trigger the update event if the editedValue is NOT invalid', async () => {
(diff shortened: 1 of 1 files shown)


lib/core/src/lib/card-view/components/card-view-textitem/card-view-textitem.component.spec.ts
```diff
@@ -509,15 +509,14 @@ describe('CardViewTextItemComponent', () => {
...
             const cardViewUpdateService = TestBed.inject(CardViewUpdateService);
             spyOn(cardViewUpdateService, 'update');
-            component.property.isValid = () => false;
 
             updateTextField(component.property.key, 'updated-value');
             await fixture.whenStable();
 
-            expect(cardViewUpdateService.update).not.toHaveBeenCalled();
+            expect(cardViewUpdateService.update).toHaveBeenCalled();
         });
```
DudaRobert, line 520: Maybe there should be two tests:
- was `update` called when `property.isValid` returned `true`,
- was `update` called when `property.isValid` returned `false`.

Just to check for false positives as well.

DudaRobert (approved):

PR #8611, opened 2023-05-30
[AAE-14469] added constraint for typed value

tomaszhanaj:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://alfresco.atlassian.net/browse/AAE-14469

We can provide minimum number of characters for typed value to run the search.

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/process-services-cloud/src/lib/people/components/people-cloud.component.ts
+++ b/lib/process-services-cloud/src/lib/people/components/people-cloud.component.ts
@@ -123,9 +123,6 @@ export class PeopleCloudComponent implements OnInit, OnChanges, OnDestroy {
     @Input()
     title: string;
 
-    @Input()
-    minNrOfCharacters?: number;
-
     /** Emitted when a user is selected. */
     @Output()
     selectUser = new EventEmitter<IdentityUserModel>();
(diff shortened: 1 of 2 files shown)


lib/process-services-cloud/src/lib/people/components/people-cloud.component.ts
```diff
@@ -123,6 +123,9 @@ export class PeopleCloudComponent implements OnInit, OnChanges, OnDestroy {
     @Input()
     title: string;
 
+    @Input()
+    minNrOfCharacters?: number;
```
VitoAlbano, line 127: Why do we want this check configurable? i mean we want to avoid making the calls for empty string that is something normal, is there a scenario where this makes sense?

VitoAlbano (changes requested):

lib/process-services-cloud/src/lib/people/components/people-cloud.component.ts
tomaszhanaj: This component is used in many places and I was not sure if it is not going to break anything. So this is why I made this optional.

tomaszhanaj: removed optional property

VitoAlbano (approved):

PR #8766, opened 2023-07-18
[ACS-5183] properties facet file size and file type

AleksanderSklorz:

> - [ ] Bugfix
> - [x] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://alfresco.atlassian.net/browse/ACS-5183

Added properties facet which allows search by file size and type. 

> - [ ] Yes
> - [x] No

**Other information**:
ACA PR https://github.com/Alfresco/alfresco-content-app/pull/3347


lib/content-services/src/lib/search/components/search-properties/search-properties.ts
```diff
@@ -0,0 +1,6 @@
+import { FileSizeCondition } from './file-size-condition';
```
DenysVuika, line 1: missing license header

AleksanderSklorz: done

DenysVuika (approved):

lib/content-services/src/lib/search/components/search-properties/search-properties.component.ts
```diff
@@ -0,0 +1,216 @@
...
+        };
+    }
+
+    setValue(searchProperties: SearchProperties) {
+        this.form.patchValue(searchProperties.fileSizeCondition);
+        this.selectedExtensions = searchProperties.fileExtensions;
+        this.submitValues();
+    }
+
+    private getOperatorNameWidth(operator: string, font: string): number {
```
MichalKinas, line 203: Might be obsolete if we decide on setting `max-width`, same for other methods lower

lib/content-services/src/lib/search/components/search-properties/search-properties.component.ts
```diff
@@ -0,0 +1,216 @@
...
+                this.settings.fileExtensions = [];
+            }
+            [this.sizeField, this.nameField] = this.settings.field.split(',');
+        }
+        if (this.startValue) {
+            this.setValue(this.startValue);
+        }
+    }
+
+    ngAfterViewChecked() {
```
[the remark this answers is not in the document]
AleksanderSklorz: Unfortunately that is not possible because we can have different languages etc. Only way to do it properly and working for each language is calculate width of options and take the highest one. 
Solution which you suggested would work in case where we have only one language but that is not a case here.

lib/content-services/src/lib/search/components/search-properties/search-properties.component.ts
```diff
@@ -0,0 +1,216 @@
...
+                const extraFreeSpace = 20;
+                this._fileSizeOperatorsMaxWidth = Math.max(...this._fileSizeOperators.map((operator) =>
+                    this.getOperatorNameWidth(operator, this.getCanvasFont(this.fileSizeOperatorSelectElement.nativeElement)))) +
+                    this.fileSizeOperatorSelectElement.nativeElement.querySelector('.mat-select-arrow-wrapper').clientWidth +
+                    extraFreeSpace;
+            });
+        }
+    }
+
+    narrowDownAllowedCharacters(event: Event) {
```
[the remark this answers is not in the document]
AleksanderSklorz: it is the same approach like for created/modified dates:

https://github.com/Alfresco/alfresco-ng2-components/pull/8714/files#diff-3d9b6561e3d273c811ef01fd21125aa8beafdbf53e18815843346c45a40d27b1R167

Basically pattern would mark this as invalid but we don't want to even allow to type incorrect character because we don't want to display error message at all. We just want to prevent typing incorrect characters. Pattern will allow type incorrect characters and just mark field as invalid

lib/content-services/src/lib/search/components/search-properties/search-properties.component.ts
```diff
@@ -0,0 +1,216 @@
...
+        }
+    }
+
+    clearNumberFieldWhenInvalid(event: FocusEvent) {
+        if (!(event.target as HTMLInputElement).validity.valid) {
+            this.form.controls.fileSize.setValue(undefined);
+        }
+    }
+
+    preventIncorrectNumberCharacters(event: KeyboardEvent): boolean {
```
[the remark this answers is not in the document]
AleksanderSklorz: see answer above

lib/content-services/src/lib/search/components/search-properties/search-properties.component.ts
AleksanderSklorz: Do you have any other suggestion for naming? 
That function takes some text (operator name) like 'At Least' and calculates actual width which that text takes. 
Text has not 'max-width', text has just some width which it takes.

MichalKinas (approved):

lib/content-services/src/lib/search/components/search-properties/search-properties.component.ts
MichalKinas: No no no, the comment was about potential removal of those method if we wanted to set `max-width` instead of calculating the width but we're doing this now so you can ignore this comment

PR #8771, opened 2023-07-20
[AAE-12065] Add isDefault flag to dropdown

DiogoABastos:

> - [ ] Bugfix
> - [x] Feature
> - [ ] Build related changes
> - [ ] Documentation

If dropdown field had the same id as the `DEFAULT_OPTION.id`, it would be taken as the default option

https://alfresco.atlassian.net/browse/AAE-12065
https://github.com/Alfresco/alfresco-apps/pull/5894

Add `isDefault` flag to allow users to use the `DEFAULT_OPTION.id` as an id, without making such an option the default option
Is compatible with JSON forms previously created

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/core/src/lib/form/components/widgets/core/form-field-option.ts
+++ b/lib/core/src/lib/form/components/widgets/core/form-field-option.ts
@@ -20,4 +20,5 @@
 export interface FormFieldOption {
     id: string;
     name: string;
+    isDefault?: boolean;
 }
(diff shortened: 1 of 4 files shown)


eromano (changes requested):
what happens if isDefault is not present in the JSON because is a form previously created?

DiogoABastos:
> what happens if isDefault is not present in the JSON because is a form previously created?

`isDefault` flag is defaulted to false. The problem is that when the dropdown has the default option enabled, it will be regarded as a normal option because of the `isDefault`'s default value. I'll make some changes to account for that

eromano (changes requested):
can we add a unit that cover this scenario?

DiogoABastos:
No need for this changes anymore. https://alfresco.atlassian.net/browse/AAE-12065?focusedCommentId=1192290

PR #8775, opened 2023-07-21
[ACS-5179] Add search facet tabbed component for creator and modifier

MichalKinas:

> - [ ] Bugfix
> - [x] Feature
> - [ ] Build related changes
> - [ ] Documentation

New tabbed facet component for creator and modifier has been added. 
search-facet-chip-tabbed

> - [ ] Yes
> - [x] No

**Other information**:


nikita-web-ua:
Guess versionIndex.md should be updated

lib/content-services/src/lib/search/components/search-filter-chips/search-facet-chip-tabbed/search-facet-chip-tabbed.component.html
```diff
@@ -0,0 +1,60 @@
...
+                                     (close)="menuTrigger.closeMenu()">
+            <ng-container ngProjectAs="filter-title">
+                {{ tabbedFacet.label | translate }}
+            </ng-container>
+            <ng-container ngProjectAs="filter-content">
+                <adf-search-filter-tabbed>
+                    <ng-container *ngFor="let field of tabbedFacet.fields">
+                        <adf-search-chip-autocomplete-input
+                            *adf-search-filter-tab="tabbedFacet.facets[field]?.label"
+                            [autocompleteOptions]="autocompleteOptions[field]"
```
nikita-web-ua, line 42: We should provide [compareOption] to `adf-search-chip-autocomplete-input` so that selected options would be greyed out and disabled in autocomplete. 

Something like:  

optionComparator(option1: AutocompleteOption, option2: AutocompleteOption): boolean => option1.value.toUpperCase() === option2.value.toUpperCase();

lib/content-services/src/lib/search/components/search-filter-chips/search-facet-chip-tabbed/search-facet-chip-tabbed.component.spec.ts
```diff
@@ -0,0 +1,232 @@
...
+    });
+
+    it('should display remove icon and disable facet when no items are loaded',  () => {
+        const chip = fixture.debugElement.query(By.css('mat-chip'));
+        const icon = fixture.debugElement.query(By.css('mat-chip mat-icon')).nativeElement.innerText;
+        expect(chip.classes['mat-chip-disabled']).toBeTrue();
+        expect(icon).toEqual('remove');
+    });
+
+    it('should not open context menu when no items are loaded',  () => {
```
nikita-web-ua, line 112: double space before ()

lib/content-services/src/lib/search/components/search-filter-chips/search-facet-chip-tabbed/search-facet-chip-tabbed.component.spec.ts
```diff
@@ -0,0 +1,232 @@
...
+        const label = fixture.debugElement.query(By.css('.adf-search-filter-ellipsis')).nativeElement.innerText;
+        expect(label).toBe(component.tabbedFacet.label + ':');
+    });
+
+    it('should display any as display value when nothing is selected', () => {
+        const displayValue = getDisplayValue();
+        expect(displayValue).toBe('SEARCH.FILTER.ANY');
+    });
+
+    it('should display remove icon and disable facet when no items are loaded',  () => {
```
nikita-web-ua, line 105: double space before ()

lib/content-services/src/lib/search/components/search-filter-chips/search-facet-chip-tabbed/search-facet-chip-tabbed.component.spec.ts
```diff
@@ -0,0 +1,232 @@
...
+        openFacet();
+        let activeTabLabel = fixture.debugElement.query(By.css('.mat-tab-label-active .mat-tab-label-content')).nativeElement.innerText;
+        expect(activeTabLabel).toBe(component.tabbedFacet.facets['field'].label);
+
+        changeTab(1);
+        activeTabLabel = fixture.debugElement.query(By.css('.mat-tab-label-active .mat-tab-label-content')).nativeElement.innerText;
+        expect(activeTabLabel).toBe(component.tabbedFacet.facets['field2'].label);
+    });
+
+    it('should display arrow down icon and not disable the chip when items are loaded',  () => {
```
nikita-web-ua, line 143: here also double space and below as well check please

lib/content-services/src/lib/search/components/search-filter-chips/search-facet-chip-tabbed/search-facet-chip-tabbed.component.ts
```diff
@@ -0,0 +1,173 @@
...
+
+    @ViewChild('menuContainer', { static: false })
+    menuContainer: ElementRef;
+
+    @ViewChild('menuTrigger', { static: false })
+    menuTrigger: MatMenuTrigger;
+
+    private resetSubject$ = new Subject<void>();
+
+    displayValue$: Subject<string> = new Subject<string>();
```
nikita-web-ua, line 48: left hand type is redundant

lib/content-services/src/lib/search/components/search-filter-chips/search-facet-chip-tabbed/search-facet-chip-tabbed.component.ts
```diff
@@ -0,0 +1,173 @@
...
+    reset$: Observable<void> = this.resetSubject$.asObservable();
+    focusTrap: ConfigurableFocusTrap;
+    chipIcon = 'keyboard_arrow_down';
+    autocompleteOptions = {};
+    selectedOptions = {};
```
nikita-web-ua, line 53:
```suggestion
    selectedOptions: AutocompleteOption[] = [];
```

lib/content-services/src/lib/search/components/search-filter-chips/search-facet-chip-tabbed/search-facet-chip-tabbed.component.ts
```diff
@@ -0,0 +1,173 @@
...
+    @ViewChild('menuTrigger', { static: false })
+    menuTrigger: MatMenuTrigger;
+
+    private resetSubject$ = new Subject<void>();
+
+    displayValue$: Subject<string> = new Subject<string>();
+    reset$: Observable<void> = this.resetSubject$.asObservable();
+    focusTrap: ConfigurableFocusTrap;
+    chipIcon = 'keyboard_arrow_down';
+    autocompleteOptions = {};
```
nikita-web-ua, line 52:
```suggestion
    autocompleteOptions: AutocompleteOption[] = [];
```

lib/content-services/src/lib/search/components/search-filter-chips/search-facet-chip-tabbed/search-facet-chip-tabbed.component.ts
```diff
@@ -0,0 +1,173 @@
...
+        this.reset();
+        this.menuTrigger.closeMenu();
+    }
+
+    onApply() {
+        this.submitValues();
+        this.menuTrigger.closeMenu();
+    }
+
+    onEnterKeydown(): void {
```
nikita-web-ua, line 107: For some methods you specify `void` type but for the rest you omit, maybe worth to omit for all

lib/content-services/src/lib/search/components/search-facet-field/search-facet-field.component.spec.ts
```diff
@@ -62,7 +62,7 @@ describe('SearchFacetFieldComponent', () => {
             component.onToggleBucket(event, field, bucket);
 
             expect(bucket.checked).toBeTruthy();
-            expect(queryBuilder.addUserFacetBucket).toHaveBeenCalledWith(field, bucket);
+            expect(queryBuilder.addUserFacetBucket).toHaveBeenCalledWith(field.field, bucket);
```
swapnil-verma-gl, line 65: Can we maybe rename the unit test variable with something else? field.field sounds weird

lib/content-services/src/lib/search/components/search-facet-field/search-facet-field.component.ts
```diff
@@ -61,7 +61,7 @@ export class SearchFacetFieldComponent implements FacetWidget {
     selectFacetBucket(field: FacetField, bucket: FacetFieldBucket) {
         if (bucket) {
             bucket.checked = true;
-            this.queryBuilder.addUserFacetBucket(field, bucket);
+            this.queryBuilder.addUserFacetBucket(field.field, bucket);
```
swapnil-verma-gl, line 64: Same as unit tests, field.field sounds weird

lib/content-services/src/lib/search/components/search-filter-chips/search-facet-chip-tabbed/search-facet-chip-tabbed.component.spec.ts
```diff
@@ -0,0 +1,232 @@
...
+        openFacet();
+        const tabLabels = getTabs();
+        expect(tabLabels.length).toBe(2);
+        expect(tabLabels[0].innerText).toBe(component.tabbedFacet.facets['field'].label);
+        expect(tabLabels[1].innerText).toBe(component.tabbedFacet.facets['field2'].label);
+    });
+
+    it('should display creator tab as active initially and allow navigation', () => {
+        openFacet();
+        let activeTabLabel = fixture.debugElement.query(By.css('.mat-tab-label-active .mat-tab-label-content')).nativeElement.innerText;
```
swapnil-verma-gl, line 135: Maybe add a null safety check here on nativeElement. Something like 

```
let activeTabLabel = fixture.debugElement.query(By.css('.mat-tab-label-active .mat-tab-label-content')).nativeElement?.innerText;
```

lib/content-services/src/lib/search/components/search-filter-chips/search-facet-chip-tabbed/search-facet-chip-tabbed.component.spec.ts
```diff
@@ -0,0 +1,232 @@
...
+        openFacet();
+        const applyButton = fixture.debugElement.query(By.css('#apply-filter-button'));
+        applyButton.triggerEventHandler('click', {});
+        expect(queryBuilder.update).toHaveBeenCalled();
+        expect(component.menuTrigger.closeMenu).toHaveBeenCalled();
+        expect(component.updateDisplayValue).toHaveBeenCalled();
+        expect(searchFacetService.updateSelectedBuckets).toHaveBeenCalled();
+    });
+
+  it('should update search query and display value when cancel btn is clicked',  () => {
```
swapnil-verma-gl, line 222: Indentation not correct here it seems

docs/content-services/components/search-facet-chip-tabbed.md
```diff
@@ -0,0 +1,49 @@
+---
+Title: Search facet chip tabbed component
+Added: v6.2.0
+Status: Active
+Last reviewed: 2023-07-18
+---
+
+# [Search facet chip tabbed component](../../../lib/content-services/src/lib/search/components/search-filter-chips/search-facet-chip-tabbed/search-facet-chip-tabbed.component.ts "Defined in search-facet-chip-tabbed.component.ts")
+
+Implements a [facet widget](../../../lib/content-services/src/lib/search/models/facet-widget.interface.ts) consisting of creator and modifier facets insided tabbed component.
```
AleksanderSklorz, line 10: probably typo - should be inside

lib/content-services/src/lib/search/components/search-filter-chips/search-facet-chip-tabbed/search-facet-chip-tabbed.component.ts
```diff
@@ -0,0 +1,173 @@
...
+    private resetSubject$ = new Subject<void>();
+
+    displayValue$: Subject<string> = new Subject<string>();
+    reset$: Observable<void> = this.resetSubject$.asObservable();
+    focusTrap: ConfigurableFocusTrap;
+    chipIcon = 'keyboard_arrow_down';
+    autocompleteOptions = {};
+    selectedOptions = {};
+
+    constructor(@Inject(SEARCH_QUERY_SERVICE_TOKEN) public queryBuilder: SearchQueryBuilderService,
```
swapnil-verma-gl, line 55: Do we need the queryBuilder to be public? Or can we set it as private. I don't see any usages of this components queryBuilder object outside of this class

lib/content-services/src/lib/search/components/search-chip-autocomplete-input/search-chip-autocomplete-input.component.ts
```diff
@@ -105,7 +105,7 @@ export class SearchChipAutocompleteInputComponent implements OnInit, OnDestroy,
 
     ngOnChanges(changes: SimpleChanges) {
         if (changes.autocompleteOptions) {
-            this.filteredOptions = changes.autocompleteOptions.currentValue.length > 0 ? this.filter(changes.autocompleteOptions.currentValue, this.formCtrl.value) : [];
+            this.filteredOptions = changes.autocompleteOptions.currentValue?.length > 0 ? this.filter(changes.autocompleteOptions.currentValue, this.formCtrl.value) : [];
```
AleksanderSklorz, line 108: Just from curiosity. Did you add '?' because you saw following error in console? 
image

If yes then I had similar error and I had discussion with @nikita-web-ua and it was probably because of some missing configuration from Mykyta's jira. If that was a reason please check with @nikita-web-ua if this really needs to be fixed.

lib/content-services/src/lib/search/components/search-filter-chips/search-facet-chip-tabbed/search-facet-chip-tabbed.component.html
```diff
@@ -0,0 +1,60 @@
+<mat-chip [attr.data-automation-id]="'search-filter-chip-tabbed-' + tabbedFacet.label"
+          disableRipple
+          class="adf-search-filter-chip-tabbed"
+          [class.adf-search-toggle-chip]="(displayValue$ | async) || menuTrigger.menuOpen"
+          [disabled]="!isPopulated()"
```
AleksanderSklorz, line 5: can we avoid using function's call in html?

lib/content-services/src/lib/search/components/search-filter-chips/search-facet-chip-tabbed/search-facet-chip-tabbed.component.html
```diff
@@ -0,0 +1,60 @@
...
+
+    <span class="adf-search-filter-placeholder">
+        <span class="adf-search-filter-ellipsis">{{ tabbedFacet.label | translate }}:</span>
+    </span>
+
+    <span class="adf-search-filter-ellipsis adf-filter-value" *ngIf="displayValue$ | async as displayValue; else showAny">
+        &nbsp; {{ displayValue | translate }}
+    </span>
+    <ng-template #showAny><span class="adf-search-filter-ellipsis adf-filter-value">&nbsp;{{ 'SEARCH.FILTER.ANY' | translate }}</span></ng-template>
+    <mat-icon *ngIf="isPopulated(); else disabledIcon">{{ chipIcon }}</mat-icon>
```
AleksanderSklorz, line 22: here as well - can we avoid call function?

lib/content-services/src/lib/search/components/search-filter-chips/search-facet-chip-tabbed/search-facet-chip-tabbed.component.html
```diff
@@ -0,0 +1,60 @@
...
+          [disabled]="!isPopulated()"
+          tabIndex="0"
+          [matMenuTriggerFor]="menu"
+          (onMenuOpen)="onMenuOpen()"
+          (keydown.enter)="onEnterKeydown()"
+          (keydown.escape)="onEscKeydown()"
+          [attr.title]="displayValue$ | async"
+          #menuTrigger="matMenuTrigger">
+
+    <span class="adf-search-filter-placeholder">
```
AleksanderSklorz, line 14: do we need two separate spans? Maybe we can combine it into single span?

lib/content-services/src/lib/search/components/search-filter-chips/search-facet-chip-tabbed/search-facet-chip-tabbed.component.html
```diff
@@ -0,0 +1,60 @@
...
+        <span class="adf-search-filter-ellipsis">{{ tabbedFacet.label | translate }}:</span>
```
AleksanderSklorz, line 15: I see that some of css classes in that file are probably unused? Do you need them or can we remove?

lib/content-services/src/lib/search/components/search-filter-chips/search-facet-chip-tabbed/search-facet-chip-tabbed.component.scss
```diff
@@ -0,0 +1,13 @@
+.adf-search-filter-chip-tabbed {
```
AleksanderSklorz, line 1: I would suggest to wrap these all styles in component's selector to avoid styles leaking

lib/content-services/src/lib/search/components/search-filter-chips/search-facet-chip-tabbed/search-facet-chip-tabbed.component.spec.ts
```diff
@@ -0,0 +1,232 @@
...
+    fixture = TestBed.createComponent(SearchFacetChipTabbedComponent);
+    component = fixture.componentInstance;
+    queryBuilder = TestBed.inject(SearchQueryBuilderService);
+    searchFacetService = TestBed.inject(SearchFacetFiltersService);
+    spyOn(queryBuilder, 'update').and.stub();
+
+    const facet1: FacetField = { type: 'field', label: 'field', field: 'field', buckets: new SearchFilterList() };
+    const facet2: FacetField = { type: 'field', label: 'field2', field: 'field2', buckets: new SearchFilterList() };
+
+    const tabbedFacet: TabbedFacetField = {
```
AleksanderSklorz, line 52: this is redundant because you directly assign it to component.tabbedFacet in next line, you can remove that const

lib/content-services/src/lib/search/components/search-filter-chips/search-facet-chip-tabbed/search-facet-chip-tabbed.component.spec.ts
```diff
@@ -0,0 +1,232 @@
...
+      });
+    fixture.detectChanges();
+  }
+
+  function addBucketItem(field: string, displayValue: string) {
+    component.tabbedFacet.facets[field].buckets.items.push({ count: 1, label: displayValue, display: displayValue, filterQuery: '' });
+    triggerComponentChanges();
+  }
+
+    it('should display correct label for tabbed facet', () => {
```
AleksanderSklorz, line 95: Please align all test cases, they seem not aligned looking at above functions. Functions are closed to left side

lib/content-services/src/lib/search/components/search-filter-chips/search-filter-chips.component.html
```diff
@@ -3,6 +3,12 @@
         <adf-search-widget-chip [category]="category"></adf-search-widget-chip>
     </ng-container>
 
+    <ng-container *ngIf="facetFiltersService.tabbedFacet && showContextFacets">
+        <adf-search-facet-chip-tabbed
+            [tabbedFacet]="facetFiltersService.tabbedFacet"
+            [attr.data-automation-id]="'search-fact-chip-tabbed' + facetFiltersService.tabbedFacet.fields.concat('')" >
```
swapnil-verma-gl, line 9: Maybe we should add hypens (-) here between 'search-facet-chip-tabbed' and the fields,  as well as between the fields themselves. That way we would have an automation id like 
` search-facet-chip-tabbed-created-modified `
instead of 
` search-facet-chip-tabbedcreatedmodified `

Having said that, adding hyphens between the fields can be tricky as we don't want to end up with a trailing hyphen like 
` search-facet-chip-tabbed-created-modified- `

AleksanderSklorz, line 9: can we avoid calling functions in html? (concat)

lib/content-services/src/lib/search/services/search-facet-filters.service.ts
```diff
@@ -134,6 +138,34 @@ export class SearchFacetFiltersService implements OnDestroy {
         });
     }
 
+    private parseTabbedFacetField() {
+        if (this.responseFacets) {
+            let fields = '';
+            this.responseFacets.forEach((facet) => fields = `${fields},${facet.field}`);
```
AleksanderSklorz, line 144: maybe better would be reduce as you convert array to single value?

lib/content-services/src/lib/search/services/search-facet-filters.service.ts
```diff
@@ -361,7 +392,7 @@ export class SearchFacetFiltersService implements OnDestroy {
     unselectFacetBucket(field: FacetField, bucket: FacetFieldBucket) {
         if (bucket) {
             bucket.checked = false;
-            this.queryBuilder.removeUserFacetBucket(field, bucket);
+            this.queryBuilder.removeUserFacetBucket(field.field, bucket);
```
swapnil-verma-gl, line 395: Same here, field.field does not seem correct

AleksanderSklorz:
Search is applied without clicking on Apply button (it is enough to just select any user). I think for all other places we need to click on Apply button so it seems a bit inconsistent. 

https://github.com/Alfresco/alfresco-ng2-components/assets/115619721/9b95a31b-2b9a-4896-9bfb-4ea19f1266dc

AleksanderSklorz:
When I click on Clear button then it applies search as well. It does not work in that way for other facets/filters like properties. 

https://github.com/Alfresco/alfresco-ng2-components/assets/115619721/ecf88b87-e91e-49a5-baaf-adef599c581f

AleksanderSklorz:
@MichalKinas for what case we need Apply button if all changes trigger search automatically?

AleksanderSklorz:
@MichalKinas above observations also make problem that when you select one created by user like Administrator then you can't select different one because list is refreshed so you have no more users to select and you can select only 1 user

AleksanderSklorz:
@MichalKinas I also have following doubts: 
- should we really combine Created and Modified with AND? If in Created you have A and B users and then in modified B then A is completely redundant for Created... 
- it would be easier to allow OR but then I see also that issue: in Modified you will have only users from Created based on search results so you can't have A, B for created OR C for modified because C does not exist in created...

lib/content-services/src/lib/search/components/search-filter-chips/search-facet-chip-tabbed/search-facet-chip-tabbed.component.ts
MichalKinas: Actually it's not that easy as it seems, `selectedOptions` is a object that holds options for both `creator` and `modifier`. If you check in `ngOnChanges` you can see that options assigned for each field are created with appropriate `AutocompleteOption[]` type

lib/content-services/src/lib/search/components/search-filter-chips/search-facet-chip-tabbed/search-facet-chip-tabbed.component.ts
MichalKinas: Please check my response above

lib/content-services/src/lib/search/components/search-chip-autocomplete-input/search-chip-autocomplete-input.component.ts
MichalKinas: Yeah I think this was the reason for the change, @nikita-web-ua do I need to add anything additional for this to work?

nikita-web-ua: @MichalKinas, @AleksanderSklorz I checked with latest ACA and I do not see an error, have you checked with latest ACA?

MichalKinas: Will check

MichalKinas: Okay it works thanks!

lib/content-services/src/lib/search/components/search-filter-chips/search-facet-chip-tabbed/search-facet-chip-tabbed.component.html
MichalKinas: All of them are in use, some styles are coming from global search styles that are defined

lib/content-services/src/lib/search/components/search-filter-chips/search-filter-chips.component.html
MichalKinas: I thought we want to avoid function calls inside the html templates and we can still use them for attributes but maybe I'm wrong, could you please confirm?

AleksanderSklorz: Only exception is using them for events like click etc. because we can't avoid binding functions to event and that is fine because then these functions are executed only when event is triggered. 

All other usages like inside {{}}, linking to properties, ngIf etc. are disallowed because they are recalculated with every change detection which affects on performance.

AleksanderSklorz: @swapnil-verma-gl can you confirm too? You were working on replacing functions in html

MichalKinas: Okay makes sense, I will move it to .ts part of component

swapnil-verma-gl: @AleksanderSklorz  You are right, function calls in templates should only be reserved for event handlers. All input properties should have a direct variable reference in order to maximise performance. Using functions will cause them to be executed in every change detection cycle. 

Having said that, we did eventually decide that if we come across situations where replacing a function call with a variable proves to be too complex to implement and understand, then in those cases we can leave them in. Angular itself has a few areas where it uses function calls for attributes (Most common that I could find was with forms, where a lot of  functionality is hidden behind getters. For e.g., form.valid).

AleksanderSklorz: @swapnil-verma-gl thank you for confirmation

lib/content-services/src/lib/search/components/search-chip-autocomplete-input/search-chip-autocomplete-input.component.ts
MichalKinas: @AleksanderSklorz @nikita-web-ua guys actually without this null check my unit tests are failing since I'm populating autocompleteOptions in `ngOnChanges` and they're undefined before, would it be okay for you to leave it there?

AleksanderSklorz: it's small thing so should not be a problem for me

lib/content-services/src/lib/search/components/search-filter-chips/search-facet-chip-tabbed/search-facet-chip-tabbed.component.html
MichalKinas: Added

lib/content-services/src/lib/search/components/search-filter-chips/search-facet-chip-tabbed/search-facet-chip-tabbed.component.spec.ts
MichalKinas: Removed

lib/content-services/src/lib/search/components/search-filter-chips/search-facet-chip-tabbed/search-facet-chip-tabbed.component.spec.ts
MichalKinas: Removed

lib/content-services/src/lib/search/components/search-filter-chips/search-facet-chip-tabbed/search-facet-chip-tabbed.component.spec.ts
MichalKinas: Removed

lib/content-services/src/lib/search/components/search-filter-chips/search-facet-chip-tabbed/search-facet-chip-tabbed.component.ts
MichalKinas: Removed

lib/content-services/src/lib/search/components/search-filter-chips/search-facet-chip-tabbed/search-facet-chip-tabbed.component.ts
MichalKinas: Removed

lib/content-services/src/lib/search/components/search-facet-field/search-facet-field.component.spec.ts
MichalKinas: Renamed

lib/content-services/src/lib/search/components/search-facet-field/search-facet-field.component.ts
MichalKinas: Renamed

lib/content-services/src/lib/search/components/search-filter-chips/search-facet-chip-tabbed/search-facet-chip-tabbed.component.spec.ts
MichalKinas: Fixed

docs/content-services/components/search-facet-chip-tabbed.md
MichalKinas: Fixed

lib/content-services/src/lib/search/components/search-filter-chips/search-facet-chip-tabbed/search-facet-chip-tabbed.component.ts
MichalKinas: Good call, made it private

lib/content-services/src/lib/search/components/search-filter-chips/search-facet-chip-tabbed/search-facet-chip-tabbed.component.html
MichalKinas: Fixed

lib/content-services/src/lib/search/components/search-filter-chips/search-facet-chip-tabbed/search-facet-chip-tabbed.component.html
MichalKinas: Fixed

lib/content-services/src/lib/search/components/search-filter-chips/search-facet-chip-tabbed/search-facet-chip-tabbed.component.html
MichalKinas: Merged into one span

lib/content-services/src/lib/search/components/search-filter-chips/search-facet-chip-tabbed/search-facet-chip-tabbed.component.scss
MichalKinas: Wrapped

lib/content-services/src/lib/search/components/search-filter-chips/search-facet-chip-tabbed/search-facet-chip-tabbed.component.spec.ts
MichalKinas: Right, changed

lib/content-services/src/lib/search/components/search-filter-chips/search-facet-chip-tabbed/search-facet-chip-tabbed.component.spec.ts
MichalKinas: Aligned

lib/content-services/src/lib/search/components/search-filter-chips/search-facet-chip-tabbed/search-facet-chip-tabbed.component.ts
```diff
@@ -0,0 +1,173 @@
...
+                value: [],
+                writable: true
+            });
+        });
+    }
+
+    ngOnChanges(changes: SimpleChanges): void {
+        if (changes.tabbedFacet) {
+            this.tabbedFacet.fields.forEach((field) => {
+                const options: AutocompleteOption[] = this.tabbedFacet.facets[field].buckets.items.map((item) => {
```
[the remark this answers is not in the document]
MichalKinas: Changed

lib/content-services/src/lib/search/components/search-filter-chips/search-filter-chips.component.html
MichalKinas: Fixed

lib/content-services/src/lib/search/services/search-facet-filters.service.ts
MichalKinas: Changed to reduce

lib/content-services/src/lib/search/services/search-facet-filters.service.ts
MichalKinas: Changed

AleksanderSklorz:
All comments regarding to functionality we discussed on slack

lib/content-services/src/lib/search/components/search-filter-chips/search-filter-chips.component.ts
```diff
@@ -36,4 +36,7 @@ export class SearchFilterChipsComponent {
         public queryBuilder: SearchQueryBuilderService,
         public facetFiltersService: SearchFacetFiltersService) {}
 
+    get facetChipTabbedId() {
```
AleksanderSklorz, line 39: @MichalKinas this kind of getters are also disallowed. If you call function in getter and then getter is called during change detection in html then that function will be called too like it would be directly in HTML (you use join function here). 

We can use only simple getters without extra functions and too complex logic. Can we avoid that somehow?

AleksanderSklorz (approved):

lib/content-services/src/lib/search/components/search-filter-chips/search-filter-chips.component.ts
MichalKinas: I used different approach setting the id when search results gets updated, please check if that works for you

AleksanderSklorz: thanks, now it looks correct :)

lib/content-services/src/lib/search/components/search-facet-field/search-facet-field.component.ts
swapnil-verma-gl: I still see instances of field.field in this file, along with some others in 

- search-filter-chips.component.html
- search-query-builder.service.spect.ts

These 2 aren't introduced by your changes though. These are pre-existing

PR #8787, opened 2023-07-25
[ACS-5703] Comment List code and styles cleanup

DenysVuika:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

- cleanup the code and styles for the Comment List
- remove technical debts
- (security) remove unsafe "innerHTML" binding, remove the need for custom sanitize due to Angular protections
- drop selection animations and selected item (the component never had any selection actions besides storing the selection internally, can be re-introduced in the future once any actions are supported like editing, deletion, etc)
- move "multiline" test from e2e to unit

> - [ ] Yes
> - [x] No

**Other information**:
https://alfresco.atlassian.net/browse/ACS-5703


lib/core/src/lib/comments/comment-list/comment-list.component.html
```diff
@@ -2,29 +2,22 @@
...
         <div class="adf-comment-contents">
-            <div matLine [attr.id]="'comment-user-' + comment.id" class="adf-comment-user-name">
-                {{comment.createdBy?.firstName}} {{comment.createdBy?.lastName}}
-            </div>
-            <div matLine [attr.id]="'comment-message-' + comment.id" class="adf-comment-message" [innerHTML]="comment.message">
-            </div>
-            <div matLine [attr.id]="'comment-time-' + comment.id" class="adf-comment-message-time">
-                {{ comment.created | adfTimeAgo: currentLocale }}
-            </div>
+            <div matLine class="adf-comment-user-name">{{ getUserFullName(comment.createdBy) }}</div>
```
AleksanderSklorz, line 18: I see in that file we have many function calls (you added couple new like getUserFullName). Can we avoid that and refer to fields directly? We have jira for replacing all places (https://alfresco.atlassian.net/browse/ADF-5539) but if we can avoid adding new functions calls now then we will have less work. 

For example you can try to map comments list to list of different type objects and then refer to their fields instead of functions.

lib/core/src/lib/comments/comment-list/comment-list.component.ts
```diff
@@ -71,24 +49,34 @@ export class CommentListComponent implements OnInit, OnDestroy {
...
+            result = `${user.firstName || ''} ${user.lastName || ''}`;
+        }
+
+        return result.trim();
     }
 
-    isPictureDefined(user: any): boolean {
-        return user.pictureId || user.avatarId;
+    isPictureDefined(user: User): boolean {
+        return user['pictureId'] || user['avatarId'];
```
AleksanderSklorz, line 76: hmm if we expect that User has these two fields - should not we update model instead of doing workarounds with []? It is a bit confusing that object has fields which are not defined in model

lib/core/src/lib/comments/comments.component.html
```diff
@@ -29,8 +29,8 @@
             </div>
     </div>
 
-    <div *ngIf="comments.length > 0">
+    <ng-container *ngIf="comments.length > 0">
```
AleksanderSklorz, line 32: Do we need that ng-container at all? It contains single element so no need for container as we can probably just add ngIf on adf-comment-list

lib/core/src/lib/comments/comment-list/comment-list.component.ts
DenysVuika: in both cases the types are missing and it requires some better approach as this interface is shared between different teams with different expectations.

lib/core/src/lib/comments/comment-list/comment-list.component.html
DenysVuika: I am try to keep the cleanup here, not rewriting the component or changing the scope of work. There's plenty of things here to be done, and if I start rewriting all the function calls, it will be definitely a big task that won't land any time soon

DenysVuika: moved some code to the Comment model

AleksanderSklorz (approved):

PR #8809, opened 2023-08-09
[AAE-15836] Include email in full name pipe if required

pmartinezga:

> - [ ] Bugfix
> - [X] Feature
> - [ ] Build related changes
> - [ ] Documentation

The fullName pipe returns the name for a user

Optionally we can include the email when using the fullName pipe. This can be achieved by using a parameter in the pipe or globally for the whole application when using the `ADF_FULL_NAME_PIPE_INCLUDE_EMAIL` injection token

> - [ ] Yes
> - [X] No

**Other information**:

### DIFF
--- a/lib/core/src/lib/pipes/full-name-email-required.token.ts
+++ b/lib/core/src/lib/pipes/full-name-email-required.token.ts
@@ -1,0 +1,23 @@
+/*!
+ * @license
+ * Copyright © 2005-2023 Hyland Software, Inc. and its affiliates. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* eslint-disable rxjs/no-subject-value */
+/* eslint-disable @typescript-eslint/naming-convention */
+
+import { InjectionToken } from '@angular/core';
+
+export const ADF_FULL_NAME_PIPE_INCLUDE_EMAIL = new InjectionToken<boolean>('ADF_FULL_NAME_PIPE_INCLUDE_EMAIL');
--- a/lib/core/src/lib/pipes/full-name.pipe.spec.ts
+++ b/lib/core/src/lib/pipes/full-name.pipe.spec.ts
@@ -16,42 +16,168 @@
  */
 
 import { FullNamePipe } from './full-name.pipe';
+import { UserLike } from './user-like.interface';
+
+interface TestCases {
+    [key: string]: {
+        title: string;
+        includeEmailToken: boolean | undefined;
+        includeEmailParameter: boolean | undefined;
+        testCases: TestCase[];
+    };
+};
+
+interface TestCase {
+    title: string;
+    user: UserLike;
+    result: string;
+};
 
 describe('FullNamePipe', () => {
 
     let pipe: FullNamePipe;
 
-    beforeEach(() => {
-        pipe = new FullNamePipe();
-    });
+    const cosntBaseTestCases: TestCase[] = [
+        {
+            title: 'should return empty string when there is no name',
+            user: { firstName: '', lastName: '', username: '', email: '' },
+            result: ''
+        },
+        {
+            title: 'should return only firstName as fullName when there is no lastName',
+            user: { firstName: 'Abc', lastName: '', username: '', email: '' },
+            result: 'Abc'
+        },
+        {
+            title: 'should return only lastName as fullName when there is no firstName',
+            user: { firstName: '', lastName: 'Xyz', username: '', email: '' },
+            result: 'Xyz'
+        },
+        {
+            title: 'should return fullName when firstName and lastName are available',
+            user: { firstName: 'Abc', lastName: 'Xyz', username: '', email: '' },
+            result: 'Abc Xyz'
+        },
+        {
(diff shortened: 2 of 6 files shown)


lib/core/src/lib/pipes/full-name.pipe.spec.ts
```diff
@@ -16,42 +16,168 @@
...
+};
 
 describe('FullNamePipe', () => {
 
     let pipe: FullNamePipe;
 
-    beforeEach(() => {
-        pipe = new FullNamePipe();
-    });
+    const cosntBaseTestCases: TestCase[] = [
```
DenysVuika, line 40: what does "cosnt" prefix mean?

lib/core/src/lib/pipes/full-name.pipe.spec.ts
```diff
@@ -16,42 +16,168 @@
...
+            includeEmailToken: true,
+            includeEmailParameter: true,
+            testCases: getTestCasesWithEmailPatched()
+        }
+    };
 
-    it('should return user eamil when firstName, lastName and username are not available', () => {
-        const user = {firstName : '', lastName : '', username: '', email: 'abcXyz@gmail.com'};
-        expect(pipe.transform(user)).toBe('abcXyz@gmail.com');
+    Object.keys(testCases).forEach(block => {
```
DenysVuika, line 174: this is some unconventional way of testing the pipe?

DenysVuika: if you followed Uncle Bob, the tests should be addressing one concern at a time, not a giant foreach having all tests in a single run

lib/core/src/lib/pipes/full-name-email-required.token.ts
```diff
@@ -0,0 +1,23 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* eslint-disable rxjs/no-subject-value */
+/* eslint-disable @typescript-eslint/naming-convention */
+
+import { InjectionToken } from '@angular/core';
+
+export const ADF_FULL_NAME_PIPE_INCLUDE_EMAIL = new InjectionToken<boolean>('ADF_FULL_NAME_PIPE_INCLUDE_EMAIL');
```
DenysVuika, line 23: I would say this is too over-engineered way to format a user full name... you literally introduce tone of code and tests just to compose "first name" and "last name". should not the applications or components just do that in the code instead?

lib/core/src/lib/pipes/full-name.pipe.spec.ts
```diff
@@ -16,42 +16,168 @@
  */
 
 import { FullNamePipe } from './full-name.pipe';
+import { UserLike } from './user-like.interface';
+
+interface TestCases {
```
DenysVuika, line 21: it's not a traditional unit testing

DenysVuika (changes requested):

DenysVuika:
@pmartinezga maybe you could introduce this pipe in the HxP, if there's a need for that? I don't see why this pipe should in the Core library to be honest

lib/core/src/lib/pipes/full-name.pipe.spec.ts
pmartinezga: My problem here is that if I try to test all the combinations for the injection token value and pipe parameter values, then I end up with a comment from sonar saying that I have a lot of duplicated code because the tests are always the same, call the pipe and check the output, and in many cases the output and the user are the same

lib/core/src/lib/pipes/full-name-email-required.token.ts
pmartinezga: The problem here is about the requirement requested by the POs.

They want to display also the email in the adf-people component that uses this pipe. But they want it to be displayed in HxP but not in APA. 

This is the less intrusive way I could think on, because the other alternative would be create a custom people component in hxp-apps which is exactly the same as the adf-people but with a custom pipe but anyway the pipe must be like this as it must be used differently in APA and in HxP

pmartinezga:
Thanks for the review!!

> @pmartinezga maybe you could introduce this pipe in the HxP, if there's a need for that? I don't see why this pipe should in the Core library to be honest

As I said before

> The problem here is about the requirement requested by the POs.
>They want to display also the email in the adf-people component that uses this pipe. But they want it to be displayed in HxP but not in APA. 
>This is the less intrusive way I could think on, because the other alternative would be create a custom people component in hxp-apps which is exactly the same as the adf-people but with a custom pipe but anyway the pipe must be like this as it must be used differently in APA and in HxP

Any suggestion is welcomed

lib/core/src/lib/pipes/full-name.pipe.spec.ts
pmartinezga: Yep my bad, I will change this

lib/core/src/lib/pipes/full-name.pipe.spec.ts
```diff
@@ -16,42 +16,81 @@
...
+            user: { firstName: '', lastName: '', username: 'username' },
+            result: 'username'
+        }
+    ];
 
-    it('should return only lastName as fullName when there is no firstName ', () => {
-        const user = {lastName : 'Xyz'};
-        expect(pipe.transform(user)).toBe('Xyz');
-    });
+    [undefined, false, true].forEach(injectionToken => {
```
DenysVuika, line 67: can you please split that into traditional unit tests, so that other colleagues will be able to read that? honestly, this looks like scientific code, not readable for newcomers

lib/core/src/lib/pipes/full-name-email-required.token.ts
DenysVuika: ok, could probably work for now, but this pipe looks too over-engineered already, we can completely get rid of this pipe in the future releases probably

lib/core/src/lib/pipes/full-name.pipe.spec.ts
pmartinezga: Sure, but I will get the duplication alert from Sonar for sure, but I think it is not a required check right?

PR #8821, opened 2023-08-14
[ACS-5857] Migrated adfMomentDate Pipe to date-fns equivalent 

SheenaMalhotra182:

> - [ ] Bugfix
> - [x] Feature
> - [ ] Build related changes
> - [ ] Documentation

Moment Library is being used in adfMomentDate Pipe.

This PR includes migration of _'lib/core/src/lib/pipes/moment-date.pipe.ts'_ into date-fns
1. Created a new Pipe (adfDate) which used date-fns library instead of Moment.
2. Replaced occurrences of adfMomentDate with adfDate and accordingly changed the ts and spec files

> - [ ] Yes
> - [X] No

**Other information**:
https://alfresco.atlassian.net/browse/ACS-5857

### DIFF
--- a/e2e/resources/forms/custom-date-form.ts
+++ b/e2e/resources/forms/custom-date-form.ts
@@ -93,7 +93,7 @@ export const customDateFormAPS1 = `{
                                     "existingColspan": 1,
                                     "maxColspan": 2
                                 },
-                                "dateDisplayFormat": "YY-M-D",
+                                "dateDisplayFormat": "yy-M-d",
                                 "layout": {
                                     "row": -1,
                                     "column": -1,
@@ -157,7 +157,7 @@ export const customDateFormAPS2 = `{
                            "existingColspan":1,
                            "maxColspan":2
                         },
-                        "dateDisplayFormat":"YY-M-D"
+                        "dateDisplayFormat":"yy-M-d"
                      }
                   ],
                   "2":[
--- a/e2e/resources/forms/custom-date-form.ts
+++ b/e2e/resources/forms/custom-date-form.ts
@@ -93,7 +93,7 @@ export const customDateFormAPS1 = `{
                                     "existingColspan": 1,
                                     "maxColspan": 2
                                 },
-                                "dateDisplayFormat": "yy-M-d",
+                                "dateDisplayFormat": "YY-M-D",
                                 "layout": {
                                     "row": -1,
                                     "column": -1,
@@ -157,7 +157,7 @@ export const customDateFormAPS2 = `{
                            "existingColspan":1,
                            "maxColspan":2
                         },
-                        "dateDisplayFormat":"yy-M-d"
+                        "dateDisplayFormat":"YY-M-D"
                      }
                   ],
                   "2":[
--- a/lib/core/src/lib/form/components/mock/form-renderer.component.mock.ts
+++ b/lib/core/src/lib/form/components/mock/form-renderer.component.mock.ts
@@ -1505,7 +1505,7 @@ export const formDateVisibility = {
                     existingColspan: 1,
                     maxColspan: 2
                   },
-                  dateDisplayFormat: 'yyyy-M-d'
+                  dateDisplayFormat: 'yyyy-MM-dd'
                 }
               ],
               2: [
--- a/lib/core/src/lib/form/components/mock/form-renderer.component.mock.ts
+++ b/lib/core/src/lib/form/components/mock/form-renderer.component.mock.ts
@@ -1505,7 +1505,7 @@ export const formDateVisibility = {
                     existingColspan: 1,
                     maxColspan: 2
                   },
-                  dateDisplayFormat: 'YYYY-MM-DD'
+                  dateDisplayFormat: 'd-M-yyyy'
                 }
               ],
               2: [
@@ -1525,7 +1525,7 @@ export const formDateVisibility = {
                     leftType: 'field',
                     leftValue: 'Date0hwq20',
                     operator: '==',
-                    rightValue: '2019-11-19',
+                    rightValue: '19-11-2019',
                     rightType: 'value',
                     nextConditionOperator: '',
                     nextCondition: null
@@ -1562,7 +1562,7 @@ export const formDateVisibility = {
                     leftType: 'field',
                     leftValue: 'Date0hwq20',
                     operator: '!=',
-                    rightValue: '2019-11-19',
+                    rightValue: '19-11-2019',
                     rightType: 'value',
                     nextConditionOperator: '',
                     nextCondition: null
--- a/lib/core/src/lib/form/components/mock/form-renderer.component.mock.ts
+++ b/lib/core/src/lib/form/components/mock/form-renderer.component.mock.ts
@@ -1505,7 +1505,7 @@ export const formDateVisibility = {
                     existingColspan: 1,
                     maxColspan: 2
                   },
-                  dateDisplayFormat: 'd-M-yyyy'
+                  dateDisplayFormat: 'yyyy-M-d'
                 }
               ],
               2: [
@@ -1525,7 +1525,7 @@ export const formDateVisibility = {
                     leftType: 'field',
                     leftValue: 'Date0hwq20',
                     operator: '==',
-                    rightValue: '19-11-2019',
+                    rightValue: '2019-11-19',
                     rightType: 'value',
                     nextConditionOperator: '',
                     nextCondition: null
@@ -1562,7 +1562,7 @@ export const formDateVisibility = {
                     leftType: 'field',
                     leftValue: 'Date0hwq20',
                     operator: '!=',
-                    rightValue: '19-11-2019',
+                    rightValue: '2019-11-19',
                     rightType: 'value',
                     nextConditionOperator: '',
                     nextCondition: null
--- a/lib/core/src/lib/form/components/mock/form-renderer.component.mock.ts
+++ b/lib/core/src/lib/form/components/mock/form-renderer.component.mock.ts
@@ -1842,7 +1842,7 @@ export const dateWidgetFormVisibilityMock = {
                                     existingColspan: 1,
                                     maxColspan: 2
                                 },
-                                dateDisplayFormat: 'D-M-YYYY'
+                                dateDisplayFormat: 'd-M-yyyy'
                             }
                         ]
                     }
--- a/lib/core/src/lib/form/components/widgets/core/form-field-validator.spec.ts
+++ b/lib/core/src/lib/form/components/widgets/core/form-field-validator.spec.ts
@@ -729,7 +729,7 @@ describe('FormFieldValidator', () => {
 
         it('should take into account that max value is in UTC and fail validating value checking the time', () => {
             const maxValueFromActivitiInput = '31-3-2018 12:00 AM';
-            const maxValueSavedInForm = format(parse(maxValueFromActivitiInput, 'd-M-yyyy hh:mm a', new Date()), `yyyy-MM-dd'T'HH:mm:ssxxx`);
+            const maxValueSavedInForm = format(parse(maxValueFromActivitiInput, 'dd-M-yyyy hh:mm a', new Date()), `yyyy-MM-dd'T'HH:mm:ssxxx`);
 
             const localInvalidValue = '2018-3-31 12:01 AM';
 
--- a/lib/core/src/lib/form/components/widgets/core/form-field-validator.spec.ts
+++ b/lib/core/src/lib/form/components/widgets/core/form-field-validator.spec.ts
@@ -714,7 +714,7 @@ describe('FormFieldValidator', () => {
 
         it('should take into account that max value is in UTC and NOT fail validating value checking the time', () => {
             const maxValueFromActivitiInput = '31-3-2018 12:00 AM';
-            const maxValueSavedInForm = format(parse(maxValueFromActivitiInput, 'dd-MM-yyyy hh:mm a', new Date()), "yyyy-MM-dd'T'HH:mm:ssXXX");
+            const maxValueSavedInForm = format(parse(maxValueFromActivitiInput, 'dd-MM-yyyy hh:mm a', new Date()), `yyyy-MM-dd'T'HH:mm:ssXXX`);
 
             const localValidValue = '2018-3-30 11:59 PM';
 
@@ -729,7 +729,7 @@ describe('FormFieldValidator', () => {
 
         it('should take into account that max value is in UTC and fail validating value checking the time', () => {
             const maxValueFromActivitiInput = '31-3-2018 12:00 AM';
-            const maxValueSavedInForm = format(parse(maxValueFromActivitiInput, 'd-M-yyyy hh:mm a', new Date()), "yyyy-MM-dd'T'HH:mm:ssxxx");
+            const maxValueSavedInForm = format(parse(maxValueFromActivitiInput, 'd-M-yyyy hh:mm a', new Date()), `yyyy-MM-dd'T'HH:mm:ssxxx`);
 
             const localInvalidValue = '2018-3-31 12:01 AM';
 
@@ -840,7 +840,7 @@ describe('FormFieldValidator', () => {
 
         it('should take into account that min value is in UTC and NOT fail validating value checking the time', () => {
             const minValueFromActivitiInput = '02-3-2018 06:00 AM';
-            const minValueSavedInForm = format(parse(minValueFromActivitiInput, 'dd-MM-yyyy hh:mm a', new Date()), "yyyy-MM-dd'T'HH:mm:ssXXX");
+            const minValueSavedInForm = format(parse(minValueFromActivitiInput, 'dd-MM-yyyy hh:mm a', new Date()), `yyyy-MM-dd'T'HH:mm:ssXXX`);
 
             const localValidValue = '2018-3-02 06:01 AM';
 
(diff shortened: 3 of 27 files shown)


rbahirsheth (approved):
Approving the PR to check the e2e and build pass.

eromano (changes requested):
this looks a Braking change means that all the form definition with             dateDisplayFormat: 'DD-MM-YYYY' will not work anynore as before or not work at all?

SheenaMalhotra182:
> this looks a Braking change means that all the form definition with dateDisplayFormat: 'DD-MM-YYYY' will not work anynore as before or not work at all?

This will work the same as before. We have changed the pattern here as date-fns does not support DD-MM-YYYY.

lib/process-services-cloud/src/lib/form/components/widgets/date/date-cloud.widget.ts
```diff
@@ -51,51 +49,52 @@ import { DATE_FORMAT_CLOUD } from '../../../../models/date-format-cloud.model';
...
-                    this.maxDate = moment(this.field.maxValue, DATE_FORMAT_CLOUD);
+                    this.maxDate = format(new Date(this.field.maxValue), DATE_FORMAT_CLOUD);
                 }
             }
         }
     }
 
-    getTodaysFormattedDate() {
-        return moment().format(DATE_FORMAT_CLOUD);
+    getTodaysDate() {
```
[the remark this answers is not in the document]
SheenaMalhotra182: We no longer need formatted date here as the formatter is already being used when we assign value to minDate/maxDate around line 76.

The reason why I haven't called `new Date()` directly in the code is that, In unit tests, we need to spy on `getTodaysFormattedDate()` and return a value.
Also, I have renamed `getTodaysFormattedDate()` to `getTodaysDate()` because we are not formatting the date in that function anymore.

eromano:
> > this looks a Braking change means that all the form definition with dateDisplayFormat: 'DD-MM-YYYY' will not work anynore as before or not work at all?
> 
> This will work the same as before. We have changed the pattern here as date-fns does not support DD-MM-YYYY.

It's not going to work as before in order to make those test pass you have modified also the test forms definition                         "dateDisplayFormat":"yy-M-d"
 but our customers in theyr database will still have the previous rappresentation aspecting to have the same result as before instead it will not work anymore we may need to provide some sort of compatibility layer here @DenysVuika  can you please help here

lib/process-services-cloud/src/lib/form/components/widgets/date/date-cloud.widget.ts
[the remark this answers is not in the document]
MichalKinas: @SheenaMalhotra182 adding a function just to be able to create a spy for it doesn't make any sense. We shouldn't alter the component for the sake of unit tests. And you don't have to add the spy to make unit tests work in this case. You can directly call `new Date()` and then use the same in unit tests to create expected value

MichalKinas (changes requested):

MichalKinas:
Also @SheenaMalhotra182 where did you get the information that `DD-MM-YYYY` will not work with date-fns? Could you please point me to some sources/docs saying that? I've just tested it on date-fns playground and  it seems that `format(new Date(), 'dd-MM-yyyy')` works as expected printing `31-08-2023`

SheenaMalhotra182:
> Also @SheenaMalhotra182 where did you get the information that `DD-MM-YYYY` will not work with date-fns? Could you please point me to some sources/docs saying that? I've just tested it on date-fns playground and it seems that `format(new Date(), 'dd-MM-yyyy')` works as expected printing `31-08-2023`

In date-fns, 'DD' refers to the days of the year (1st, 2nd, ...., 365th) and 'dd' refers to the days of the month (1st, 2nd, ...., 31st).
We have changed the format accordingly. 
Official Documentation => https://date-fns.org/v2.30.0/docs/format.

And what you have quoted as an example above `format(new Date(), 'dd-MM-yyyy')` is also using the lowercase format therefore you're getting the correct date.

Also, using 'DD' giver errors if used in 'DD-MM-YYYY' format. 
date-fns

lib/core/src/lib/form/components/mock/form-renderer.component.mock.ts
```diff
@@ -1505,7 +1505,7 @@ export const formDateVisibility = {
                     existingColspan: 1,
                     maxColspan: 2
                   },
-                  dateDisplayFormat: 'YYYY-MM-DD'
+                  dateDisplayFormat: 'yyyy-M-d'
```
MichalKinas, line 1508: I think this should be `yyyy-MM-dd` to provide compatibility for previous implementation

lib/core/src/lib/form/components/widgets/core/form-field-validator.spec.ts
```diff
@@ -729,7 +729,7 @@ describe('FormFieldValidator', () => {
 
         it('should take into account that max value is in UTC and fail validating value checking the time', () => {
             const maxValueFromActivitiInput = '31-3-2018 12:00 AM';
-            const maxValueSavedInForm = moment(maxValueFromActivitiInput, 'DD-M-YYYY hh:mm A').utc().format();
+            const maxValueSavedInForm = format(parse(maxValueFromActivitiInput, 'd-M-yyyy hh:mm a', new Date()), `yyyy-MM-dd'T'HH:mm:ssxxx`);
```
MichalKinas, line 732:
```suggestion
            const maxValueSavedInForm = format(parse(maxValueFromActivitiInput, 'dd-M-yyyy hh:mm a', new Date()), `yyyy-MM-dd'T'HH:mm:ssxxx`);
```

rbahirsheth:
> > > this looks a Braking change means that all the form definition with dateDisplayFormat: 'DD-MM-YYYY' will not work anynore as before or not work at all?
> > 
> > 
> > This will work the same as before. We have changed the pattern here as date-fns does not support DD-MM-YYYY.
> 
> It's not going to work as before in order to make those test pass you have modified also the test forms definition "dateDisplayFormat":"yy-M-d" but our customers in theyr database will still have the previous rappresentation aspecting to have the same result as before instead it will not work anymore we may need to provide some sort of compatibility layer here @DenysVuika can you please help here

@DenysVuika Please suggest how we can make the changes that would be compatible with the existing customers too.
@eromano There might be changes required from the HxP side also. Am I right?

lib/process-services-cloud/src/lib/form/components/widgets/date/date-cloud.widget.ts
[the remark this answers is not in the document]
SheenaMalhotra182: changed. calling `new Date()` directly in code now

lib/core/src/lib/form/components/mock/form-renderer.component.mock.ts
SheenaMalhotra182: changed to `yyyy-MM-dd `

lib/core/src/lib/form/components/widgets/core/form-field-validator.spec.ts
SheenaMalhotra182: changed to 'dd'

lib/core/src/lib/form/components/widgets/date/date.widget.ts
```diff
@@ -50,36 +50,31 @@ import { takeUntil } from 'rxjs/operators';
 })
 export class DateWidgetComponent extends WidgetComponent implements OnInit, OnDestroy {
 
-    DATE_FORMAT = 'DD-MM-YYYY';
+    DATE_FORMAT = 'd-M-yyyy';
```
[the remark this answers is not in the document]
SheenaMalhotra182: changed to 'dd'

e2e/resources/forms/custom-date-form.ts
```diff
@@ -93,7 +93,7 @@ export const customDateFormAPS1 = `{
                                     "existingColspan": 1,
                                     "maxColspan": 2
                                 },
-                                "dateDisplayFormat": "YY-M-D",
+                                "dateDisplayFormat": "yy-M-d",
```
DenysVuika, line 96: this format is coming from the backend, it should remain the same not to break all existing forms

DenysVuika (changes requested):

e2e/resources/forms/custom-date-form.ts
SheenaMalhotra182: I remember making that format change in custom-date.form.ts because some e2es were breaking because of the format. I will revert it and test it again.

SheenaMalhotra182: I have reverted the changes from _'custom-date-form.ts'_. 
For the e2e to run properly, we still need the _dateDisplayFormat_ to be _'yy-M-d'_ so that the form config can be rendered as a form  (as the date widget now supports date-fns formats).

So, I have added an `updateDateFormat()` method in _'form-demo.page.ts'_ to change dateFormat while running the e2e test case.

SheenaMalhotra182: Upon re-analyzing, I think that this approach would be wrong too because ideally, we should not change the form config to make the e2es work because the functionality would still be impacted unless we change the 'dateDisplayFormat' from the backend as well. So for now, I will revert the fixes I did for this (`updateDateFormat()`).

Also, the 5 e2es that are failing in the build ( C268818, C268819, C290040, C313200, C310199 ) are fetching forms from backend ( e2e/resources/activiti7/candidatebaseapp.zip, e2e/resources/activiti7/simpleapp.zip, e2e/resources/apps/WidgetApps.zip ). We probably need to get the 'dateDisplayFormat' changed in these form jsons as well. 
@DenysVuika @eromano

lib/core/src/lib/form/components/mock/form-renderer.component.mock.ts
```diff
@@ -1505,7 +1505,7 @@ export const formDateVisibility = {
                     existingColspan: 1,
                     maxColspan: 2
                   },
-                  dateDisplayFormat: 'YYYY-MM-DD'
+                  dateDisplayFormat: 'yyyy-MM-dd'
```
DenysVuika, line 1508: should not be changed

lib/core/src/lib/form/components/mock/form-renderer.component.mock.ts
```diff
@@ -1842,7 +1842,7 @@ export const dateWidgetFormVisibilityMock = {
                                     existingColspan: 1,
                                     maxColspan: 2
                                 },
-                                dateDisplayFormat: 'D-M-YYYY'
+                                dateDisplayFormat: 'd-M-yyyy'
```
DenysVuika, line 1845: should not be changed

lib/core/src/lib/form/components/widgets/core/form-field.model.spec.ts
```diff
@@ -257,25 +257,25 @@ describe('FormFieldModel', () => {
             id: 'ddmmyyy',
             name: 'DD-MM-YYYY',
             type: 'date',
-            value: 'today',
```
DenysVuika, line 260: should not be changed

lib/core/src/lib/form/components/widgets/core/form-field.model.spec.ts
```diff
@@ -245,7 +245,7 @@ describe('FormFieldModel', () => {
                     readOnly: false
                 }
             },
-            dateDisplayFormat: 'DD-MM-YYYY'
+            dateDisplayFormat: 'dd-MM-yyyy'
```
DenysVuika, line 248: should not be changed

lib/core/src/lib/form/components/widgets/core/form-field.model.spec.ts
```diff
@@ -167,7 +167,7 @@ describe('FormFieldModel', () => {
                     readOnly: false
                 }
             },
-            dateDisplayFormat: 'MM-DD-YYYY'
+            dateDisplayFormat: 'MM-dd-yyyy'
```
DenysVuika, line 170: should not be changed

lib/core/src/lib/form/components/widgets/core/form-field.model.spec.ts
```diff
@@ -193,7 +193,7 @@ describe('FormFieldModel', () => {
                     readOnly: false
                 }
             },
-            dateDisplayFormat: 'MM-YY-DD'
+            dateDisplayFormat: 'MM-yy-dd'
```
DenysVuika, line 196: should not be changed

lib/core/src/lib/form/components/widgets/core/form-field.model.spec.ts
```diff
@@ -257,25 +257,25 @@ describe('FormFieldModel', () => {
...
+            value: startOfDay(new Date()),
             required: false,
             readOnly: false,
             params: {
                 field: {
                     id: 'ddmmyyy',
                     name: 'DD-MM-YYYY',
                     type: 'date',
-                    value: 'today',
+                    value: startOfDay(new Date()),
```
DenysVuika, line 268: server data, should not be changed

lib/core/src/lib/form/components/widgets/core/form-field.model.spec.ts
```diff
@@ -257,25 +257,25 @@ describe('FormFieldModel', () => {
...
                     required: false,
                     readOnly: false
                 }
             },
-            dateDisplayFormat: 'DD-MM-YYYY'
+            dateDisplayFormat: 'dd-MM-yyyy'
```
DenysVuika, line 273: server data, should not be changed

lib/core/src/lib/form/components/widgets/core/form-field.model.spec.ts
```diff
@@ -288,25 +288,28 @@ describe('FormFieldModel', () => {
...
                     name: 'date and time',
                     type: 'datetime',
-                    value: 'now',
+                    value: new Date(),
                     required: false,
                     readOnly: false
                 }
             },
-            dateDisplayFormat: 'YYYY-MM-DD HH:mm'
+            dateDisplayFormat: 'yyyy-MM-dd HH:mm'
```
DenysVuika, line 304: server data, should not be changed

lib/core/src/lib/form/components/widgets/date-time/date-time.widget.spec.ts
```diff
@@ -217,7 +218,7 @@ describe('DateTimeWidgetComponent', () => {
                 id: 'date-field-id',
                 name: 'date-name',
                 value: '12-30-9999 10:30 AM',
-                dateDisplayFormat: 'MM-DD-YYYY HH:mm A',
+                dateDisplayFormat: 'MM-dd-yyyy hh:mm a',
```
DenysVuika, line 221: server data, should not be changed

lib/core/src/lib/form/components/widgets/date-time/date-time.widget.spec.ts
```diff
@@ -243,7 +244,7 @@ describe('DateTimeWidgetComponent', () => {
             readOnly: 'false'
         });
         field.isVisible = true;
-        field.dateDisplayFormat = 'MM-DD-YYYY HH:mm A';
+        field.dateDisplayFormat = 'MM-dd-yyyy hh:mm a';
```
DenysVuika, line 247: server data, should not be changed

lib/core/src/lib/form/components/widgets/date/date.widget.spec.ts
```diff
@@ -242,7 +244,7 @@ describe('DateWidgetComponent', () => {
             readOnly: 'false'
         });
         field.isVisible = true;
-        field.dateDisplayFormat = 'MM-DD-YYYY';
+        field.dateDisplayFormat = 'MM-dd-yyyy';
```
DenysVuika, line 247: should be in the moment format, as it's the activiti data

lib/core/src/lib/form/components/widgets/date/date.widget.spec.ts
```diff
@@ -170,23 +172,23 @@ describe('DateWidgetComponent', () => {
                 readOnly: 'false'
             });
             widget.field.isVisible = true;
-            widget.field.dateDisplayFormat = 'MM-DD-YYYY';
+            widget.field.dateDisplayFormat = 'MM-dd-yyyy';
```
DenysVuika, line 175: should be in the moment format, as it's the activiti data

lib/core/src/lib/form/components/widgets/date/date.widget.spec.ts
```diff
@@ -170,23 +172,23 @@ describe('DateWidgetComponent', () => {
...
 
             fixture.detectChanges();
 
             let dateElement: any = element.querySelector('#date-field-id');
             expect(dateElement.value).toContain('12-30-9999');
 
             widget.field.value = '5-6-2019 00:00';
-            widget.field.dateDisplayFormat = 'D-M-YYYY HH:mm';
+            widget.field.dateDisplayFormat = 'd-M-yyyy HH:mm';
```
DenysVuika, line 183: should be in the moment format, as it's the activiti data

lib/core/src/lib/form/components/widgets/date/date.widget.spec.ts
```diff
@@ -170,23 +172,23 @@ describe('DateWidgetComponent', () => {
...
 
             fixture.detectChanges();
 
             dateElement = element.querySelector('#date-field-id');
             expect(dateElement.value).toContain('5-6-2019 00:00');
 
             widget.field.value = '05.06.2019';
-            widget.field.dateDisplayFormat = 'DD.MM.YYYY';
+            widget.field.dateDisplayFormat = 'dd.MM.yyyy';
```
DenysVuika, line 191: should be in the moment format, as it's the activiti data

DenysVuika:
closing due to many issues

PR #8924, opened 2023-09-22
[ACS-5857] Moment to Date-Fns Migration for adfMomentDate Pipe.

SheenaMalhotra182:

> - [ ] Bugfix
> - [x] Feature
> - [ ] Build related changes
> - [ ] Documentation

Moment Library is being used in adfMomentDate Pipe.

This PR includes migration of _'lib/core/src/lib/pipes/moment-date.pipe.ts'_ into date-fns
1. Created a new Pipe (adfDate) which uses date-fns library instead of Moment.
2. Replaced occurrences of adfMomentDate with adfDate in **'DateWidgetComponent'** and **'DateCloudWidgetComponent'** and accordingly changed the ts and spec files
3. Replaced occurrences of adfMomentDate with adfDate in **'DateTimeWidgetComponent'** and accordingly changed the ts and spec files

> - [ ] Yes
> - [X] No

**Other information**:
https://alfresco.atlassian.net/browse/ACS-5857

There's one unit test case which needed the format of test data to be changed in order to validate it with the default format =>
https://github.com/Alfresco/alfresco-ng2-components/pull/8924/files#diff-b52520351ccf61f2557086298ce2a0d0e574463d39d0fa921085bcc2af1c1760R1116.
Rest all unit tests and e2es work fine without needing any changes.


lib/core/src/lib/form/services/date-format-translation.service.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+import { TestBed } from '@angular/core/testing';
+import { DateFormatTranslationService } from './date-format-translation.service';
+import { TranslateModule } from '@ngx-translate/core';
+import { CoreTestingModule } from '../../testing/core.testing.module';
+
+describe('DateFormatTranslationService', () => {
+    let service: DateFormatTranslationService;
+
+    beforeEach(() => {
+        TestBed.configureTestingModule({
```
DenysVuika, line 27: you don't need this block, the service does not use Translation or Root, and it has providedIn: root, so you don't need setting the "providers" section too

lib/core/src/lib/pipes/date.pipe.ts
```diff
@@ -0,0 +1,30 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { Pipe, PipeTransform } from '@angular/core';
+import { isValid } from 'date-fns';
+import { DateFormatTranslationService } from '../form/services/date-format-translation.service';
+
+@Pipe({ name: 'adfDate' })
+export class DatePipe implements PipeTransform {
```
DenysVuika, line 23: the naming of class is going to clash with Angular DatePipe: https://angular.io/api/common/DatePipe
please use AdfDatePipe as the pipe name suggests

lib/core/src/lib/form/components/widgets/core/form-field.model.ts
```diff
@@ -336,20 +350,34 @@ export class FormFieldModel extends FormWidgetModel {
...
          */
-        if (this.isDateField(json) || this.isDateTimeField(json)) {
+         if (this.isDateTimeField(json)) {
             if (value) {
                 let dateValue;
                 if (isNumberValue(value)) {
                     dateValue = moment(value);
                 } else {
-                    dateValue = this.isDateTimeField(json) ? moment.utc(value, 'YYYY-MM-DD hh:mm A') : moment.utc(value.split('T')[0], 'YYYY-M-D');
+                    dateValue = moment.utc(value, 'YYYY-MM-DD hh:mm A');
```
DenysVuika, line 359: this is still using moment, but the rest of the pipes in your PR already deal with new DatePipe that expects date only...

lib/core/src/lib/form/components/widgets/core/form-field.model.spec.ts
```diff
@@ -19,9 +19,16 @@ import moment from 'moment';
 import { FormFieldTypes } from './form-field-types';
 import { FormFieldModel } from './form-field.model';
 import { FormModel } from './form.model';
+import { DateFormatTranslationService } from '../../../public-api';
```
DenysVuika, line 22: incorrect import path

lib/core/src/lib/form/components/widgets/core/form-field.model.ts
```diff
@@ -239,6 +244,15 @@ export class FormFieldModel extends FormWidgetModel {
     }
 
     private getDefaultDateFormat(jsonField: any): string {
+        if(jsonField.fields) {
+            Object.keys(jsonField.fields).forEach((el) => {
+                if(jsonField.fields[el]) {
+                    jsonField.fields[el].forEach((element) => {
+                        element.dateDisplayFormat = element.dateDisplayFormat? dateFormatTranslationService.convertMomentToDateFnsFormat(element.dateDisplayFormat): element.dateDisplayFormat;
```
DenysVuika, line 251: this is missing the time

lib/core/src/lib/form/services/date-format-translation.service.ts
```diff
@@ -0,0 +1,60 @@
...
+            dateDisplayFormat = dateDisplayFormat.replace(new RegExp(search, 'g'), replace);
+        }
+        return dateDisplayFormat;
+    }
+
+    format(date: number|Date, dateFormat: string): string {
+        return format(date, this.convertMomentToDateFnsFormat(dateFormat));
+    }
+
+    parse(value: string, dateFormat: string, date = new Date()): Date {
```
DenysVuika, line 57: this function is confusing, why it takes date parameter and why it sets it to new date() ?

lib/process-services-cloud/src/lib/form/components/widgets/date/date-cloud.widget.ts
```diff
@@ -51,62 +52,59 @@ import { DATE_FORMAT_CLOUD } from '../../../../models/date-format-cloud.model';
...
     ngOnDestroy() {
         this.onDestroy$.next(true);
         this.onDestroy$.complete();
     }
 
     onDateChanged(newDateValue) {
-        const date = moment(newDateValue, this.field.dateDisplayFormat, true);
-        if (date.isValid()) {
-            this.field.value = date.format(this.field.dateDisplayFormat);
+        const date = this.dateFormatTranslationService.parse(newDateValue, this.field.dateDisplayFormat, new Date());
```
DenysVuika, line 105: why do you pass new date there as third param? what is the reason for that?

lib/core/src/lib/form/components/widgets/date/date.widget.html
```diff
@@ -1,6 +1,6 @@
 <div class="{{field.className}}" id="data-widget" [class.adf-invalid]="!field.isValid && isTouched()">
     <mat-form-field class="adf-date-widget" [hideRequiredMarker]="true">
-        <label class="adf-label" [attr.for]="field.id">{{field.name | translate }} ({{field.dateDisplayFormat}})<span class="adf-asterisk"
+        <label class="adf-label" [attr.for]="field.id">{{field.name | translate }} ({{dateFormatTranslationService.convertDateFnsToMomentFormat(field.dateDisplayFormat)}})<span class="adf-asterisk"
```
DenysVuika, line 3: please extract to a separate function so that template does not deal with services conversion directly:

```ts
formatLabel(field: FormField): string {
  const displayName = this.translationService.instant(field.name);
  const displayFormat = this.dateFormatTranslationService.convertDateFnsToMomentFormat(field.dateDisplayFormat);

  return `${displayName} (${displayFormat})`; 
}
```

then in HTML:

```html
<label>{formatLabel(field)}</label>
```

also provide unit tests for the formatLabel function as it becomes easier

lib/core/src/lib/pipes/date.pipe.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+import { TranslateModule } from '@ngx-translate/core';
+import { DatePipe } from './date.pipe';
+import { DateFormatTranslationService } from '../form/public-api';
+
+describe('DatePipe', () => {
+    let datePipe: DatePipe;
+
+    beforeEach(() => {
+        TestBed.configureTestingModule({
+            imports: [TranslateModule.forRoot(), CoreTestingModule],
```
DenysVuika, line 29: check if you need those imports, probably not

lib/core/src/lib/pipes/date.pipe.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { TestBed } from '@angular/core/testing';
+import { CoreTestingModule } from '../testing/core.testing.module';
+import { TranslateModule } from '@ngx-translate/core';
+import { DatePipe } from './date.pipe';
+import { DateFormatTranslationService } from '../form/public-api';
```
DenysVuika, line 22: incorrect import path

lib/process-services-cloud/src/lib/form/components/widgets/date/date-cloud.widget.html
```diff
@@ -1,11 +1,11 @@
 <div class="{{field.className}}" id="data-widget" [class.adf-invalid]="!field.isValid && isTouched()" [class.adf-left-label-input-container]="field.leftLabels">
     <div *ngIf="field.leftLabels">
-        <label class="adf-label adf-left-label" [attr.for]="field.id">{{field.name | translate }} ({{field.dateDisplayFormat}})<span class="adf-asterisk"
+        <label class="adf-label adf-left-label" [attr.for]="field.id">{{field.name | translate }} ({{dateFormatTranslationService.convertDateFnsToMomentFormat(field.dateDisplayFormat)}})<span class="adf-asterisk"
```
DenysVuika, line 3: this needs to be extracted to a function, UI templates should not deal with services directly, see https://github.com/Alfresco/alfresco-ng2-components/pull/8924/files#r1335592062

lib/process-services-cloud/src/lib/form/components/widgets/date/date-cloud.widget.html
```diff
@@ -1,11 +1,11 @@
...
             *ngIf="isRequired()">*</span></label>
      </div>
     <div>
         <mat-form-field class="adf-date-widget" [class.adf-left-label-input-datepicker]="field.leftLabels" [hideRequiredMarker]="true">
-            <label class="adf-label" *ngIf="!field.leftLabels" [attr.for]="field.id">{{field.name | translate }} ({{field.dateDisplayFormat}})<span class="adf-asterisk"
+            <label class="adf-label" *ngIf="!field.leftLabels" [attr.for]="field.id">{{field.name | translate }} ({{dateFormatTranslationService.convertDateFnsToMomentFormat(field.dateDisplayFormat)}})<span class="adf-asterisk"
```
DenysVuika, line 8: this needs to be extracted to a function, UI templates should not deal with services directly, see https://github.com/Alfresco/alfresco-ng2-components/pull/8924/files#r1335592062

lib/process-services-cloud/src/lib/form/components/widgets/date/date-cloud.widget.ts
```diff
@@ -51,62 +52,59 @@ import { DATE_FORMAT_CLOUD } from '../../../../models/date-format-cloud.model';
...
                 if (Number.isInteger(this.field.maxDateRangeValue)) {
-                    this.maxDate = moment(today).add(this.field.maxDateRangeValue, 'days');
-                    this.field.maxValue = this.maxDate.format(DATE_FORMAT_CLOUD);
+                    this.maxDate = this.dateFormatTranslationService.format(addDays(today, this.field.maxDateRangeValue), DATE_FORMAT_CLOUD);
+                    this.field.maxValue = this.maxDate;
                 }
             } else {
                 if (this.field.minValue) {
-                    this.minDate = moment(this.field.minValue, DATE_FORMAT_CLOUD);
+                    this.minDate = this.dateFormatTranslationService.format(new Date(this.field.minValue), DATE_FORMAT_CLOUD);
```
DenysVuika, line 89: why do you need translation if you already have `new Date()` and `export const DATE_FORMAT_CLOUD = 'yyyy-MM-dd';` ? you can probably use datefns directly?

lib/process-services-cloud/src/lib/form/components/widgets/date/date-cloud.widget.ts
```diff
@@ -51,62 +52,59 @@ import { DATE_FORMAT_CLOUD } from '../../../../models/date-format-cloud.model';
...
             .pipe(takeUntil(this.onDestroy$))
-            .subscribe(locale => this.dateAdapter.setLocale(locale));
+            .subscribe(locale => this.dateAdapter.setLocale(DateFnsUtils.getLocaleFromString(locale)));
 
-        const momentDateAdapter = this.dateAdapter as MomentDateAdapter;
-        momentDateAdapter.overrideDisplayFormat = this.field.dateDisplayFormat;
+        this.dateFormatConfig.display.dateInput = this.field.dateDisplayFormat;
 
         if (this.field) {
             if (this.field.dynamicDateRangeSelection) {
```
DenysVuika, line 77: review this block as you probably don't need format translation service as `today` is a Date object

lib/core/src/lib/pipes/date.pipe.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+
+        const result = datePipe.transform(inputDate, dateFormat);
+
+        expect(result).toBe(expectedOutput);
+    });
+
+    it('should return the input value when given an invalid date object', () => {
+        const inputDate = new Date('invalid');
+        const dateFormat = 'DD-MM-YYYY';
+        const expectedOutput = inputDate.toString();
```
DenysVuika, line 48: your expected output should not be based on .toString() conversion, as whatever happens there the test will be just green. Add exact value it expects instead.

lib/core/src/lib/form/services/date-format-translation.service.ts
DenysVuika: if there are no cases when you pass the date here, please use "new Date()" inside the "parse" call instead

lib/core/src/lib/form/components/widgets/core/form-field-validator.ts
```diff
@@ -583,3 +584,39 @@ export const FORM_FIELD_VALIDATORS = [
     new MinDateTimeFieldValidator(),
     new MaxDateTimeFieldValidator()
 ];
+
+export const momentToDateFnsMap = {
```
DenysVuika, line 588: you are using this maps in 2 places, second is translation service; extract to a separate class and use same import in both cases

lib/core/src/lib/form/components/widgets/core/form-field-validator.ts
```diff
@@ -583,3 +584,39 @@ export const FORM_FIELD_VALIDATORS = [
...
+};
+
+export const dateFnsToMomentMap = {
+    M: 'M',
+    d: 'D',
+    y: 'Y',
+    a: 'A'
+};
+
+export function convertMomentToDateFnsFormat(dateDisplayFormat: string): string {
```
DenysVuika, line 602: provide documentation comments for public functions

lib/core/src/lib/form/services/date-format-translation.service.ts
```diff
@@ -0,0 +1,60 @@
...
+ */
+
+import { Injectable } from '@angular/core';
+import { format, parse } from 'date-fns';
+
+@Injectable({
+    providedIn: 'root'
+})
+export class DateFormatTranslationService {
+    private momentToDateFnsMap = {
```
DenysVuika, line 25: you are using this maps in 2 places, extract to a separate class and use same import in both cases

lib/core/src/lib/pipes/date.pipe.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+ * limitations under the License.
+ */
+
+import { TestBed } from '@angular/core/testing';
+import { CoreTestingModule } from '../testing/core.testing.module';
+import { TranslateModule } from '@ngx-translate/core';
+import { DatePipe } from './date.pipe';
+import { DateFormatTranslationService } from '../form/public-api';
+
+describe('DatePipe', () => {
```
DenysVuika, line 24: please provide tests with different variations of the date format, i.e. 
- YYYY-MM-DD
- YYYY-DD-MM
- MM-DD-YY

lib/core/src/lib/pipes/pipe.module.ts
```diff
@@ -36,6 +36,7 @@ import { MomentDateTimePipe } from './moment-datetime.pipe';
 import { FilterStringPipe } from './filter-string.pipe';
 import { FilterOutArrayObjectsByPropPipe } from './filter-out-every-object-by-prop.pipe';
 import { DateTimePipe } from './date-time.pipe';
+import { DatePipe } from './date.pipe';
```
DenysVuika, line 39: should be AdfDatePipe not to clash with Angular DatePipe

lib/core/src/lib/form/components/widgets/core/form-field-validator.ts
```diff
@@ -583,3 +584,39 @@ export const FORM_FIELD_VALIDATORS = [
...
+    for (const [search, replace] of Object.entries(momentToDateFnsMap)) {
+        dateDisplayFormat = dateDisplayFormat.replace(new RegExp(search, 'g'), replace);
+    }
+    return dateDisplayFormat;
+}
+
+export function convertDateFnsToMomentFormat(dateDisplayFormat: string): string {
```
DenysVuika, line 609: move all these functions to `DateFnsUtils` class to have a single place for all date/time code

lib/core/src/lib/form/components/widgets/core/form-field-validator.ts
```diff
@@ -583,3 +584,39 @@ export const FORM_FIELD_VALIDATORS = [
...
+    for (const [search, replace] of Object.entries(dateFnsToMomentMap)) {
+        dateDisplayFormat = dateDisplayFormat.replace(new RegExp(search, 'g'), replace);
+    }
+    return dateDisplayFormat;
+}
+
+export function formatUpdate(date: number|Date, dateFormat: string): string {
```
DenysVuika, line 616: move to DateFnsUtils class and use better naming, like `formatDate`, it is already implied that some update is going to happen

lib/core/src/lib/form/components/widgets/core/form-field-validator.ts
```diff
@@ -583,3 +584,39 @@ export const FORM_FIELD_VALIDATORS = [
...
+    return format(date, convertMomentToDateFnsFormat(dateFormat));
+}
+
+export function parseUpdate(value: string, dateFormat: string, date = new Date()): Date {
```
DenysVuika, line 620: move to DateFnsUtils class and use better naming, like `parseDate`, it is already implied that some update is going to happen

lib/core/src/lib/form/services/date-format-translation.service.ts
```diff
@@ -0,0 +1,60 @@
...
+ * limitations under the License.
+ */
+
+import { Injectable } from '@angular/core';
+import { format, parse } from 'date-fns';
+
+@Injectable({
+    providedIn: 'root'
+})
+export class DateFormatTranslationService {
```
DenysVuika, line 24: This service has a lot of duplicated code with Field Validator funcitons. 
you probably don't need this service at all, if you move all the field validator functions to DateFnsUtils, you can use those utils without this service

lib/core/src/lib/common/utils/date-fns-utils.spec.ts
```diff
@@ -55,4 +55,13 @@ describe('Date Format Translations', () => {
 
         expect(result).toEqual(expectedParsedDate);
     });
+
+    it('should add seconds to a date time value', () => {
+        const inputDateTime = '2023-09-02T10:00:00Z';
```
rbahirsheth, line 60: Please correct the millisecond in time.

lib/core/src/lib/form/components/widgets/core/form-field.model.spec.ts
```diff
@@ -15,10 +15,10 @@
  * limitations under the License.
  */
 
-import moment from 'moment';
 import { FormFieldTypes } from './form-field-types';
 import { FormFieldModel } from './form-field.model';
 import { FormModel } from './form.model';
+import { DateFnsUtils } from '../../../../common/utils/date-fns-utils';
```
rbahirsheth, line 21: Please check the import path all places.

lib/core/src/lib/form/components/widgets/core/form-field.model.ts
```diff
@@ -239,6 +241,15 @@ export class FormFieldModel extends FormWidgetModel {
     }
 
     private getDefaultDateFormat(jsonField: any): string {
+        if(jsonField.fields) {
+            Object.keys(jsonField.fields).forEach((el) => {
+                if(jsonField.fields[el]) {
+                    jsonField.fields[el].forEach((element) => {
+                        element.dateDisplayFormat = element.dateDisplayFormat? DateFnsUtils.convertMomentToDateFnsFormat(element.dateDisplayFormat): element.dateDisplayFormat;
```
rbahirsheth, line 248: Please add the space before '?'.
What if element.dateDisplayFormat is null?

SheenaMalhotra182: fixed the linting. 
This already includes null and '' checks.

lib/core/src/lib/form/components/widgets/date-time/date-time.widget.ts
```diff
@@ -80,10 +87,17 @@ export class DateTimeWidgetComponent extends WidgetComponent implements OnInit,
         this.onDestroy$.complete();
     }
 
+    formatLabel(field: FormFieldModel): string {
```
rbahirsheth, line 90: Missed the unit test cases.

SheenaMalhotra182: added

lib/core/src/lib/form/services/date-format-translation.service.spec.ts
SheenaMalhotra182: removed

lib/core/src/lib/form/services/date-format-translation.service.ts
```diff
@@ -0,0 +1,60 @@
...
+    };
+
+    private dateFnsToMomentMap = {
+        M: 'M',
+        d: 'D',
+        y: 'Y',
+        a: 'A'
+    };
+
+    convertMomentToDateFnsFormat(dateDisplayFormat: string): string {
```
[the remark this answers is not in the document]
SheenaMalhotra182: added comments

lib/core/src/lib/pipes/date.pipe.ts
SheenaMalhotra182: renamed

lib/core/src/lib/form/components/widgets/core/form-field.model.spec.ts
SheenaMalhotra182: fixed import path

lib/core/src/lib/form/services/date-format-translation.service.ts
SheenaMalhotra182: updated to call new Date() inside parse itself.

lib/core/src/lib/form/components/widgets/date/date.widget.html
SheenaMalhotra182: implemented

lib/core/src/lib/pipes/date.pipe.spec.ts
SheenaMalhotra182: removed unnecessary imports

lib/core/src/lib/pipes/date.pipe.spec.ts
SheenaMalhotra182: fixed

lib/process-services-cloud/src/lib/form/components/widgets/date/date-cloud.widget.ts
SheenaMalhotra182: after dome changes now, DATE_FORMAT_CLOUD is 'YYYY-MM-DD' like before.
It also helps to maintain uniformity across all places if we use format or parse methods from DateFnsUtils.

lib/core/src/lib/pipes/date.pipe.spec.ts
SheenaMalhotra182: fixed

lib/core/src/lib/form/components/widgets/core/form-field-validator.ts
SheenaMalhotra182: moved all logic to DateFnsUtils

lib/core/src/lib/form/components/widgets/core/form-field-validator.ts
SheenaMalhotra182: added documentation

lib/core/src/lib/form/services/date-format-translation.service.ts
SheenaMalhotra182: moved all logic to DateFnsUtils

lib/core/src/lib/pipes/date.pipe.spec.ts
SheenaMalhotra182: added steps to check wth different date formats

lib/core/src/lib/pipes/pipe.module.ts
SheenaMalhotra182: renamed

lib/core/src/lib/form/components/widgets/core/form-field-validator.ts
SheenaMalhotra182: moved all logic to DateFnsUtils

SheenaMalhotra182:
Hi @DenysVuika a part of the review comments related to moving reusable code to 'DateFnsUtils' were addressed as a part of https://github.com/Alfresco/alfresco-ng2-components/pull/8934 which is merged now.

1. I have addressed other review comments received for 'DateWidgetComponent', and 'DateCloudWidgetComponent' 
2. and added changes for replacing occurrences of adfMomentDate with adfDate in 'DateTimeWidgetComponent' as well.

lib/core/src/lib/form/components/widgets/core/form-field-validator.spec.ts
```diff
@@ -1113,7 +1113,7 @@ describe('FormFieldValidator', () => {
         it('should validate dateTime format with default format', () => {
             const field = new FormFieldModel(new FormModel(), {
                 type: FormFieldTypes.DATETIME,
-                value: '2021-06-09 14:10'
+                value: '2021-06-09 02:10 AM'
```
DenysVuika, line 1116: you are changing the mock class, it should remain the same

lib/core/src/lib/form/components/widgets/date-time/date-time.widget.spec.ts
```diff
@@ -52,7 +52,7 @@ describe('DateTimeWidgetComponent', () => {
     });
 
     it('should setup min value for date picker', () => {
-        const minValue = '1982-03-13T10:00:000Z';
+        let minValue = '1982-03-13T10:00:000Z';
```
DenysVuika, line 55: should be const as you are not changing it

lib/core/src/lib/form/components/widgets/date-time/date-time.widget.spec.ts
```diff
@@ -79,14 +82,17 @@ describe('DateTimeWidgetComponent', () => {
     });
 
     it('should setup max value for date picker', () => {
-        const maxValue = '1982-03-13T10:00:000Z';
+        let maxValue = '1982-03-13T10:00:000Z';
```
DenysVuika, line 85: should be const as you are not changing it

lib/core/src/lib/form/components/widgets/core/form-field-validator.spec.ts
```diff
@@ -708,7 +708,7 @@ describe('FormFieldValidator', () => {
 
         it('should take into account that max value is in UTC and NOT fail validating value checking the time', () => {
             const maxValueFromActivitiInput = '31-3-2018 12:00 AM';
-            const maxValueSavedInForm = moment(maxValueFromActivitiInput, 'DD-M-YYYY hh:mm A').utc().format();
+            const maxValueSavedInForm = DateFnsUtils.formatDate(DateFnsUtils.parseDate(maxValueFromActivitiInput, 'DD-M-YYYY hh:mm A'), `YYYY-MM-DDTHH:mm:ssZ`);
```
DenysVuika, line 711: hardcoded format, should be using UTC via

```ts
const maxValueSavedInForm = DateFnsUtils.parseDate(maxValueFromActivitiInput, 'DD-M-YYYY hh:mm A').toISOString();
```

can you check if that works and update other test cases?

lib/core/src/lib/form/components/widgets/core/form-field-validator.ts
```diff
@@ -301,44 +302,44 @@ export class MinDateTimeFieldValidator implements FormFieldValidator {
     private supportedTypes = [
         FormFieldTypes.DATETIME
     ];
-    MIN_DATETIME_FORMAT = 'YYYY-MM-DD hh:mm AZ';
```
DenysVuika, line 304: why timezone is removed here?

lib/core/src/lib/form/components/widgets/date-time/date-time.widget.ts
```diff
@@ -17,73 +17,90 @@
...
+        const dateTimeDisplayName = this.translationService.instant(field.name);
+        const dateTimeDisplayFormat = DateFnsUtils.convertDateFnsToMomentFormat(field.dateDisplayFormat);
+
+        return `${dateTimeDisplayName} (${dateTimeDisplayFormat})`;
+    }
+
     onDateChanged(newDateValue) {
-        const date = moment(newDateValue, this.field.dateDisplayFormat, true);
-        if (date.isValid()) {
-            this.field.value = moment(date).utc().local().format(this.field.dateDisplayFormat);
```
DenysVuika, line 86: you are removing the "utc().local()" here, what is the "local" equivalent date/time is in the right side?

lib/core/src/lib/form/components/widgets/date/date.widget.ts
```diff
@@ -52,34 +55,30 @@ export class DateWidgetComponent extends WidgetComponent implements OnInit, OnDe
...
             .pipe(takeUntil(this.onDestroy$))
-            .subscribe(locale => this.dateAdapter.setLocale(locale));
+            .subscribe(locale => this.dateAdapter.setLocale(DateFnsUtils.getLocaleFromString(locale)));
 
-        const momentDateAdapter = this.dateAdapter as MomentDateAdapter;
-        momentDateAdapter.overrideDisplayFormat = this.field.dateDisplayFormat;
+        this.dateFormatConfig.display.dateInput = this.field.dateDisplayFormat;
 
         if (this.field) {
-            if (this.field.minValue) {
```
DenysVuika, line 76: you removed the check for null/undefined, if there's no particular reason for that please restore it

lib/core/src/lib/form/components/widgets/date/date.widget.ts
```diff
@@ -52,34 +55,30 @@ export class DateWidgetComponent extends WidgetComponent implements OnInit, OnDe
...
-                this.minDate = moment(this.field.minValue, this.DATE_FORMAT);
-            }
-
-            if (this.field.maxValue) {
```
DenysVuika, line 80: you removed the check for null/undefined, if there's no particular reason for that please restore it

lib/process-services-cloud/src/lib/form/components/widgets/date/date-cloud.widget.ts
```diff
@@ -51,62 +52,66 @@ import { DATE_FORMAT_CLOUD } from '../../../../models/date-format-cloud.model';
...
-                const today = this.getTodaysFormattedDate();
+                const today = new Date();
                 if (Number.isInteger(this.field.minDateRangeValue)) {
-                    this.minDate = moment(today).subtract(this.field.minDateRangeValue, 'days');
-                    this.field.minValue = this.minDate.format(DATE_FORMAT_CLOUD);
+                    this.minDate = DateFnsUtils.formatDate(subDays(today, this.field.minDateRangeValue), DATE_FORMAT_CLOUD);
```
DenysVuika, line 80: should be date instead of string

lib/core/src/lib/form/components/widgets/date-time/date-time.widget.spec.ts
```diff
@@ -101,7 +107,7 @@ describe('DateTimeWidgetComponent', () => {
         });
 
         widget.field = field;
-        const mockDate = moment('1982-03-13T10:00:000Z', 'YYYY-MM-DDTHH:mm:ssZ');
+        const mockDate = DateFnsUtils.formatDate(new Date('1982-03-13 10:00 AM'), 'YYYY-MM-DDTHH:mm:ssZ');
```
DenysVuika, line 110: Incorrect format for `new Date`, use the original `'1982-03-13T10:00:000Z'`

lib/core/src/lib/form/components/widgets/date-time/date-time.widget.spec.ts
```diff
@@ -25,6 +24,7 @@ import { TranslateModule } from '@ngx-translate/core';
 import { MatTooltipModule } from '@angular/material/tooltip';
 import { FormFieldTypes } from '../core/form-field-types';
 import { By } from '@angular/platform-browser';
+import { DateFnsUtils } from '../../../../common';
```
DenysVuika, line 27: incorrect import, should be exact class path here

lib/core/src/lib/form/components/widgets/core/form-field.model.spec.ts
```diff
@@ -15,10 +15,10 @@
  * limitations under the License.
  */
 
-import moment from 'moment';
 import { FormFieldTypes } from './form-field-types';
 import { FormFieldModel } from './form-field.model';
 import { FormModel } from './form.model';
+import { DateFnsUtils } from '../../../../common';
```
DenysVuika, line 21: incorrect import path

lib/core/src/lib/form/components/widgets/core/form-field-validator.ts
```diff
@@ -192,10 +193,10 @@ export class DateTimeFieldValidator implements FormFieldValidator {
 
     validate(field: FormFieldModel): boolean {
         if (this.isSupported(field) && field.value && field.isVisible) {
-            if (DateFieldValidator.isValidDate(field.value, field.dateDisplayFormat)) {
+            if (DateTimeFieldValidator.isValidDate(field.value, DateFnsUtils.convertMomentToDateFnsFormat(field.dateDisplayFormat))) {
```
DenysVuika, line 196: lot of code publication with this line.... the DateTimeFieldValidator should encapsulate the conversion . use the original code and update the validator instead: https://github.com/Alfresco/alfresco-ng2-components/blob/98c0a3c7bef00e6069a279ef47a01d2732784ee0/lib/core/src/lib/form/components/widgets/core/form-field-validator.ts#L180
the validator should internally call the "convertMomentTo..."

lib/core/src/lib/form/components/widgets/core/form-field-validator.ts
```diff
@@ -17,10 +17,11 @@
 
 /* eslint-disable @angular-eslint/component-selector */
 
-import moment from 'moment';
 import { FormFieldTypes } from './form-field-types';
 import { isNumberValue } from './form-field-utils';
 import { FormFieldModel } from './form-field.model';
+import { isAfter, isBefore, isValid } from 'date-fns';
+import { DateFnsUtils } from '../../../../common';
```
DenysVuika, line 24: incorrect import path

lib/core/src/lib/form/components/widgets/core/form-field-validator.spec.ts
```diff
@@ -15,6 +15,7 @@
  * limitations under the License.
  */
 
+import { DateFnsUtils } from '../../../../common';
```
DenysVuika, line 18: incorrect import path

lib/core/src/lib/form/components/widgets/core/form-field-validator.ts
```diff
@@ -347,45 +348,46 @@ export class MaxDateTimeFieldValidator implements FormFieldValidator {
     private supportedTypes = [
         FormFieldTypes.DATETIME
     ];
-    MAX_DATETIME_FORMAT = 'YYYY-MM-DD hh:mm AZ';
```
DenysVuika, line 350: dropping the time zone here?

lib/core/src/lib/form/components/widgets/core/form-field.model.ts
```diff
@@ -335,16 +348,30 @@ export class FormFieldModel extends FormWidgetModel {
...
          */
-        if (this.isDateField(json) || this.isDateTimeField(json)) {
+        if (this.isDateTimeField(json)) {
             if (value) {
                 let dateValue;
                 if (isNumberValue(value)) {
-                    dateValue = moment(value);
+                    dateValue = new Date(value);
                 } else {
-                    dateValue = this.isDateTimeField(json) ? moment.utc(value, 'YYYY-MM-DD hh:mm A') : moment.utc(value.split('T')[0], 'YYYY-M-D');
```
DenysVuika, line 344: is the UTC still preserved here?

lib/core/src/lib/form/components/widgets/core/form-field.model.ts
```diff
@@ -431,13 +459,14 @@ export class FormFieldModel extends FormWidgetModel {
             }
             case FormFieldTypes.DATETIME: {
                 if (typeof this.value === 'string' && this.value === 'now') {
-                    this.value = moment(new Date()).utc().format(this.dateDisplayFormat);
```
DenysVuika, line 434: is UTC preserved here? probably not

lib/core/src/lib/form/components/widgets/date-time/date-time.widget.spec.ts
SheenaMalhotra182: changed back to const

lib/core/src/lib/form/components/widgets/date-time/date-time.widget.spec.ts
SheenaMalhotra182: changed back to const

lib/core/src/lib/form/components/widgets/core/form-field-validator.spec.ts
SheenaMalhotra182: Yes, it worked with .toISOString(). I have updated it in other tests also,

lib/core/src/lib/form/components/widgets/core/form-field-validator.ts
SheenaMalhotra182: added timezone

lib/core/src/lib/form/components/widgets/date-time/date-time.widget.ts
SheenaMalhotra182: I think the utc().local() conversion from moment.js is not needed in date-fns, as JavaScript's Date object automatically handles local time and UTC conversion based on the system's time zone settings.

lib/core/src/lib/form/components/widgets/date/date.widget.ts
SheenaMalhotra182: added null/undefined check

lib/core/src/lib/form/components/widgets/date/date.widget.ts
SheenaMalhotra182: added back

lib/process-services-cloud/src/lib/form/components/widgets/date/date-cloud.widget.ts
```diff
@@ -51,62 +52,66 @@ import { DATE_FORMAT_CLOUD } from '../../../../models/date-format-cloud.model';
 export class DateCloudWidgetComponent extends WidgetComponent implements OnInit, OnDestroy {
     typeId = 'DateCloudWidgetComponent';
 
-    minDate: Moment;
-    maxDate: Moment;
+    minDate: string;
```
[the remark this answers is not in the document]
SheenaMalhotra182: changed to Date

lib/process-services-cloud/src/lib/form/components/widgets/date/date-cloud.widget.ts
SheenaMalhotra182: changed to Date

lib/process-services-cloud/src/lib/form/components/widgets/date/date-cloud.widget.ts
```diff
@@ -51,62 +52,66 @@ import { DATE_FORMAT_CLOUD } from '../../../../models/date-format-cloud.model';
...
                 if (Number.isInteger(this.field.minDateRangeValue)) {
-                    this.minDate = moment(today).subtract(this.field.minDateRangeValue, 'days');
-                    this.field.minValue = this.minDate.format(DATE_FORMAT_CLOUD);
+                    this.minDate = DateFnsUtils.formatDate(subDays(today, this.field.minDateRangeValue), DATE_FORMAT_CLOUD);
+                    this.field.minValue = this.minDate;
                 }
                 if (Number.isInteger(this.field.maxDateRangeValue)) {
-                    this.maxDate = moment(today).add(this.field.maxDateRangeValue, 'days');
-                    this.field.maxValue = this.maxDate.format(DATE_FORMAT_CLOUD);
+                    this.maxDate = DateFnsUtils.formatDate(addDays(today, this.field.maxDateRangeValue), DATE_FORMAT_CLOUD);
```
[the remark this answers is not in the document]
SheenaMalhotra182: changed to Date

lib/core/src/lib/form/components/widgets/date-time/date-time.widget.ts
```diff
@@ -17,73 +17,90 @@
...
     selector: 'date-time-widget',
     templateUrl: './date-time.widget.html',
     styleUrls: ['./date-time.widget.scss'],
     encapsulation: ViewEncapsulation.None
 })
 export class DateTimeWidgetComponent extends WidgetComponent implements OnInit, OnDestroy {
-
-    minDate: Moment;
-    maxDate: Moment;
+    minDate: string;
```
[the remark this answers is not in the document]
SheenaMalhotra182: changed to Date

lib/core/src/lib/form/components/widgets/date-time/date-time.widget.spec.ts
SheenaMalhotra182: using '1982-03-13T10:00:000Z' now

lib/core/src/lib/form/components/widgets/date-time/date-time.widget.spec.ts
SheenaMalhotra182: fixed

lib/core/src/lib/form/components/widgets/core/form-field.model.spec.ts
SheenaMalhotra182: fixed

lib/core/src/lib/form/components/widgets/core/form-field-validator.ts
SheenaMalhotra182: convertToMoment is already being handled in isValidDate() so removed from here

lib/core/src/lib/form/components/widgets/core/form-field-validator.ts
SheenaMalhotra182: fixed

lib/core/src/lib/form/components/widgets/core/form-field-validator.spec.ts
SheenaMalhotra182: fixed

lib/core/src/lib/form/components/widgets/core/form-field-validator.ts
SheenaMalhotra182: added timezone.

lib/core/src/lib/form/components/widgets/core/form-field.model.ts
SheenaMalhotra182: updated code to use parseISO() to preserve UTC info.

lib/core/src/lib/form/components/widgets/core/form-field.model.ts
SheenaMalhotra182: updated code to use parseISO() to preserve UTC info.

lib/core/src/lib/form/components/widgets/core/form-field-validator.spec.ts
```diff
@@ -708,9 +708,9 @@ describe('FormFieldValidator', () => {
 
         it('should take into account that max value is in UTC and NOT fail validating value checking the time', () => {
             const maxValueFromActivitiInput = '31-3-2018 12:00 AM';
-            const maxValueSavedInForm = moment(maxValueFromActivitiInput, 'DD-M-YYYY hh:mm A').utc().format();
+            const maxValueSavedInForm = DateFnsUtils.parseDate(maxValueFromActivitiInput, 'DD-M-YYYY hh:mm A').toISOString();
 
-            const localValidValue = '2018-3-30 11:59 PM';
```
DenysVuika, line 713: you are again changing the test data coming  from backend, do not touch this value

lib/core/src/lib/form/components/widgets/core/form-field-validator.spec.ts
```diff
@@ -723,9 +723,9 @@ describe('FormFieldValidator', () => {
 
         it('should take into account that max value is in UTC and fail validating value checking the time', () => {
             const maxValueFromActivitiInput = '31-3-2018 12:00 AM';
-            const maxValueSavedInForm = moment(maxValueFromActivitiInput, 'DD-M-YYYY hh:mm A').utc().format();
+            const maxValueSavedInForm = DateFnsUtils.parseDate(maxValueFromActivitiInput, 'DD-M-YYYY hh:mm A').toISOString();
 
-            const localInvalidValue = '2018-3-31 12:01 AM';
```
DenysVuika, line 728: modifying server related value

lib/process-services-cloud/src/lib/form/components/widgets/date/date-cloud.widget.spec.ts
```diff
@@ -424,15 +425,15 @@ describe('DateWidgetComponent', () => {
                 fixture.detectChanges();
                 await fixture.whenStable();
 
-                const expectedMaxValueString = '2022-07-30';
```
DenysVuika, line 427: modifying test data, it should be the same

lib/process-services-cloud/src/lib/form/components/widgets/date/date-cloud.widget.spec.ts
```diff
@@ -424,15 +425,15 @@ describe('DateWidgetComponent', () => {
...
+                const currentDate = new Date();
+                const expectedMaxValueString = DateFnsUtils.formatDate(addDays(currentDate, 8), DATE_FORMAT_CLOUD);
```
DenysVuika, line 429: this makes no sense, the expected value was already provided, you don't need to convert anything

lib/process-services-cloud/src/lib/form/components/widgets/date/date-cloud.widget.spec.ts
```diff
@@ -463,8 +463,9 @@ describe('DateWidgetComponent', () => {
                 fixture.detectChanges();
                 await fixture.whenStable();
 
-                const expectedMaxValueString = '2022-07-30';
```
DenysVuika, line 466: test data, should not be modified

lib/core/src/lib/form/components/widgets/core/form-field-validator.spec.ts
```diff
@@ -833,9 +833,9 @@ describe('FormFieldValidator', () => {
 
         it('should take into account that min value is in UTC and NOT fail validating value checking the time', () => {
             const minValueFromActivitiInput = '02-3-2018 06:00 AM';
-            const minValueSavedInForm = moment(minValueFromActivitiInput, 'DD-M-YYYY hh:mm A').utc().format();
+            const minValueSavedInForm = DateFnsUtils.parseDate(minValueFromActivitiInput, 'DD-M-YYYY hh:mm A').toISOString();
 
-            const localValidValue = '2018-3-02 06:01 AM';
```
DenysVuika, line 838: the test data should not be changed, it's the activity form field model... the purpose of validator is to do the magic inside and validate what's coming into...

lib/core/src/lib/form/components/widgets/date/date.widget.spec.ts
```diff
@@ -170,23 +171,23 @@ describe('DateWidgetComponent', () => {
                 readOnly: 'false'
             });
             widget.field.isVisible = true;
-            widget.field.dateDisplayFormat = 'MM-DD-YYYY';
```
DenysVuika, line 173: modified test data

lib/core/src/lib/form/components/widgets/date/date.widget.spec.ts
```diff
@@ -170,23 +171,23 @@ describe('DateWidgetComponent', () => {
...
+            widget.field.dateDisplayFormat = DateFnsUtils.convertMomentToDateFnsFormat('MM-DD-YYYY');
 
             fixture.detectChanges();
 
             let dateElement: any = element.querySelector('#date-field-id');
             expect(dateElement.value).toContain('12-30-9999');
 
             widget.field.value = '5-6-2019 00:00';
-            widget.field.dateDisplayFormat = 'D-M-YYYY HH:mm';
```
DenysVuika, line 181: this is data coming from the backend

lib/core/src/lib/form/components/widgets/core/form-field-validator.spec.ts
SheenaMalhotra182: We changed this to accomodate the suggested parseISO() approach used in form-field-model.ts https://github.com/Alfresco/alfresco-ng2-components/pull/8924/commits/e6e0b27b803ba48bebc26729f9185112aceb5e5c#diff-231637c71df1cd50fe2ef9ba319639d7eeb28657b17f552fc979be0b9f508cd0R357.

We were not attempting to change the data here, but parsing it so that it's compatible with .ts logic.

In the previous approach where we were using parseDate(), where we did not need any changes in the test case https://github.com/Alfresco/alfresco-ng2-components/pull/8924/commits/e6e0b27b803ba48bebc26729f9185112aceb5e5c#diff-231637c71df1cd50fe2ef9ba319639d7eeb28657b17f552fc979be0b9f508cd0L357

upon revisiting this I think, the previous approach was better as no modifications were needed in the test case then.

DenysVuika: `const localInvalidValue = '2018-3-31 12:01 AM';` means invalid date as string, and you are parsing it and converting to iso

lib/core/src/lib/form/components/widgets/date/date.widget.ts
```diff
@@ -52,33 +55,34 @@ export class DateWidgetComponent extends WidgetComponent implements OnInit, OnDe
...
-    maxDate: Moment;
+    minDate: Date;
+    maxDate: Date;
 
     private onDestroy$ = new Subject<boolean>();
 
     constructor(public formService: FormService,
-                private dateAdapter: DateAdapter<Moment>,
-                private userPreferencesService: UserPreferencesService) {
+                private dateAdapter: DateAdapter<DateFnsAdapter>,
```
DenysVuika, line 64: incorrect parameter, should be the value type, not another adapter,
valid cases are `DateAdapter<Moment | Date | string>` etc

DenysVuika:
closing as too many issues

PR #8936, opened 2023-09-26
[ACS-6003] - waitRendition in rendition.service.ts not working

DominikIwanek:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

the Retry feature from method: getRendition does not work, because of wrong condition in rendition service

https://alfresco.atlassian.net/browse/ACS-6003

Calling the getRendition waits until the rendition is available.

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/content-services/src/lib/common/services/rendition.service.ts
+++ b/lib/content-services/src/lib/common/services/rendition.service.ts
@@ -88,9 +88,8 @@ export class RenditionService {
     }
 
     private async waitRendition(nodeId: string, renditionId: string, retries: number): Promise<RenditionEntry> {
-        const rendition = await this.renditionsApi.getRendition(nodeId, renditionId);
-
-        if (this.maxRetries >= retries) {
+        if (this.maxRetries > retries) {
+            const rendition = await this.renditionsApi.getRendition(nodeId, renditionId);
             const status = rendition.entry.status.toString();
 
             if (status === 'CREATED') {
--- a/lib/content-services/src/lib/common/services/rendition.service.ts
+++ b/lib/content-services/src/lib/common/services/rendition.service.ts
@@ -90,7 +90,7 @@ export class RenditionService {
     private async waitRendition(nodeId: string, renditionId: string, retries: number): Promise<RenditionEntry> {
         const rendition = await this.renditionsApi.getRendition(nodeId, renditionId);
 
-        if (this.maxRetries < retries) {
+        if (this.maxRetries >= retries) {
             const status = rendition.entry.status.toString();
 
             if (status === 'CREATED') {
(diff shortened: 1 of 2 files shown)


DenysVuika (approved):

MichalKinas (approved):

swapnil-verma-gl (approved):

lib/content-services/src/lib/common/services/rendition.service.ts
```diff
@@ -90,7 +90,7 @@ export class RenditionService {
     private async waitRendition(nodeId: string, renditionId: string, retries: number): Promise<RenditionEntry> {
         const rendition = await this.renditionsApi.getRendition(nodeId, renditionId);
 
-        if (this.maxRetries < retries) {
+        if (this.maxRetries >= retries) {
```
AleksanderSklorz, line 93: can you write unit test which will prevent that issue in the future?

DominikIwanek: @AleksanderSklorz sure, I have added a test in new commit

lib/content-services/src/lib/common/services/rendition.services.spec.ts
```diff
@@ -0,0 +1,105 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { TestBed } from '@angular/core/testing';
+import { AlfrescoApiService, LogService, TranslationService, ViewUtilService } from '@alfresco/adf-core';
+import { RenditionEntry, RenditionPaging, RenditionsApi } from '@alfresco/js-api';
+import { RenditionService } from '@alfresco/adf-content-services';
+
+const getRenditionEntry = (status: 'NOT_CREATED' | 'CREATED'): RenditionEntry => {
```
AleksanderSklorz, line 23: In js-api there is reusable const for these strings. Could you please use that const instead of hardcoded strings in whole file? It would be better to avoid typo in future etc. 
image

AleksanderSklorz (approved):

lib/content-services/src/lib/common/services/rendition.services.spec.ts
```diff
@@ -0,0 +1,105 @@
...
+                        listRenditions: jasmine.createSpy('listRenditions'),
+                        getRendition: jasmine.createSpy('getRendition'),
+                        createRendition: jasmine.createSpy('createRendition')
+                    } }
+            ]
+        });
+        renditionService = TestBed.inject(RenditionService);
+        renditionsApi = TestBed.inject(RenditionsApi);
+        // eslint-disable-next-line no-underscore-dangle
+        renditionService._renditionsApi = renditionsApi;
```
DenysVuika, line 61: this is a hack, the `_renditionsApi` should be private, you can use `spyOnProperty(renditionService, 'renditionsApi').and.returnValue(renditionsApi)`

DenysVuika (changes requested):

lib/content-services/src/lib/common/services/rendition.services.spec.ts
DenysVuika: otherwise, you are already using provider with a custom Renditions api, so it should automatically be available

PR #8942, opened 2023-09-28
[ACS-6071] fix JSDoc issues for Core lib

DenysVuika:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [x] Documentation

prior to enabling JSDoc linting, fix errors and warnigns for the Core lib

> - [ ] Yes
> - [x] No

**Other information**:
https://alfresco.atlassian.net/browse/ACS-6071

### DIFF
--- a/lib/core/src/lib/auth/services/user-access.service.spec.ts
+++ b/lib/core/src/lib/auth/services/user-access.service.spec.ts
@@ -36,6 +36,12 @@ describe('UserAccessService', () => {
         appConfigService = TestBed.inject(AppConfigService);
     });
 
+    /**
+     * spy on auth realm access
+     *
+     * @param realmRoles roles
+     * @param resourceAccess access settings
+     */
     function spyRealmAccess(realmRoles: string[], resourceAccess: any) {
         spyOn(jwtHelperService, 'getAccessToken').and.returnValue('my-access_token');
         spyOn(jwtHelperService, 'decodeToken').and.returnValue({
@@ -44,6 +50,12 @@ describe('UserAccessService', () => {
         });
     }
 
+    /**
+     * spy on HxP authorisation
+     *
+     * @param appkey app key
+     * @param roles roles list
+     */
     function spyHxpAuthorization(appkey: string, roles: string[]) {
         spyOn(jwtHelperService, 'getAccessToken').and.returnValue('my-access_token');
         spyOn(jwtHelperService, 'decodeToken').and.returnValue({
--- a/lib/core/src/lib/common/services/directionality-config-factory.ts
+++ b/lib/core/src/lib/common/services/directionality-config-factory.ts
@@ -18,6 +18,12 @@
 import { DirectionalityConfigService } from './directionality-config.service';
 
 // eslint-disable-next-line prefer-arrow/prefer-arrow-functions
+/**
+ * Creates a factory for Angular modules
+ *
+ * @param directionalityConfigService service dependency
+ * @returns a factory function
+ */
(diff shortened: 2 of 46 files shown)


lib/core/src/lib/form/components/form-renderer.component.ts
```diff
@@ -78,6 +78,9 @@ export class FormRendererComponent<T> implements OnChanges, OnDestroy {
 
     /**
      * Serializes column fields
+     *
+     * @param content container model
+     * @returns a lit of form field models
```
AleksanderSklorz, line 83: 'lit' typo

lib/core/src/lib/identity-user-info/identity-user-info.component.ts
```diff
@@ -54,11 +54,12 @@ export class IdentityUserInfoComponent implements OnDestroy {
...
     showName: boolean = true;
 
-    /** When the username is shown, this defines its position relative to the user info button.
+    /**
+     * When the username is shown, this defines its position relative to the user info button.
      * Can be `right` or `left`.
      */
     @Input()
-    namePosition: string = 'right';
+    namePosition: 'right' | 'left';
```
AleksanderSklorz, line 62: Is not that breaking change? Is that expected? In previous version it has 'right' as default value, now it will be undefined I guess?

DenysVuika: good catch, fixed

lib/core/src/lib/form/components/form-renderer.component.ts
DenysVuika: thanks, fixed

lib/core/src/lib/auth/services/user-access.service.spec.ts
```diff
@@ -36,6 +36,12 @@ describe('UserAccessService', () => {
         appConfigService = TestBed.inject(AppConfigService);
     });
 
+    /**
```
AleksanderSklorz, line 39: I'm wondering if we need documentation for functions in spec files as they are mostly used by tests for example to reduce repeated code and won't be used by external people like in public feature code. But if we want documentation for test functions then it's fine to me.

AleksanderSklorz (approved):

lib/core/src/lib/common/services/directionality-config-factory.ts
```diff
@@ -18,6 +18,12 @@
 import { DirectionalityConfigService } from './directionality-config.service';
 
 // eslint-disable-next-line prefer-arrow/prefer-arrow-functions
```
nikita-web-ua, line 20: Probably should be before the function?

nikita-web-ua (approved):

lib/core/src/lib/auth/services/user-access.service.spec.ts
DenysVuika: just following the default docs setup for the eslint, we can later improve the configuration

lib/core/src/lib/common/services/directionality-config-factory.ts
DenysVuika: it's a comment block, so should be valid, the eslint positioned it like that automatically

lib/core/src/lib/auth/services/user-access.service.spec.ts
AleksanderSklorz: Ok @DenysVuika fine. For future we can maybe try something similarly like we have in ACA: 

image


You can see there is excludeFields field so it allows to exclude some files from specific rule.

PR #8970, opened 2023-10-06
migrate node lock dialog to date-fns

DenysVuika:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

- migrate node-lock dialog to date-fns
- clean useless tests

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/content-services/src/lib/dialogs/node-lock.dialog.spec.ts
+++ b/lib/content-services/src/lib/dialogs/node-lock.dialog.spec.ts
@@ -22,8 +22,7 @@ import { ContentTestingModule } from '../testing/content.testing.module';
 import { TranslateModule } from '@ngx-translate/core';
 import { addMinutes } from 'date-fns';
 
-// eslint-disable-next-line ban/ban, @cspell/spellchecker
-fdescribe('NodeLockDialogComponent', () => {
+describe('NodeLockDialogComponent', () => {
 
     let fixture: ComponentFixture<NodeLockDialogComponent>;
     let component: NodeLockDialogComponent;
--- a/lib/content-services/src/lib/dialogs/node-lock.dialog.spec.ts
+++ b/lib/content-services/src/lib/dialogs/node-lock.dialog.spec.ts
@@ -15,19 +15,20 @@
  * limitations under the License.
  */
 
-import moment from 'moment';
-
 import { TestBed, fakeAsync, tick, ComponentFixture } from '@angular/core/testing';
 import { MatDialogRef } from '@angular/material/dialog';
 import { NodeLockDialogComponent } from './node-lock.dialog';
 import { ContentTestingModule } from '../testing/content.testing.module';
 import { TranslateModule } from '@ngx-translate/core';
+import { addMinutes } from 'date-fns';
 
-describe('NodeLockDialogComponent', () => {
+// eslint-disable-next-line ban/ban, @cspell/spellchecker
+fdescribe('NodeLockDialogComponent', () => {
 
     let fixture: ComponentFixture<NodeLockDialogComponent>;
     let component: NodeLockDialogComponent;
-    let expiryDate;
+    let expiryDate: Date;
+
     const dialogRef = {
         close: jasmine.createSpy('close')
     };
@@ -54,7 +55,7 @@ describe('NodeLockDialogComponent', () => {
 
         beforeEach(() => {
             jasmine.clock().mockDate(new Date());
-            expiryDate = moment().add(1, 'minutes');
+            expiryDate = addMinutes(new Date(), 1);
 
             component.data = {
                 node: {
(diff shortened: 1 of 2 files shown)


MateuszHY (approved):

lib/content-services/src/lib/dialogs/node-lock.dialog.spec.ts
```diff
@@ -15,19 +15,20 @@
...
 import { TestBed, fakeAsync, tick, ComponentFixture } from '@angular/core/testing';
 import { MatDialogRef } from '@angular/material/dialog';
 import { NodeLockDialogComponent } from './node-lock.dialog';
 import { ContentTestingModule } from '../testing/content.testing.module';
 import { TranslateModule } from '@ngx-translate/core';
+import { addMinutes } from 'date-fns';
 
-describe('NodeLockDialogComponent', () => {
+// eslint-disable-next-line ban/ban, @cspell/spellchecker
+fdescribe('NodeLockDialogComponent', () => {
```
nikita-web-ua, line 26: fdescribe on purpose?

lib/content-services/src/lib/dialogs/node-lock.dialog.ts
```diff
@@ -55,18 +53,20 @@ export class NodeLockDialogComponent implements OnInit {
         const { node } = this.data;
         this.nodeName = node.name;
 
+        const isTimeLock = !!node.properties['cm:expiryDate'];
+        const time = isTimeLock ? new Date(node.properties['cm:expiryDate']) : new Date();
```
swapnil-verma-gl, line 57: The one problem I can see here when we don't specify a format to work with would be how it would distinguish between dates and months when they are both less than 12. For e.g., is 04/03/2023 treated as 4th March, 2023, or 3rd April 2023.?
The older code also would probably have this bug, since no format is provided there either

lib/content-services/src/lib/dialogs/node-lock.dialog.ts
```diff
@@ -35,7 +33,7 @@ export class NodeLockDialogComponent implements OnInit {
     node: Node = null;
     nodeName: string;
 
-    _nodesApi: NodesApi;
+    private _nodesApi: NodesApi;
     get nodesApi(): NodesApi {
         this._nodesApi = this._nodesApi ?? new NodesApi(this.alfrescoApi.getInstance());
```
swapnil-verma-gl, line 38: Just curious, but wouldn't it be better to just have 
```
private _nodesApi;;
get nodesApi(): NodesApi {
    return this._nodesApi;
}
constructor (alfrescoApi: AlfrescoApiService, ...) {
   this._nodesApi = new NodesApi(this.alfrescoApi.getInstance());
}
```
This way we can avoid the check for the existence of _nodesApi and the needless assignment of `this._nodesApi = this._nodesApi`  everytime the getter is called?

Since this seems like the only usage of the alfrescoApi object in this class, we would also not needlessly have to keep the object of AlfrescoApiService in memory  while this component is active (see lack of 'private' keyword in constructor argument).

lib/content-services/src/lib/dialogs/node-lock.dialog.spec.ts
```diff
@@ -15,19 +15,20 @@
...
-
 import { TestBed, fakeAsync, tick, ComponentFixture } from '@angular/core/testing';
 import { MatDialogRef } from '@angular/material/dialog';
 import { NodeLockDialogComponent } from './node-lock.dialog';
 import { ContentTestingModule } from '../testing/content.testing.module';
 import { TranslateModule } from '@ngx-translate/core';
+import { addMinutes } from 'date-fns';
 
-describe('NodeLockDialogComponent', () => {
+// eslint-disable-next-line ban/ban, @cspell/spellchecker
```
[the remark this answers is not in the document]
DenysVuika: sorry, my bad

lib/content-services/src/lib/dialogs/node-lock.dialog.ts
DenysVuika: this is a deferred/lazy init, the reason is when Angular builds all singletons for services, JS-API is not yet initialised

DenysVuika: so we build the NodesApi instance on demand, when UI component is rendered, and that means auth and init was done before

swapnil-verma-gl: Ah, that makes sense. Thanks!!

lib/content-services/src/lib/dialogs/node-lock.dialog.ts
DenysVuika: we need better typing support for JS-API, actually it comes already as a Date object from what I remember, so you don't need to format it (already includes time zones, etc)

nikita-web-ua (approved):

AleksanderSklorz (approved):

PR #8983, opened 2023-10-09
[APPS-2108] ADF date-fns adapter implementation

DenysVuika:

> - [ ] Bugfix
> - [x] Feature
> - [ ] Build related changes
> - [ ] Documentation

- introduce ADF date-fns adapter 
  - handle moment and date-fns formats
  - automatically switch locales based on user preferences
- migrate Date Widget to date-fns
  - fix unit tests to always have correct Date value for fields (as per ISO and backend)
  - extra tests for min/max validation
  - tests now use proper form validators 
 - remove protractor e2e for date widget that are already covered by unit tests

Screenshot 2023-10-09 at 19 58 15

> - [ ] Yes
> - [x] No

**Other information**:
https://alfresco.atlassian.net/browse/APPS-2108


lib/core/src/lib/form/components/widgets/date/date.widget.spec.ts
```diff
@@ -197,13 +199,13 @@ describe('DateWidgetComponent', () => {
             fixture.detectChanges();
 
             let dateButton = element.querySelector<HTMLButtonElement>('button');
-            expect(dateButton.disabled).toBeFalsy();
+            expect(dateButton?.disabled).toBeFalsy();
```
swapnil-verma-gl, line 202: Maybe instead of 
```
expect(dateButton?.disabled).toBeFalsy();
```
we should have 
```
expect(dateButton).toBeDefined();
expect(dateButton.disabled).toBeFalsy();
```
It should cover the case of when dateButton is undefined/null. I tried running the following test case, and it passed
```
    fit('temp unit test', () => {
       let obj = null;
       expect(obj?.disabled).toBeFalsy();
       obj = undefined;
       expect(obj?.disabled).toBeFalsy();
    });
```

lib/core/src/lib/common/utils/date-fns-adapter.ts
```diff
@@ -0,0 +1,88 @@
...
+    get displayFormat(): string | null {
+        return this._displayFormat;
+    }
+
+    set displayFormat(value: string | null) {
+        this._displayFormat = value ? DateFnsUtils.convertMomentToDateFnsFormat(value) : null;
+    }
+
+    constructor(
+        @Optional() @Inject(MAT_DATE_LOCALE) matDateLocale: any,
```
MichalKinas, line 61: Can we avoid using `any` here?

lib/core/src/lib/form/components/widgets/date/date.widget.spec.ts
DenysVuika: thanks, will try... it's not about test actually, but the most recent typescript flagging it was "obj is potentially null"

lib/core/src/lib/common/utils/date-fns-adapter.ts
DenysVuika: I'll see what is the options, took it from here https://github.com/angular/components/blob/main/src/material-date-fns-adapter/adapter/date-fns-adapter.ts#L55 but latest TS flags empty object as a warning, and in moment adapter Angular is using "any"... but maybe I can find a type

lib/core/src/lib/form/components/widgets/date/date.widget.spec.ts
DenysVuika: the ESLint converts "expect(dateButton && dateButton.disabled)" to the same short form I used

MichalKinas (approved):

lib/core/src/lib/common/utils/date-fns-adapter.ts
DenysVuika: used `Locale` type from date-fns as that's what expected

swapnil-verma-gl (approved):

AleksanderSklorz (approved):

AleksanderSklorz (approved):

MichalKinas (approved):

swapnil-verma-gl (approved):

Issue #4627, opened 2019-04-19
provide data column type as number as an option


Abdul-Quadar-Quazi:
**Type of issue:**  (check with "[x]")

> - [x] New feature request
> - [ ] Bug  
> - [ ] Support request
> - [ ] Documentation

**Current behaviour:**
currently, the data column is taking numbers as a text and sorting as a string. which is causing an issue in sorting the number based column 

**Expected behavior:**
the data column should sort correctly for numeric values.

**New feature request:**
the column should provide a type = number and sort accordingly based on numeric sorting,
and please provide a custom sorting option so that we can use the sorting based on the requirement.

tomgny:
I think these PRs should solve the problem:

- https://github.com/Alfresco/alfresco-ng2-components/pull/8984
- https://github.com/Alfresco/alfresco-ng2-components/pull/9013

PR #8984, opened 2023-10-10
[AAE-16964] Handle amount data table cloumn type

tomgny:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://alfresco.atlassian.net/browse/AAE-16964

Amount column type
image

> - [ ] Yes
> - [ ] No

**Other information**:

### DIFF
--- a/lib/core/src/lib/datatable/components/amount-cell/amount-cell.component.html
+++ b/lib/core/src/lib/datatable/components/amount-cell/amount-cell.component.html
@@ -1,0 +1,10 @@
+<ng-container *ngIf="value$ | async as amount">
+    <span [title]="amount">
+        {{ amount | currency:
+            (currencyConfig?.code || defaultCurrencyConfig.code):
+            (currencyConfig?.display || defaultCurrencyConfig.display):
+            (currencyConfig?.digitsInfo || defaultCurrencyConfig.digitsInfo):
+            (currencyConfig?.locale || defaultCurrencyConfig.locale)
+        }}
+    </span>
+</ng-container>
--- a/lib/core/src/lib/datatable/components/amount-cell/amount-cell.component.ts
+++ b/lib/core/src/lib/datatable/components/amount-cell/amount-cell.component.ts
@@ -44,7 +44,7 @@ export class AmountCellComponent extends DataTableCellComponent implements OnIni
     @Input()
     currencyConfig: CurrencyConfig;
 
-    readonly defaultCurrencyConfig = {
+    readonly defaultCurrencyConfig: CurrencyConfig = {
         code: this.defaultCurrencyCode,
         display: 'symbol',
         digitsInfo: undefined,
--- a/lib/core/src/lib/datatable/components/amount-cell/amount-cell.component.ts
+++ b/lib/core/src/lib/datatable/components/amount-cell/amount-cell.component.ts
@@ -1,0 +1,66 @@
+/*!
+ * @license
+ * Copyright © 2005-2023 Hyland Software, Inc. and its affiliates. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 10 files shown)


lib/core/src/lib/datatable/components/amount-cell/amount-cell.component.ts
```diff
@@ -0,0 +1,66 @@
...
+import { DataTableService } from '../../services/datatable.service';
+import { CurrencyConfig } from '../../data/data-column.model';
+import { CommonModule } from '@angular/common';
+
+@Component({
+    standalone: true,
+    imports: [CommonModule],
+    selector: 'adf-amount-cell',
+    templateUrl: './amount-cell.component.html',
+    host: { class: 'adf-datatable-content-cell' },
```
wiktord2000, line 38: WDYT about add another class to the host 
```suggestion
    host: { class: 'adf-amount-cell adf-datatable-content-cell' },
```

lib/core/src/lib/datatable/components/amount-cell/amount-cell.component.html
```diff
@@ -0,0 +1,10 @@
+<ng-container *ngIf="value$ | async as amount">
+    <span [title]="amount">
```
wiktord2000, line 2: WDYT about add also the following class
```suggestion
    <span [title]="amount" class="adf-amount-cell-value">
```

lib/core/src/lib/datatable/components/amount-cell/amount-cell.component.ts
tomgny: as far as i know, we don't need to have additional styles for this component, correct me if im wrong

lib/core/src/lib/datatable/components/amount-cell/amount-cell.component.html
tomgny: same as above https://github.com/Alfresco/alfresco-ng2-components/pull/8984#discussion_r1352075874

eromano (commented):
Do we need to update the documentation?

tomgny:
> Do we need to update the documentation?
https://alfresco.atlassian.net/browse/AAE-17190
I added a task to our epic to update how-to confluence documentation with all new types that we will handle. Based on that document, doc team will update the documentation (correct me if something changed and we should update the docs)

lib/core/src/lib/datatable/components/amount-cell/amount-cell.component.ts
wiktord2000: not but I notice this pattern in the current cell, maybe it's better for testing instead of querying 'span' we can query 'adf-amount-cell-value' (according to the previous comment)

lib/core/src/lib/datatable/components/amount-cell/amount-cell.component.html
```diff
@@ -0,0 +1,10 @@
+<ng-container *ngIf="value$ | async as amount">
```
wiktord2000, line 1: How this solution handle wrong values like a objects? The angular pipe is prepared for strings&numbers only.

lib/core/src/lib/datatable/components/amount-cell/amount-cell.component.ts
tomgny: i don't think is a good practice to use fake class only for test purposes. in that component we have only one element, so it's no problem to query it as a span. btw, from e2e point of view, we have a data automation id added at the datatable component level, which contains type of widget (amount) + value, so for each row we have a different values. from this point, there is no benefit to add it also here if it is the point (testing)
image

eromano:
the doc in ADF is part of the definition of done is our duty upgrade it

tomgny:
> the doc in ADF is part of the definition of done is our duty upgrade it

Doc updated https://github.com/Alfresco/alfresco-ng2-components/pull/8984/commits/7084c3c936b27f7a898380898fdc998635017bc5

lib/core/src/lib/datatable/components/amount-cell/amount-cell.component.ts
wiktord2000: Ok, agree

lib/core/src/lib/datatable/components/amount-cell/amount-cell.component.html
tomgny: this is default error behaviour - display nothing
example:
(first row has incorrect value of car price)
image

wiktord2000 (approved):

DenysVuika:
does this also address https://github.com/Alfresco/alfresco-ng2-components/issues/4627 ?

PR #8986, opened 2023-10-10
[ACS-5611] Add the option to initially expand custom panel

MichalKinas:

> - [ ] Bugfix
> - [x] Feature
> - [ ] Build related changes
> - [ ] Documentation

Added the option to expand custom metadata panel initially. https://alfresco.atlassian.net/browse/ACS-5611

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/content-services/src/lib/content-metadata/components/content-metadata/content-metadata.component.html
+++ b/lib/content-services/src/lib/content-metadata/components/content-metadata/content-metadata.component.html
@@ -78,7 +78,7 @@
                 </adf-categories-management>
             </div>
         </ng-container>
-        <mat-expansion-panel *ngFor="let customPanel of customPanels">
+        <mat-expansion-panel *ngFor="let customPanel of customPanels" [expanded]="canExpandTheCard(customPanel.panelTitle)">
             <mat-expansion-panel-header>
                 <mat-panel-title>{{ customPanel.panelTitle | translate }}</mat-panel-title>
             </mat-expansion-panel-header>
@@ -90,7 +90,7 @@
                      class="adf-metadata-grouped-properties-container">
                     <mat-expansion-panel *ngIf="showGroup(group) || editable"
                                          [attr.data-automation-id]="'adf-metadata-group-' + group.title"
-                                         [expanded]="canExpandTheCard(group) || !displayDefaultProperties && first">
+                                         [expanded]="canExpandTheCard(group.title) || !displayDefaultProperties && first">
                         <mat-expansion-panel-header>
                             <mat-panel-title>
                                 {{ group.title | translate }}
--- a/lib/content-services/src/lib/content-metadata/components/content-metadata/content-metadata.component.ts
+++ b/lib/content-services/src/lib/content-metadata/components/content-metadata/content-metadata.component.ts
@@ -296,8 +296,8 @@ export class ContentMetadataComponent implements OnChanges, OnInit, OnDestroy {
         return properties.length > 0;
     }
 
-    canExpandTheCard(group: CardViewGroup): boolean {
-        return group.title === this.displayAspect;
+    canExpandTheCard(groupTitle: string): boolean {
+        return groupTitle === this.displayAspect;
     }
 
     canExpandProperties(): boolean {


DominikIwanek (approved):

lib/content-services/src/lib/content-metadata/components/content-metadata/content-metadata.component.html
```diff
@@ -78,7 +78,7 @@
                 </adf-categories-management>
             </div>
         </ng-container>
-        <mat-expansion-panel *ngFor="let customPanel of customPanels">
+        <mat-expansion-panel *ngFor="let customPanel of customPanels" [expanded]="canExpandTheCard(customPanel.panelTitle)">
```
swapnil-verma-gl, line 81: Can we not use function call here? Something like
```
[expanded]="customPanel.panelTitle === displayAspect"
```
Although since this is an expression here instead of a fixed variable, I'm not sure if it will actually have any impact or not. @AleksanderSklorz what would you suggest?

AleksanderSklorz: Maybe we can do something like add additional field in CustomPanel and that field would be set when 'displayAspect' input is changed because canExpandTheCard depends on displayAspect. 
So then we can refer to CustomPanel's field instead of calling function.

lib/content-services/src/lib/content-metadata/components/content-metadata/content-metadata.component.html
```diff
@@ -90,7 +90,7 @@
                      class="adf-metadata-grouped-properties-container">
                     <mat-expansion-panel *ngIf="showGroup(group) || editable"
                                          [attr.data-automation-id]="'adf-metadata-group-' + group.title"
-                                         [expanded]="canExpandTheCard(group) || !displayDefaultProperties && first">
+                                         [expanded]="canExpandTheCard(group.title) || !displayDefaultProperties && first">
```
AleksanderSklorz, line 93: Here similar case like above mentioned by @swapnil-verma-gl. Can we correct it too as we touch this place?

lib/content-services/src/lib/content-metadata/components/content-metadata/content-metadata.component.html
MichalKinas: The only problem that I can think of with this approach is the fact that by adding `expanded` property to the custom panel we kind of allow other programmers to change its state outside of the metadata component which should only be possible by changing `displayAspect` value @AleksanderSklorz what do you think?

AleksanderSklorz: I see your point @MichalKinas. Only idea which I see is having object like: 
someObject: { [key: string]: boolean } = {}; 

in your component and then fill this object with pairs when at the left side will be something what identifies panel (like panelTitle) and at right side boolean if it is expanded. 

So then you can fill that object in your component and it can't be changed from parent component + you can then read this without calling functions in HTML like: 

[expanded]="someObject[customPanel.panelTitle]"

AleksanderSklorz: So in that way you will get this value like by referring to just field instead of by function.

MichalKinas: I was just wondering which solution would be most effective in this case, to achieve the solution you proposed we would have to add custom setter to `customPanels`, `grouppedProperties` and `displayAspect` to make sure we always get correct panel to expand which will add additional operations and complexity to the code. And in this case the function we're calling executes only one string comparison so it is not really expensive that is why I wonder what do you guys think? @AleksanderSklorz @swapnil-verma-gl

AleksanderSklorz: Yeah good explanation. This function is not very complex so we can leave it but let's see what @swapnil-verma-gl says

swapnil-verma-gl: I'm fine with it. If I remember correctly, it was decided to leave the instances where moving from a function call to a fixed variable proved to be too complex. As you've already explained, updating the code would add too much complexity, so we can leave it

AleksanderSklorz: I'm closing my comments then

lib/content-services/src/lib/content-metadata/components/content-metadata/content-metadata.component.html
AleksanderSklorz: closing this as discussed in previous comment

lib/content-services/src/lib/content-metadata/components/content-metadata/content-metadata.component.html
MichalKinas: Thanks guys!

AleksanderSklorz (approved):

swapnil-verma-gl (approved):

lib/content-services/src/lib/content-metadata/components/content-metadata/content-metadata.component.html
swapnil-verma-gl: Marking my original comment as resolved. Thanks!!!

PR #9033, opened 2023-10-26
[AAE-16995] Code refactor of LocationCellComponent

wiktord2000:

> - [X] Bugfix
> - [ ] Feature
> - [X] Code style update (formatting, local variables)
> - [ ] Build related changes
> - [X] Documentation

> - [ ] Yes
> - [X] No

**Other information**:
We were trying to use this type to handle simple links but at some point we realized that it isn't the best place and we are going to compose another type in the future. But because I spent some time here I took this opportunity and made a small code refactor.

### DIFF
--- a/lib/core/src/lib/datatable/components/datatable-cell/datatable-cell.component.ts
+++ b/lib/core/src/lib/datatable/components/datatable-cell/datatable-cell.component.ts
@@ -78,22 +78,7 @@ export class DataTableCellComponent implements OnInit, OnDestroy {
 
     ngOnInit() {
         this.updateValue();
-        if (this.dataTableService) {
-            this.dataTableService.rowUpdate.pipe(takeUntil(this.onDestroy$)).subscribe((data) => {
-                if (data?.id) {
-                    if (this.row.id === data.id) {
-                        if (this.row.obj && data.obj) {
-                            this.row.obj = data.obj;
-                            this.row['cache'][this.column.key] = this.column.key
-                                .split('.')
-                                .reduce((source, key) => (source ? source[key] : ''), data.obj);
-
-                            this.updateValue();
-                        }
-                    }
-                }
-            });
-        }
+        this.subscribeToRowUpdates();
     }
 
     protected updateValue() {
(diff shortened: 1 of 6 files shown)


lib/core/src/lib/datatable/components/datatable-cell/datatable-cell.component.ts
```diff
@@ -108,6 +93,24 @@ export class DataTableCellComponent implements OnInit, OnDestroy {
...
+                    this.row['cache'][this.column.key] = this.column.key
+                        .split('.')
+                        .reduce((source, key) => (source ? source[key] : ''), data.obj);
+
+                    this.updateValue();
+                }
+            }
+        });
+    }
+
```
tomgny, line 113: i think we can split it more to make it more understandable, something like:

```suggestion
private subscribeToRowUpdates(): void {
    if (!this.dataTableService || !this.row.obj) {
        return;
    }

    this.dataTableService.rowUpdate.pipe(takeUntil(this.onDestroy$)).subscribe((data) => {
        if (data?.id === this.row?.id && data.obj) {
            this.row.obj = data.obj;
            this.row['cache'][this.column.key] = this.getNestedPropertyValue(data.obj, this.column.key);
            
            this.updateValue();
            }
        }
    });
}

private getNestedPropertyValue(obj: any, path: string) {
    return path.split('.').reduce((source, key) => (source ? source[key] : ''), obj);
}
```

lib/core/src/lib/datatable/components/datatable-cell/datatable-cell.component.ts
```diff
@@ -108,6 +93,24 @@ export class DataTableCellComponent implements OnInit, OnDestroy {
         }
     }
 
+    protected subscribeToRowUpdates() {
```
tomgny, line 96: why it's protected? shouldn't be private?

lib/core/src/lib/datatable/components/location-cell/location-cell.component.ts
```diff
@@ -18,19 +18,24 @@
...
     template: `
         <ng-container>
-            <a href="" [title]="tooltip" [routerLink]="link">
+            <a [title]="tooltip" [routerLink]="link">
                 {{ value$ | async }}
             </a>
         </ng-container>
     `,
     encapsulation: ViewEncapsulation.None,
-    host: { class: 'adf-location-cell adf-datatable-content-cell' }
```
tomgny, line 33: don't we need `adf-location-cell` anymore?

tomgny (commented):
please check comments

lib/core/src/lib/datatable/components/datatable-cell/datatable-cell.component.ts
wiktord2000: Yes, I should be, thx

lib/core/src/lib/datatable/components/location-cell/location-cell.component.ts
wiktord2000: i've just noticed that we use it inside datatable.component.scss so I will revert it, thx

lib/core/src/lib/datatable/components/datatable-cell/datatable-cell.component.ts
wiktord2000: Good point, I'll change it, thx

tomgny (approved):

PR #9089, opened 2023-11-14
[AAE-18057] Fix token is not refreshed after silent-refresh is called…

alep85:
…. Get another token from the auth server.

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://alfresco.atlassian.net/browse/AAE-18057

`silent-refresh.html`  send the received tokens to the main application to get another token from the auth server

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/core/src/lib/assets/silent-refresh.html
+++ b/lib/core/src/lib/assets/silent-refresh.html
@@ -2,5 +2,8 @@
     <title>Alfresco silent static page</title>
 </head>
 <body>
+    <script>
+        (window.opener || window.parent).postMessage(location.hash || ('#' + location.search), location.origin);
+    </script>
     Silent refresh static page
 </body>


popovicsandras (approved):

lib/core/src/lib/assets/silent-refresh.html
```diff
@@ -2,5 +2,8 @@
     <title>Alfresco silent static page</title>
 </head>
 <body>
+    <script>
```
popovicsandras, line 5: Seems quite workaround-y, but for the sake of release every solution is good solution! 😅 
Please let's create a ticket for post-release to investigate how this could be cleaned up.

popovicsandras (approved):

lib/core/src/lib/assets/silent-refresh.html
alep85: I've got that solution from the [official documentation ](https://manfredsteyer.github.io/angular-oauth2-oidc/docs/additional-documentation/silent-refresh.html)

tomgny (approved):

lib/core/src/lib/assets/silent-refresh.html
mauriziovitale: we were doing kind of the same the only difference was that js api was injecting the post into a div dynamically on silentRefresh page. The only concerne is... have we tested it? have u checked it the refreshToken is present or the idToken got refreshed ?

alep85: I tested yesterady locally with studio app, with this change it refresh the token before expire and all the other API call work good, of course I should try it merging the upstream

PR #9101, opened 2023-11-21
[ACS-6211] Fixed UI alignment issues in create rules screen (CardView and child components)

swapnil-verma-gl:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation
Minor UI and styling adjustments

UI for the cardView Component when viewed inside CreateRules section in ACA had alignment issues

Adjusted the UI/styling slightly, to fix alignment issues

> - [ ] Yes
> - [x] No

**Other information**:
This PR links to [this](https://github.com/Alfresco/alfresco-content-app/pull/3505#issuecomment-1794497287) comment on [this](https://github.com/Alfresco/alfresco-content-app/pull/3531) PR

### DIFF
--- a/lib/core/src/lib/card-view/components/card-view-boolitem/card-view-boolitem.component.ts
+++ b/lib/core/src/lib/card-view/components/card-view-boolitem/card-view-boolitem.component.ts
@@ -22,7 +22,14 @@ import { BaseCardView } from '../base-card-view';
 
 @Component({
     selector: 'adf-card-view-boolitem',
-    templateUrl: './card-view-boolitem.component.html'
+    templateUrl: './card-view-boolitem.component.html',
+    styles: [
+        `
+            .adf-property-value {
+                padding: 15px 0;
+            }
+        `
+    ]
 })
 
 export class CardViewBoolItemComponent extends BaseCardView<CardViewBoolItemModel> {
--- a/lib/core/src/lib/card-view/components/card-view-selectitem/card-view-selectitem.component.spec.ts
+++ b/lib/core/src/lib/card-view/components/card-view-selectitem/card-view-selectitem.component.spec.ts
@@ -26,6 +26,7 @@ import { AppConfigService } from '../../../app-config/app-config.service';
 import { HarnessLoader } from '@angular/cdk/testing';
 import { TestbedHarnessEnvironment } from '@angular/cdk/testing/testbed';
 import { MatSelectHarness } from '@angular/material/select/testing';
+import { MatLabel } from '@angular/material/form-field';
 
 describe('CardViewSelectItemComponent', () => {
     let loader: HarnessLoader;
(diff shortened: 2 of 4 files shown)


lib/core/src/lib/card-view/components/card-view-boolitem/card-view-boolitem.component.ts
```diff
@@ -22,7 +22,14 @@ import { BaseCardView } from '../base-card-view';
 
 @Component({
     selector: 'adf-card-view-boolitem',
-    templateUrl: './card-view-boolitem.component.html'
+    templateUrl: './card-view-boolitem.component.html',
+    styles: [
```
DominikIwanek, line 26: Isn't it better and more clear to move it to .scss file?

lib/core/src/lib/card-view/components/card-view-selectitem/card-view-selectitem.component.spec.ts
```diff
@@ -73,14 +73,6 @@ describe('CardViewSelectItemComponent', () => {
     });
 
     describe('Rendering', () => {
-        it('should render the label', () => {
```
AleksanderSklorz, line 76: should this test be removed? Label is still rendered, just in different place so we can still check if that is rendered: 
image

lib/core/src/lib/card-view/components/card-view-selectitem/card-view-selectitem.component.html
```diff
@@ -11,7 +7,8 @@
             data-automation-class="read-only-value">{{ (property.displayValue | async) | translate }}
         </div>
         <div *ngIf="isEditable()">
-            <mat-form-field class="adf-select-item-padding-editable adf-property-value">
+            <mat-form-field class="adf-property-value" appearance="standard">
+                <mat-label>{{ property.label }}</mat-label>
```
MichalKinas, line 11: Isn't it the breaking change for some use cases? Right now the label will only be displayed when the field is editable

lib/core/src/lib/card-view/components/card-view-boolitem/card-view-boolitem.component.ts
swapnil-verma-gl: The only reason I added it here is because it's only a single class, and the scss file did not exist. However, if needed, I can create a new file and add the styling to that instead.

lib/core/src/lib/card-view/components/card-view-selectitem/card-view-selectitem.component.spec.ts
swapnil-verma-gl: I only removed this test since after the change, we basically went from our custom implementation of showing a label for a form field, to angular material's proposed way of showing labels for form fields. Since it was something coming from angular material, I figured there is no need to have a unit test for it anymore, since it would already be covered by them. 
However, the label code might change after @MichalKinas 's review comment. If that is indeed the case then I will restore this unit test

AleksanderSklorz: @swapnil-verma-gl Fine for me, you can skip this comment. Only thing which I thought what is not tested by "Angular material" is that they can test only if label is displayed but they are not testing that this is coming from "property.label" so if anyone will change this part: 

"{{ property.label }}" 

then it won't be protected. So somebody can use different field here or something. But this is small thing so if you think it is not needed then I'm fine to skip it. 


I will approve that PR when you resolve this comment from ACA https://github.com/Alfresco/alfresco-content-app/pull/3505#issuecomment-1820596560 as there is still some issue in ACA even with these ADF changes @swapnil-verma-gl

swapnil-verma-gl: Thank you @AleksanderSklorz. The alignment issue has been resolved with [this](https://github.com/Alfresco/alfresco-ng2-components/pull/9101/commits/9b8001578fb3f4876fb38c14aaf5dfe941b51184) commit

lib/core/src/lib/card-view/components/card-view-selectitem/card-view-selectitem.component.html
swapnil-verma-gl: Re-added the label only for non-editable mode. For editable mode, we are still using the mat-label introduced in this PR. The reason I went for this approach is to maintain consistency of this component with the rest of the applications. I've also added corresponding unit tests for both the labels

lib/core/src/lib/card-view/components/card-view-selectitem/card-view-selectitem.component.spec.ts
swapnil-verma-gl: I've re-added this unit test, and added a another unit test corresponding to the presence of mat-label as well, as part of resolving @MichalKinas 's review comment

MichalKinas (approved):

PR #9186, opened 2023-12-18
[AAE-19070] Disable broken column types ("location" and "image") in data table widget

wiktord2000:

> - [X] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://alfresco.atlassian.net/browse/AAE-19070

> - [ ] Yes
> - [X] No

**Other information**:

### DIFF
--- a/lib/process-services-cloud/src/lib/form/components/widgets/data-table/data-table-adapter.widget.ts
+++ b/lib/process-services-cloud/src/lib/form/components/widgets/data-table/data-table-adapter.widget.ts
@@ -23,6 +23,7 @@ import {
 
 export class WidgetDataTableAdapter extends ObjectDataTableAdapter {
 
+    static readonly hiddenColumnTypes = ['image', 'location'];
     private rows: DataRow[];
     private columns: DataColumn[];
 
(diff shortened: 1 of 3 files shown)


lib/process-services-cloud/src/lib/form/components/widgets/data-table/data-table-adapter.widget.ts
```diff
@@ -44,6 +46,12 @@ export class WidgetDataTableAdapter extends ObjectDataTableAdapter {
         return this.hasAllColumnsLinkedToData() && this.hasAllMandatoryColumnPropertiesHaveValues();
     }
 
+    private hideColumnTypes(typesToHide: string[]): void {
+        this.columns
+            .filter(column => typesToHide.includes(column.type))
+            .forEach(column => column.isHidden = true);
```
tomgny, line 52: i think instead of setting isHidden property for some columns, we can just filter them out at all (i guess with this solution once we complete the form, our payload will still contain hidden columns which makes no sense but im not sure, have you tested it?)
something like 
```suggestion
            super.getColumns().filter....
```

wiktord2000: PR will be closed

wiktord2000:
In this case we need only update backend validation for different types of column

PR #9227, opened 2024-01-11
[AAE-17458] Move default apps away from repository

MichalFidor:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

Currently we are relying on the default applications which are in this repository. The problem is that it's public and we have them zipped which makes harder to track what change has been made in the particular PR.

This PR contains:
- Removing the apps from this repository,
- Added the logic of downloading the repo as zip and then unzipping it and zip again the particular app - seems that it could be simplified, but - the other repo is private, so we need to use a token. Because of that, I need to use a GH API which allows only to download whole repo as zip. There is no possibility to download one particular thing,
- Introduced a new logic to the script `init-aee-env`. Let me know, if I need to update also the two other - **I don't think I need**
- I've removed unneeded information from `resource.ts` file as it's not needed anymore - most probably this file should be also moved to the private repo 🤔 - **it's needed for hxp-frontend apps, bringing it back**

I think that...
> - [x] Yes
> - [ ] No

**Other information**:
- We need to generate a new token, that will have a permissions to private repositories inside of Hyland org
- Other things that I've missed and hopefully you will catch during the PR review


package.json
```diff
@@ -205,8 +206,7 @@
     "tsconfig-paths": "^4.1.1",
     "typescript": "4.7.4",
     "webdriver-manager": "12.1.9",
-    "webpack": "^5.88.2",
-    "webpack-cli": "^5.1.2"
```
acichon-h, line 209: Is this change intentional?

lib/cli/scripts/init-aae-env.ts
```diff
@@ -94,7 +98,7 @@ async function healthCheck(nameService: string) {
  */
 async function getApplications(): Promise<{ list: { entries: any[] } }> {
     const url = `${args.host}/deployment-service/v1/applications`;
-
+    logger.info(url)
```
acichon-h, line 101: I believe it would be good to log some description besides url for context

lib/cli/scripts/github-folder-downloader.ts
```diff
@@ -0,0 +1,51 @@
...
+    private token: string;
+    private repo: string;
+    private subFolderName: string;
+
+    constructor(token: string, repo: string) {
+        this.token = token;
+        this.repo = repo;
+    }
+
+    async downloadRepoAndUnpackRepository(owner: string, branch = 'main'): Promise<void> {
```
acichon-h, line 16: This doesn't sound right 😅 
```suggestion
    async downloadAndUnpackRepository(owner: string, branch = 'main'): Promise<void> {
```

lib/cli/scripts/github-folder-downloader.ts
```diff
@@ -0,0 +1,51 @@
...
+            Accept: 'application/vnd.github+json',
+            'X-GitHub-Api-Version': '2022-11-28'
+        };
+
+        try {
+            const response: AxiosResponse = await axios.get(url, {
+                responseType: 'arraybuffer',
+                headers,
+            });
+            const repoZipped = new AdmZip(response.data);
```
acichon-h, line 29:
```suggestion
            const zippedRepo = new AdmZip(response.data);
```

lib/cli/scripts/github-folder-downloader.ts
```diff
@@ -0,0 +1,51 @@
...
+            this.subFolderName = (repoZipped.getEntries().find(entry => entry.entryName.includes(owner))).entryName;
```
acichon-h, line 30: `Subfolder` is a one word
```suggestion
            this.subfolderName = (zippedRepo.getEntries().find(entry => entry.entryName.includes(owner))).entryName;
```

lib/cli/scripts/github-folder-downloader.ts
```diff
@@ -0,0 +1,51 @@
...
+
+            new AdmZip(response.data).extractAllTo(this.repo, /** overwrite **/ true);
```
acichon-h, line 32: I'm not sure if this comment should stay here
```suggestion
            new AdmZip(response.data).extractAllTo(this.repo, true);
```

lib/cli/scripts/github-folder-downloader.ts
```diff
@@ -0,0 +1,51 @@
...
+    async zipNeededApplication(appName: string): Promise<string> {
+        const pathToZip = path.resolve(`${process.cwd()}/${this.repo}`, `${appName}.zip`);
+        try {
+            const zip = new AdmZip();
+            zip.addLocalFolder(`${process.cwd()}/${this.repo}/${this.subFolderName}${appName}`);
+            await zip.writeZipPromise(pathToZip);
+        } catch (e) {
+            console.log(`Error during zipping app ${e}`)
+        }
+        return pathToZip;
```
acichon-h, line 48: Does it make any sense to return path to zip if zipping failed?

MichalFidor: Basically it should throw an error, I've missed that!

package.json
MichalFidor: Nope, I've removed it by mistake

lib/cli/scripts/init-aae-env.ts
```diff
@@ -117,12 +119,10 @@ async function getApplications(): Promise<{ list: { entries: any[] } }> {
             contentTypes,
             accepts
         );
-        result.on('error', (error) => {
-            logger.error(`Get application by status ${error} `);
-        });
+        // logger.info(result)
```
DenysVuika, line 122: commented out code

lib/cli/scripts/github-folder-downloader.ts
```diff
@@ -0,0 +1,50 @@
...
+                owner,
+                ref: branch,
+             })
+
+            const parsedResponse = Buffer.from(<ArrayBuffer>response.data);
+            const zippedRepo = new AdmZip(parsedResponse);
+            this.subfolderName = (zippedRepo.getEntries().find(entry => entry.entryName.includes(owner))).entryName;
+
+            zippedRepo.extractAllTo(this.repo, true);
+
```
acichon-h, line 32: Please remove redundant empty lines

lib/cli/scripts/github-folder-downloader.ts
```diff
@@ -0,0 +1,50 @@
...
+        } catch (error) {
+            throw new Error(`Failed to download repository: ${error}`);
+        }
+    }
+
+    async zipNeededApplication(appName: string): Promise<string> {
```
acichon-h, line 38:
```suggestion
    async zipApplication(appName: string): Promise<string> {
```

lib/cli/scripts/resources.ts
```diff
@@ -19,196 +19,192 @@
...
-        },
-        security: [
-            { role: 'APPLICATION_MANAGER', groups: [], users: ['manageruser'] },
-            { role: 'ACTIVITI_ADMIN', groups: [], users: ['superadminuser', 'processadminuser'] },
-            { role: 'ACTIVITI_USER', groups: ['hr', 'sales', 'testgroup'], users: ['hruser'] }
-        ],
-        infrastructure: { connectors: { restconnector: {} }, bridges: {} },
-        enableLocalDevelopment: true
-    },
+    // SUB_PROCESS_APP: {
```
acichon-h, line 22: I believe either we leave this code or remove it completely.
If you are still testing, then mark the PR as draft, please

MichalFidor:
It will be done in another PR, not in ADF repo

PR #9256, opened 2024-01-19
[AAE-19610] Submit query on change event for search date range tabbed widget

DiogoABastos:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/content-services/src/lib/search/components/search-date-range-tabbed/search-date-range-tabbed.component.html
+++ b/lib/content-services/src/lib/search/components/search-date-range-tabbed/search-date-range-tabbed.component.html
@@ -5,12 +5,9 @@
             [dateFormat]="settings.dateFormat"
             [maxDate]="settings.maxDate"
             [field]="field"
-            [hideDefaultAction]="settings?.hideDefaultAction"
             [initialValue]="startValue"
             (changed)="onDateRangedValueChanged($event, field)"
-            (valid)="tabsValidity[field]=$event"
-            (reset)="reset()"
-            (submit)="submitValues()">
+            (valid)="tabsValidity[field]=$event">
         </adf-search-date-range>
     </ng-container>
 </adf-search-filter-tabbed>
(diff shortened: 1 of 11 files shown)


BSekula (approved):

eromano (approved):

swapnil-verma-gl:
Having a submitValues() call inside onDateChanged() method is causing the search query to be unnecessarily triggered while the user is still setting values and not finished is input yet in ACA/ADW. See video below - 

https://github.com/Alfresco/alfresco-ng2-components/assets/92505353/05eea189-062d-42c9-8f2a-10aabe9912f6

DiogoABastos:
I've added a `debounceTime` to the form's valueChanges event which seems to have fixed the multiple search call issue. However, based on your video, we should only call the submitValues method if `hideDefaultActions` is true, so we avoid calling the search query when the apply button is available, right? @swapnil-verma-gl 
`if (settings.hideDefaultActions) this.submitValues();`

swapnil-verma-gl:
> I've added a `debounceTime` to the form's valueChanges event which seems to have fixed the multiple search call issue. However, based on your video, we should only call the submitValues method if `hideDefaultActions` is true, so we avoid calling the search query when the apply button is available, right? @swapnil-verma-gl `if (settings.hideDefaultActions) this.submitValues();`

I don't think the Clear and Apply buttons being shown here come from the hideDefaultActions parameter being false. They are separate buttons, being projected through [search-widget-chip.component](https://github.com/Alfresco/alfresco-ng2-components/blob/develop/lib/content-services/src/lib/search/components/search-filter-chips/search-widget-chip/search-widget-chip.component.html#L41). So adding the if condition as you proposed wouldn't really have any effect.

Would it serve your purpose if you were to go back to your original approach of creating the default actions buttons inside the search filter. That way, we can simply toggle the default actions to OFF via the hideDefaultActions parameter, thereby ensuring that none of the changes impact ACA/ADW

DiogoABastos:
> Would it serve your purpose if you were to go back to your original approach of creating the default actions buttons inside the search filter. That way, we can simply toggle the default actions to OFF via the hideDefaultActions parameter, thereby ensuring that none of the changes impact ACA/ADW

We want to have this component submit the value on change rather than using the buttons. With this solution, will it impact ACA/ADW?

eromano (approved):

PR #9272, opened 2024-01-24
[ACS-5704][Community request] ObjectDataTableAdapter sorting

tamaragruszka:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

Priority column on task list does not sort in order of number
[ACS-5704](https://alfresco.atlassian.net/browse/ACS-5704?atlOrigin=eyJpIjoiZDc0MzA0MDY2MThlNDIwYThkYTE5MjQyYzFmODMwZjEiLCJwIjoiaiJ9)

Alphanumeric sorting is enabled by default. There is possibility to pass options object for sorting.

> - [ ] Yes
> - [x] No

**Other information**:

[ACS-5704]: https://alfresco.atlassian.net/browse/ACS-5704?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/core/src/lib/datatable/components/datatable/datatable.component.spec.ts
```diff
@@ -73,7 +73,7 @@ export const resolverFn = (row: DataRow, col: DataColumn) => {
     return value;
 };
 
-describe('DataTable', () => {
+fdescribe('DataTable', () => {
```
DenysVuika, line 76: should not be comitted

lib/core/src/lib/datatable/data/object-datatable-adapter.ts
```diff
@@ -121,35 +121,33 @@ export class ObjectDataTableAdapter implements DataTableAdapter {
...
+                    left = left.valueOf().toString();
+                }
 
-                    return sorting.direction === 'asc' ? left.localeCompare(right) : right.localeCompare(left);
+                if (typeof right !== 'string') {
+                    right = right.valueOf().toString();
                 }
+
+                return sorting.direction === 'asc'
+                    ? left.localeCompare(right, 'en', sorting.options)
```
MichalKinas, line 136: In this case instead of using host's local you'll always use `en` local, is it fine for us?

lib/core/src/lib/datatable/components/datatable/datatable.component.ts
```diff
@@ -639,8 +639,8 @@ export class DataTableComponent implements OnInit, AfterContentInit, OnChanges,
             if (current && column.key === current.key) {
                 newDirection = current.direction?.toLowerCase() === 'asc' ? 'desc' : 'asc';
             }
-            this.sorting = [column.key, newDirection];
-            this.data.setSorting(new DataSorting(column.key, newDirection));
+            this.sorting = [column.key, newDirection, { numeric: true }];
+            this.data.setSorting(new DataSorting(column.key, newDirection, { numeric: true }));
```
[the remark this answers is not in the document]
tamaragruszka: the problem is that there is no way to check it, because column type for integers is as "text"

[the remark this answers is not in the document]
AleksanderSklorz: Additionally if I remember correctly, there was a point during our call with Agnieszka that maybe it could be good idea to have sorted mixed values like: 

a1
a10
file1
file2
file3

So you can see in above example that it sorts first by strings ('a' before 'file') and then it additionally sorts it by number (like 1, 10 or 1, 2, 3). 

In that case "numeric" field allows for this kind of sorting.

lib/core/src/lib/datatable/data/object-datatable-adapter.ts
tamaragruszka: in most cases it is safe, but that could break a little with special letters

MichalKinas: I think we should discuss this with Agnieszka and Denys to make sure we're not introducing breaking changes to the datatable

lib/core/src/lib/datatable/components/datatable/datatable.component.ts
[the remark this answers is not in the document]
MichalKinas: Okay in that case it's fine for me, maybe it would be also worth to ask someone from HX to review to ensure we're not breaking their side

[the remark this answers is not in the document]
DenysVuika: the numbers should be used only when the column type is set to 'number'

lib/core/src/lib/datatable/data/object-datatable-adapter.ts
```diff
@@ -121,35 +121,33 @@ export class ObjectDataTableAdapter implements DataTableAdapter {
...
+                let left = a.getValue(sorting.key) ?? '';
+                let right = b.getValue(sorting.key) ?? '';
 
-                    if (right) {
-                        right = right instanceof Date ? right.valueOf().toString() : right.toString();
-                    } else {
-                        right = '';
-                    }
+                if (typeof left !== 'string') {
+                    left = left.valueOf().toString();
```
DenysVuika, line 128: it's a very risky assumption that everything that is not the string is a date... it can be a number or an object for example

tamaragruszka: it is safe, method `valueOf` exists for every not empty data type

MichalKinas (approved):

DenysVuika (approved):

PR #9305, opened 2024-02-06
[AAE-19968] Field visibility condition using Integer and greater than not working

wiktord2000:

> - [X] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://alfresco.atlassian.net/browse/AAE-19968

> - [ ] Yes
> - [X] No

**Other information**:

### DIFF
--- a/lib/core/src/lib/form/models/widget-visibility.model.ts
+++ b/lib/core/src/lib/form/models/widget-visibility.model.ts
@@ -54,7 +54,7 @@ export class WidgetVisibilityModel {
     }
 
     get leftValue(): any {
-        if (this.json.leftValue != null) {
+        if (this.json.leftValue?.toString()) {
             return this.json.leftValue;
         } else if (this.leftFormFieldId) {
             return this.leftFormFieldId;
@@ -87,7 +87,7 @@ export class WidgetVisibilityModel {
     }
 
     get rightValue(): any {
-        if (this.json.rightValue != null) {
+        if (this.json.rightValue?.toString()) {
             return this.json.rightValue;
         } else if (this.rightFormFieldId) {
             return this.rightFormFieldId;
--- a/lib/core/src/lib/form/models/widget-visibility.model.ts
+++ b/lib/core/src/lib/form/models/widget-visibility.model.ts
@@ -54,7 +54,7 @@ export class WidgetVisibilityModel {
     }
 
     get leftValue(): any {
-        if (this.json.leftValue) {
+        if (this.json.leftValue || this.json.leftValue === false || this.json.leftValue === 0) {
             return this.json.leftValue;
         } else if (this.leftFormFieldId) {
             return this.leftFormFieldId;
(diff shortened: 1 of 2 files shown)


lib/core/src/lib/form/services/widget-visibility.service.spec.ts
```diff
@@ -421,6 +421,30 @@ describe('WidgetVisibilityService', () => {
             expect(rightValue).toBe('dropdown_label');
         });
 
+        it('should getRightValue be able to return 0 when rightType is set to value', () => {
```
eromano, line 424: I guess we need a unit test that check how the function behave in case rightValue is null after this change

eromano (changes requested):

lib/core/src/lib/form/models/widget-visibility.model.ts
```diff
@@ -86,7 +86,7 @@ export class WidgetVisibilityModel {
     }
 
     get rightValue(): any {
-        if (this.json.rightValue) {
+        if (this.json.rightValue != null) {
```
tomgny, line 89: shouldn't we use here strict (`!==`) checking? with this implementation, if rightValue is undefind, the condition will be met, is this intended?

lib/core/src/lib/form/services/widget-visibility.service.spec.ts
mauriziovitale, line 424: rightType is set to value 
it should return 0 when rightValue is a number
it should return false when rightValue is a boolean
it should return empty string when rightValue is a string
it should return X when rightValue is undefined or null
it should return Y when rightValue is an object. ??

mauriziovitale: I think we also need to review the same code on left 

  ```
  get leftValue(): any {
        if (this.json.leftValue) {
  ```      
      
I would better check the scenario because it looks inconsistent. We are setting the **rightType** to **value** and then the **value** is null ? Why is it null?

lib/core/src/lib/form/models/widget-visibility.model.ts
wiktord2000: It was intended but not sure whether the best solution

lib/core/src/lib/form/services/widget-visibility.service.spec.ts
```diff
@@ -421,6 +421,51 @@ describe('WidgetVisibilityService', () => {
             expect(rightValue).toBe('dropdown_label');
         });
 
+        describe('getRightValue when rightType is set to value', () => {
+            beforeEach(() => {
+                visibilityObjTest.rightType = 'value';
+            });
+
+            it('should be able to return 0 when rightValue is a number', () => {
```
wiktord2000, line 429: add. test for e.g. 10 > 0

lib/core/src/lib/form/services/widget-visibility.service.spec.ts
```diff
@@ -421,6 +421,51 @@ describe('WidgetVisibilityService', () => {
...
+                visibilityObjTest.rightValue = 0;
+                const rightValue: any = service.getRightValue(formTest, visibilityObjTest);
+
+                expect(rightValue).toBe(0);
+            });
+
+            it('should be able to return false when rightValue is a boolean', () => {
```
wiktord2000, line 436: Check true > "false"

mauriziovitale (approved):

PR #9310, opened 2024-02-07
[ACS-6643] Incorrect layout for Search widgets of the Text type

tamaragruszka:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

Constant width for textfield
[ACS-6643](https://alfresco.atlassian.net/browse/ACS-6643)

Textfield adjust to parent width

> - [ ] Yes
> - [x] No

**Other information**:

[ACS-6643]: https://alfresco.atlassian.net/browse/ACS-6643?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/content-services/src/lib/search/components/search-text/search-text.component.scss
+++ b/lib/content-services/src/lib/search/components/search-text/search-text.component.scss
@@ -1,6 +1,5 @@
 .adf-search-text {
     .mat-form-field {
         width: 100%;
-        max-width: 168px;
     }
 }


lib/content-services/src/lib/search/components/search-text/search-text.component.scss
```diff
@@ -1,6 +1,5 @@
 .adf-search-text {
     .mat-form-field {
         width: 100%;
-        max-width: 168px;
```
MichalKinas, line 4: I wonder a bit about the bug that introduced this fix before your PR, maybe we should use `hidden` instead of `ngIf` for the clear button to ensure that layout won't "jump" in case this component is used in similar scenario as in that bug

MichalKinas (changes requested):

lib/content-services/src/lib/search/components/search-text/search-text.component.scss
tamaragruszka: I checked in application, it's not jumping anymore

MichalKinas: But the bug was introduce in the HX app, so you would need to check it or apply suggested changes

MichalKinas (approved):

PR #9341, opened 2024-02-15
AAE-20480 Full Screen User Task Forms

pmartinezga:

> - [ ] Bugfix
> - [X] Feature
> - [ ] Build related changes
> - [ ] Documentation

Forms can not be displayed in full screen

AAE-20480

> - [ ] Yes
> - [ ] No

**Other information**:

Related to: 
- https://github.com/Alfresco/hxp-frontend-apps/pull/7469
- https://github.com/Alfresco/hxp-frontend-apps/pull/7471


lib/process-services-cloud/src/lib/form/components/form-cloud.component.scss
```diff
@@ -0,0 +1,98 @@
...
+        }
+    }
+
+    &-fullscreen-button {
+        position: absolute;
+        right: 70px;
+        top: 30px;
+    }
+
+    &__display-name {
```
popovicsandras, line 44: Are we sure, we are not causing any breaking changes with these?

lib/process-services-cloud/src/lib/form/services/form-cloud.service.ts
```diff
@@ -44,6 +44,8 @@ export class FormCloudService extends BaseCloudService implements FormCloudServi
         super(adfHttpClient);
     }
 
+    fullScreenMode = new Subject<boolean>();
```
popovicsandras, line 47: I'm not sure that how we display the form (whether inset or fullscreen) should be in the form-cloud.service.ts file, where most of the methods are related to the business logic and not the representation.

lib/process-services-cloud/src/lib/form/components/form-cloud.component.html
```diff
@@ -3,47 +3,82 @@
...
-                    <span *ngIf="isTitleEnabled()" class="adf-form-title" [matTooltip]="form.taskName">
-                        {{form.taskName}}
-                        <ng-container *ngIf="!form.taskName">
-                            {{'FORM.FORM_RENDERER.NAMELESS_TASK' | translate}}
-                        </ng-container>
-                    </span>
+<div *ngIf="hasForm()" class="adf-cloud-form-container"
+    [class.adf-cloud-form-overlay-container]="overlayMode"
+    [class.adf-cloud-form-inline-container]="!overlayMode">
+    <div class="adf-cloud-form-content"
```
popovicsandras, line 9: I understand that for the sake of POC, we are doing some css magic, but wouldn't it be better to use Angular Material overlaying?

lib/process-services-cloud/src/lib/services/form-fields.interfaces.ts
```diff
@@ -26,6 +26,7 @@ export interface FormRepresentation {
     version?: number;
     formDefinition?: FormDefinition;
     standAlone?: boolean;
+    forceFullScreen?: boolean;
```
popovicsandras, line 29: I think we could be more generic, and introduce the `displayMode` property here.
`default` would be our "legacy" display of the form, and now we could add `fullscreen`. Later on this way, we can enrich display modes.

lib/process-services-cloud/src/lib/form/services/form-cloud.service.ts
popovicsandras: Form Rendering engine? Or what other class we have for dealing with the form display?

popovicsandras (approved):
Approved, however please see my comments.

lib/process-services-cloud/src/lib/form/components/form-cloud.component.scss
pmartinezga: We are using a brand new selector prefix (`.adf-cloud-form`), so I don't think we can cause issues

lib/process-services-cloud/src/lib/form/services/form-cloud.service.ts
pmartinezga: Absolutely, I think it makes more sense to have this subject on the `FormRenderingService`. Thanks for the suggestion.

lib/process-services-cloud/src/lib/form/components/form-cloud.component.html
pmartinezga: I have done it the same way it is done in the ADF viewer, but I can explore this option

lib/process-services-cloud/src/lib/services/form-fields.interfaces.ts
pmartinezga: So you mean that we can introduce here instead of a `boolean` an enumerated value, and by now, we only have two options here `inline`(default) and `fullScreen`. I think that I will do something like this to be more generic, so the components receive an array of available display mode configurations, and the default display mode:

```typescript
export enum FormCloudDisplayMode {
    inline = 'inline',
    fullScreen = 'fullScreen'
}

export interface FormCloudDisplayModeConfiguration: {
	displayMode: FormCloudDisplayMode,
	options: any
};
`
``

I will create method in the `FormRenderingService` for getting the default display model configuration if not provided.

OK this makes sense, I will use a radio button group in the modeling for selecting the `displayMode` in the form properties

File: docs/process-services-cloud/components/form-cloud.component.md [hunk 5]
```diff
@@ -160,6 +163,49 @@ For an existing Task both the form and its values will be fetched and displayed.
...
 In this case, only the form definition will be fetched.
 
+#### Enabling fullscreen display for the form
+
+Provide a `displayModeConfiguration` array object containing the fullscreen configuration. You can use the configuration provided in the [`CloudFormRenderingService`](../../../lib/process-services-cloud/src/lib/form/components/cloud-form-rendering.service.ts) as a static member `CloudFormRenderingService.IMPLEMENTED_DISPLAY_MODE_CONFIGURATIONS`, or configure your own if you want to customise the options for the fullscreen display mode.
+
+**MyView.component.html**
+
+```html
+<button (click)="displayFormInFullscreenMode()">
```
DenysVuika, line 173: this example can be simplified, you don't need `@ViewChild` and extra methods, as you already provide reference to the form as `#adfCloudForm`, so the code can be simply:

```html
<button (click)="adfCloudForm.switchToDisplayMode('fullScreen')">Full screen</button>
```

docs/process-services-cloud/components/form-cloud.component.md
```diff
@@ -160,6 +163,49 @@ For an existing Task both the form and its values will be fetched and displayed.
...
+
+    @ViewChild('adfCloudForm', { static: true })
+    adfCloudForm: FormCloudComponent;
+
+    get displayConfigurations() {
+        return CloudFormRenderingService.IMPLEMENTED_DISPLAY_MODE_CONFIGURATIONS;
+    }
+
+    displayFormInFullscreenMode() {
+        this.adfCloudForm.switchToDisplayMode('fullScreen');
```
DenysVuika, line 201: this can be done via template

docs/process-services-cloud/components/form-cloud.component.md
```diff
@@ -160,6 +163,49 @@ For an existing Task both the form and its values will be fetched and displayed.
...
+```ts
+import { ViewChild } from '@angular/core';
+import { CloudFormRenderingService, FormCloudComponent } from '@alfresco/adf-process-services-cloud';
+
+export class MyView {
+
+    @ViewChild('adfCloudForm', { static: true })
+    adfCloudForm: FormCloudComponent;
+
+    get displayConfigurations() {
```
DenysVuika, line 196: for the example to showcase full screen, this can be omitted I believe

lib/process-services-cloud/src/lib/form/components/cloud-form-rendering.service.ts
```diff
@@ -1,51 +0,0 @@
...
-import { PropertiesViewerWidgetComponent } from './widgets/properties-viewer/properties-viewer.widget';
-import { RadioButtonsCloudWidgetComponent } from './widgets/radio-buttons/radio-buttons-cloud.widget';
-import { FileViewerWidgetComponent } from './widgets/file-viewer/file-viewer.widget';
-import { DisplayRichTextWidgetComponent } from './widgets/display-rich-text/display-rich-text.widget';
-import { DataTableWidgetComponent } from './widgets/data-table/data-table.widget';
-
-@Injectable({
-    providedIn: 'root'
-})
-export class CloudFormRenderingService extends FormRenderingService {
```
DenysVuika, line 34: isn't this a breaking change?

lib/process-services-cloud/src/lib/form/components/form-cloud.component.ts
```diff
@@ -376,10 +430,18 @@ export class FormCloudComponent extends FormBaseComponent implements OnChanges,
         return !args.defaultPrevented;
     }
 
-    protected storeFormAsMetadata() {}
+    protected storeFormAsMetadata() { }
```
DenysVuika, line 433: this might need to be skipped, empty method

lib/process-services-cloud/src/lib/task/task-form/components/task-form-cloud.component.ts
```diff
@@ -24,6 +24,8 @@ import { DropdownCloudWidgetComponent } from '../../../form/components/widgets/d
 import { DateCloudWidgetComponent } from '../../../form/components/widgets/date/date-cloud.widget';
 import { takeUntil } from 'rxjs/operators';
 import { Subject } from 'rxjs';
+import { FormCloudDisplayModeConfiguration } from '../../../services/form-fields.interfaces';
+import { FormCloudComponent } from '../../../form/public-api';
```
DenysVuika, line 28: incorrect import

lib/process-services-cloud/src/lib/task/task-form/components/task-form-cloud.component.ts
```diff
@@ -109,6 +117,17 @@ export class TaskFormCloudComponent implements OnInit, OnChanges, OnDestroy {
     @Output()
     onTaskLoaded = new EventEmitter<TaskDetailsCloudModel>(); /* eslint-disable-line */
 
+    /** Emitted when a display mode configuration is turned on. */
+    @Output()
```
DenysVuika, line 121: can those two events be just one "displayModeToggle" event with boolean?

DenysVuika (changes requested):

docs/process-services-cloud/components/form-cloud.component.md
pmartinezga: Thanks! I will apply this!

docs/process-services-cloud/components/form-cloud.component.md
pmartinezga: thanks!

docs/process-services-cloud/components/form-cloud.component.md
pmartinezga: I don't think so, because by default, the full screen display configuration is not provided, we provide just the inline one (the default we were using). 

If you want to be able to use the full screen mode, you need to include a configuration that provides it, for example, the one exported as static member in the `CloudFormRenderingService`. 

See the initial paragraph of this section in the MD file:

> Provide a `displayModeConfiguration` array object containing the fullscreen configuration. You can use the configuration provided in the [`CloudFormRenderingService`](../../../lib/process-services-cloud/src/lib/form/components/cloud-form-rendering.service.ts) as a static member `CloudFormRenderingService.IMPLEMENTED_DISPLAY_MODE_CONFIGURATIONS`, or configure your own if you want to customise the options for the fullscreen display mode.

lib/process-services-cloud/src/lib/form/components/cloud-form-rendering.service.ts
pmartinezga: I have just moved from one folder to another that is more meaningful, but inside the same module. If you think it can break anything, I can put it back to were it was

lib/process-services-cloud/src/lib/form/components/form-cloud.component.ts
pmartinezga: We need to provide an implementation of this `storeFormAsMetadata` method that is inherited from the `abstract` parent class. I just formatted this file, I did not change its implementation

lib/process-services-cloud/src/lib/task/task-form/components/task-form-cloud.component.ts
pmartinezga: Nice catch!

lib/process-services-cloud/src/lib/task/task-form/components/task-form-cloud.component.ts
```diff
@@ -231,6 +250,18 @@ export class TaskFormCloudComponent implements OnInit, OnChanges, OnDestroy {
         this.executeOutcome.emit(outcome);
     }
 
+    switchToDisplayMode(newDisplayMode?: string) {
+        this.adfCloudForm.switchToDisplayMode(newDisplayMode);
```
BSekula, line 254: Maybe we could call the service directly here? e.g. instead of calling child component method?

```
this.cloudFormRenderingService(we can have separate service)
  .switchToDisplayMode(this.id, FormCloudDisplayMode[newDisplayMode], this.displayMode, this.displayModeConfigurations);
```

The `id` can be passed to `adf-cloud-form`, in that case we could manipulate the form from any place in our app, we would just need to have the service and `id` (other data should also be available 🤔), what do you think?

lib/process-services-cloud/src/lib/form/components/form-cloud.component.ts
```diff
@@ -96,19 +106,32 @@ export class FormCloudComponent extends FormBaseComponent implements OnChanges,
...
+    @Output()
+    displayModeOff = new EventEmitter<FormCloudDisplayModeConfiguration>();
+
     protected subscriptions: Subscription[] = [];
     nodeId: string;
     formCloudRepresentationJSON: any;
 
     protected onDestroy$ = new Subject<boolean>();
 
+    readonly id: string;
```
BSekula, line 123: can we use `formId` instead?

lib/process-services-cloud/src/lib/services/form-fields.interfaces.ts
```diff
@@ -233,3 +234,28 @@ export enum FormFieldType {
...
+    [key: string]: any;
+};
+
+export interface FormCloudDisplayModeConfiguration {
+    displayMode: FormCloudDisplayMode;
+    options?: FormCloudDisplayModeConfigurationOptions;
+    default?: boolean;
+};
+
+// eslint-disable-next-line no-shadow
```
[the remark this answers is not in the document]
DenysVuika: @BSekula that's a famous false positive from the eslint, usually is fixed by moving enum to its own file

lib/process-services-cloud/src/lib/form/components/form-cloud.component.ts
```diff
@@ -166,6 +189,32 @@ export class FormCloudComponent extends FormBaseComponent implements OnChanges,
...
+        const formRepresentation = changes['form'];
+        if (formRepresentation?.currentValue) {
+            this.form = formRepresentation.currentValue;
+            this.onFormLoaded(this.form);
+            return;
+        }
+    }
+
+    ngOnInit(): void {
+        CloudFormRenderingService.displayMode$.pipe(filter(change => change.id === this.id), takeUntil(this.onDestroy$)).subscribe((displayModeChange) => {
```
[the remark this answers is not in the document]
pmartinezga: Done thanks for the suggestion

lib/process-services-cloud/src/lib/form/components/form-cloud.component.ts
pmartinezga: As I said in a comment before, formId is not mandatory (the renderer can receive the form definition directly, instead of a taskId or a formId) and you can break situations where you render the same form twice in the same page.

lib/process-services-cloud/src/lib/form/components/cloud-form-rendering.service.ts
pmartinezga: I adopted the proposed solution from @BSekula for creating a dedicated service for the display mode service, so this file remains now unchanged

lib/process-services-cloud/src/lib/form/services/cloud-form-rendering.service.ts
```diff
@@ -0,0 +1,170 @@
...
+import { FileViewerWidgetComponent } from '../components/widgets/file-viewer/file-viewer.widget';
+import { DisplayRichTextWidgetComponent } from '../components/widgets/display-rich-text/display-rich-text.widget';
+import { DataTableWidgetComponent } from '../components/widgets/data-table/data-table.widget';
+import { Observable, Subject } from 'rxjs';
+import { FormCloudDisplayMode, FormCloudDisplayModeChange, FormCloudDisplayModeConfiguration } from '../../services/form-fields.interfaces';
+
+@Injectable({
+    providedIn: 'root'
+})
+export class CloudFormRenderingService extends FormRenderingService {
```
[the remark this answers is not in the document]
pmartinezga: Done, thanks

PR #9365, opened 2024-02-21
[ACS-6586] "Folder" property complains about letters provided into "N…

jacekpluta:
…ame" input field

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

> - [ ] Yes
> - [x] No

**Other information**:
We are receiving requiresMatch property from backend which indicates if we should or should not match given regex.

Property explanation: 
/**
 * Set whether the regular expression must be matched or not
 * 
 * @param requiresMatch Set to <tt>true</tt> if the value must match the regular expression
 *      or <tt>false</tt> if the value must not match the regular expression
 */
public void setRequiresMatch(boolean requiresMatch)
{
    this.requiresMatch = requiresMatch;
}


lib/core/src/lib/card-view/validators/card-view-item-match.valiator.ts
```diff
@@ -19,16 +19,17 @@ import { CardViewItemValidator } from '../interfaces/card-view.interfaces';
...
     expression: string;
+    requiresMatch?: boolean;
     flags?: string;
 }
 
 export class CardViewItemMatchValidator implements CardViewItemValidator {
     message = 'CORE.CARDVIEW.VALIDATORS.MATCH_VALIDATION_ERROR';
 
-    constructor(private expression: string, private flags?: string) {}
+    constructor(private expression: string, private requiresMatch?: boolean, private flags?: string) {}
```
DenysVuika, line 29: please provide unit tests to cover the newly added checks

DenysVuika (changes requested):

lib/core/src/lib/card-view/validators/card-view-item-match.valiator.ts
AleksanderSklorz, line 29: Maybe it will be good idea to move that new parameter (requiredMatch) as last item of constructor? The reason is following: 
If any external user uses this class from ADF and initializes this class like: 
const validator = new CardViewItemMatchValidator(someExpression, someFlags); 

then this change will introduce breaking change for then because now at second parameter you have requiresMatch and not flags so they will get error that someFlags is string but boolean is expected. So we will enforce them to update their code. 

If you move requiresMatch at the end of constructor then everything will be fine for external customer because they won't need to update their code by changing order of parameters.

jacekpluta: very good explanation! fixed it and moved it as the last parameter

jacekpluta: tests added

lib/core/src/lib/card-view/models/card-view-baseitem.model.spec.ts
```diff
@@ -119,5 +119,34 @@ describe('CardViewBaseItemModel', () => {
...
+                    parameters: { expression: '(.*[\\"\\*\\\\\\>\\<\\?\\/\\:\\|]+.*)|(.*[\\.]?.*[\\.]+$)|(.*[ ]+$)', requiresMatch: false }
+                }]
+            };
+
+            const itemModel = new CarViewCustomItemModel(constrainedProperties);
+            expect(itemModel.isValid(itemModel.value)).toBe(false);
+        });
+
+        it('should validate when there is REGEX constraint and requiresMatch is true', () => {
+            const constrainedProperties = {
```
AleksanderSklorz, line 139: could you please set type to this const?

lib/core/src/lib/card-view/models/card-view-textitem.model.spec.ts
```diff
@@ -105,4 +105,24 @@ describe('CardViewTextItemModel', () => {
         itemModel.value = 'dummy';
         expect(itemModel.isValid(itemModel.value)).toBe(false, '`dummy` is not a constraint expression pattern');
     });
+
+    it('should validate based on defined constraints and not require a match to be valid', () => {
+        const constrainedProperties = {
```
AleksanderSklorz, line 110: could you please set type?

lib/core/src/lib/card-view/validators/card-view-item-match.valiator.ts
```diff
@@ -20,15 +20,16 @@ import { CardViewItemValidator } from '../interfaces/card-view.interfaces';
...
     expression: string;
     flags?: string;
+    requiresMatch?: boolean;
 }
 
 export class CardViewItemMatchValidator implements CardViewItemValidator {
     message = 'CORE.CARDVIEW.VALIDATORS.MATCH_VALIDATION_ERROR';
 
-    constructor(private expression: string, private flags?: string) {}
+    constructor(private expression: string, private flags?: string,  private requiresMatch?: boolean) {}
```
AleksanderSklorz, line 29: double spaces between private

jacekpluta: fixed

lib/core/src/lib/card-view/models/card-view-textitem.model.spec.ts
jacekpluta: added

lib/core/src/lib/card-view/models/card-view-baseitem.model.spec.ts
jacekpluta: added

lib/core/src/lib/card-view/models/card-view-baseitem.model.spec.ts
```diff
@@ -119,5 +119,34 @@ describe('CardViewBaseItemModel', () => {
             expect(isValid).toBe(false);
             expect(itemModel.getValidationErrors('test-against-this')).toEqual([validator1, validator3 ]);
         });
+
+        it('should validate when there is REGEX constraint and requiresMatch is false', () => {
+            const constrainedProperties = {
```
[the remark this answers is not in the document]
jacekpluta: yep,  added

lib/core/src/lib/card-view/models/card-view-baseitem.model.spec.ts
AleksanderSklorz: @jacekpluta I still can't see type here

AleksanderSklorz:
@jacekpluta sonarcloud found some issues, please fix them: 
image

lib/core/src/lib/card-view/models/card-view-textitem.model.spec.ts
```diff
@@ -16,8 +16,8 @@
  */
 
 import { PipeTransform } from '@angular/core';
-import { CardViewTextItemModel } from './card-view-textitem.model';
-import { CardViewTextItemProperties } from '../interfaces/card-view.interfaces';
+import {CardViewTextItemModel} from '@alfresco/adf-core';
```
DenysVuika, line 19: the import should be relative, as it's the same library

AleksanderSklorz (approved):

DenysVuika (approved):

PR #9390, opened 2024-02-28
[MNT-24082] fix aspect overwriting from dialog

g-jaskowski:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://alfresco.atlassian.net/browse/MNT-24028
Updating aspects with Aspect List Dialog Component overwrites all current node aspects.

Not listable aspects are stored and emitted on selection value change, as api requires.

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/content-services/src/lib/aspect-list/aspect-list.component.spec.ts
+++ b/lib/content-services/src/lib/aspect-list/aspect-list.component.spec.ts
@@ -152,7 +152,7 @@ describe('AspectListComponent', () => {
             spyOn(aspectListService, 'getCustomAspects').and.returnValue(of(customAspectListMock));
             spyOn(aspectListService, 'getVisibleAspects').and.returnValue(['frs:AspectOne']);
             nodeService = TestBed.inject(NodesApiService);
-            spyOn(nodeService, 'getNode').and.returnValue(of({ id: 'fake-node-id', aspectNames: ['frs:AspectOne'] } as any));
+            spyOn(nodeService, 'getNode').and.returnValue(of({ id: 'fake-node-id', aspectNames: ['frs:AspectOne', 'stored:aspect'] } as any));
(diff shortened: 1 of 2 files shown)


lib/content-services/src/lib/aspect-list/aspect-list.component.spec.ts
```diff
@@ -250,6 +250,21 @@ describe('AspectListComponent', () => {
                 component.clear();
                 expect(component.nodeAspects.length).toBe(0);
             });
+
+            it('should store node aspects that are not listed and emit then on value change', async () => {
+                const storedAspect: string[] = ['stored:aspect'];
```
AleksanderSklorz, line 255: no need to set type in this case because ts sees that you assign array with string so string[] type will be set automatically

lib/content-services/src/lib/aspect-list/aspect-list.component.spec.ts
```diff
@@ -250,6 +250,21 @@ describe('AspectListComponent', () => {
...
+
+                expect(component.notDisplayedAspects).toEqual(storedAspect);
+
+                spyOn(component.valueChanged, 'emit');
+                const panel = await loader.getAllHarnesses(MatExpansionPanelHarness);
+                await panel[0].expand();
+                const checkbox = await panel[0].getHarness(MatCheckboxHarness);
```
AleksanderSklorz, line 262: you have twice panel[0] so could you please move it to variable to don't repeat searching of array? 

You can even resolve both comments at once like: 
const panel = await loader.getAllHarnesses(MatExpansionPanelHarness)[0];

then no need even for plural as you will store only single panel there.

AleksanderSklorz (approved):

nikita-web-ua (approved):

PR #9391, opened 2024-02-29
[ACS-6233] Enable metadata e2es

nikita-web-ua:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation
> - [X] Other... Please describe:

https://alfresco.atlassian.net/browse/ACS-6233

> - [ ] Yes
> - [X] No

**Other information**:


lib/content-services/src/lib/content-metadata/components/content-metadata/content-metadata.component.html
```diff
@@ -204,9 +204,9 @@
         <div *ngFor="let group of groupedProperties; let first = first" class="adf-metadata-grouped-properties-container">
             <mat-expansion-panel
                 [attr.data-automation-id]="'adf-metadata-group-' + group.title"
-                [expanded]="currentPanel.panelTitle === group.title && currentPanel.expanded"
-                (opened)="expandPanel(group.title)"
-                (closed)="closePanel(group.title)"
+                [expanded]="group.title === displayAspect || !displayDefaultProperties && first || currentPanel.panelTitle === group.title && group.expanded"
+                (opened)="group.expanded = true"
```
MichalKinas, line 208: This doesn't seem correct `expandPanel` and `closePanel` are both properly setting the `currentPanel` which we should rely on. Otherwise integration with other panels will be broken

lib/content-services/src/lib/content-metadata/components/content-metadata/content-metadata.component.html
```diff
@@ -204,9 +204,9 @@
         <div *ngFor="let group of groupedProperties; let first = first" class="adf-metadata-grouped-properties-container">
             <mat-expansion-panel
                 [attr.data-automation-id]="'adf-metadata-group-' + group.title"
-                [expanded]="currentPanel.panelTitle === group.title && currentPanel.expanded"
-                (opened)="expandPanel(group.title)"
-                (closed)="closePanel(group.title)"
+                [expanded]="group.title === displayAspect || !displayDefaultProperties && first || currentPanel.panelTitle === group.title && group.expanded"
```
MichalKinas, line 207: `group.title === displayAspect` is already checked in `OnInit` and `OnChanges` and then compared with `currentPanel` so I think that adding `!displayDefaultProperties && first` part would be enough

MichalKinas (changes requested):

lib/content-services/src/lib/content-metadata/components/content-metadata/content-metadata.component.html
nikita-web-ua: This breaks the case when we use `multi` property to allow multi accordion to be expanded at the same time.

MichalKinas: Okay in this case we need to run different methods/logic based on `multi` state we also have to make sure not to break regular use case without multi as in this case only one panel can be expanded at once

lib/content-services/src/lib/content-metadata/components/content-metadata/content-metadata.component.html
nikita-web-ua: right, removed `group.title === displayAspect` as it's already checked

lib/content-services/src/lib/content-metadata/components/content-metadata/content-metadata.component.html
nikita-web-ua: addressed

lib/content-services/src/lib/content-metadata/components/content-metadata/content-metadata.component.ts
```diff
@@ -247,6 +247,11 @@ export class ContentMetadataComponent implements OnChanges, OnInit, OnDestroy {
             this.currentPanel.expanded = true;
         }
 
+        if (changes.displayDefaultProperties?.currentValue) {
+            this.currentPanel.panelTitle = this.DefaultPanels.PROPERTIES;
```
MichalKinas, line 251: I think you can use `this.expandPanel(this.DefaultPanels.PROPERTIES)` here to avoid duplicating the code

MichalKinas (approved):

PR #9407, opened 2024-03-05
[ACS-6581] Wrap comments into multiple lines

AleksanderSklorz:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://alfresco.atlassian.net/browse/ACS-6581

Comments for version manager are visible fully and wrapped. 

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/content-services/src/lib/version-manager/version-list.component.scss
+++ b/lib/content-services/src/lib/version-manager/version-list.component.scss
@@ -6,7 +6,7 @@
             height: auto;
             padding-top: 23.25px;
             padding-bottom: 23.25px;
-            border-bottom: 1px solid #d8d8d8;
+            border-bottom: 1px solid var(--adf-theme-foreground-divider-color);
 
             .adf-version-list-item-comment {
                 width: 100%;
--- a/lib/content-services/src/lib/version-manager/version-list.component.scss
+++ b/lib/content-services/src/lib/version-manager/version-list.component.scss
@@ -4,8 +4,8 @@
 
         .adf-version-list-item {
             height: auto;
-            padding-top: 23.25px;
-            padding-bottom: 23.25px;
+            padding-top: 23px;
+            padding-bottom: 23px;
             border-bottom: 1px solid var(--adf-theme-foreground-divider-color);
 
             .adf-version-list-item-comment {
--- a/lib/content-services/src/lib/version-manager/version-list.component.scss
+++ b/lib/content-services/src/lib/version-manager/version-list.component.scss
@@ -1,10 +1,19 @@
 .adf-version-list {
-    &-viewport {
+    .adf-version-list-viewport {
         height: 100%;
-    }
 
-    .adf-version-list-item {
-        border-bottom: 1px solid #d8d8d8;
+        .adf-version-list-item {
+            height: auto;
+            padding-top: 23.25px;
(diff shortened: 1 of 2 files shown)


g-jaskowski (approved):

lib/content-services/src/lib/version-manager/version-list.component.scss
```diff
@@ -1,10 +1,19 @@
...
-    &-viewport {
+    .adf-version-list-viewport {
         height: 100%;
-    }
 
-    .adf-version-list-item {
-        border-bottom: 1px solid #d8d8d8;
+        .adf-version-list-item {
+            height: auto;
+            padding-top: 23.25px;
```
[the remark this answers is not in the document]
AleksanderSklorz: I've changed it to 23px now

lib/content-services/src/lib/version-manager/version-list.component.scss
```diff
@@ -1,10 +1,19 @@
 .adf-version-list {
-    &-viewport {
+    .adf-version-list-viewport {
```
MichalKinas, line 2: Generally I wonder if this is the right approach, in every other place we rather use ellipsis and title attribute to handle such cases and I think we should go with the same approach here, I would leave the current ellipsis as reported in the ticket and added a title with tooltip if it's not there already, what do you think @AleksanderSklorz

MichalKinas (changes requested):

lib/content-services/src/lib/version-manager/version-list.component.scss
AleksanderSklorz: I discussed that with @akashrathod28 before work on that task and he wanted solution with wrapping text. I can go with any of these solutions, just please let me know @akashrathod28 @MichalKinas which one should I take.

MichalKinas: @akashrathod28 we can discuss this with Agnieszka but wrapping the text in this case would be inconsistent with other places that we have in application

AgnieszkaOpilska: @AleksanderSklorz @akashrathod28 I prefer consistent solutions in the app, so I give my vote to ellipsis and tooltip as @MichalKinas suggested.

AleksanderSklorz: thank you @AgnieszkaOpilska for answer, I will go with tooltip ald ellipsis approach

akashrathod28: I discussed with @AleksanderSklorz  we will go with tooltip approach.

AleksanderSklorz: done, I've pushed correction

MichalKinas (approved):

lib/content-services/src/lib/version-manager/version-list.component.scss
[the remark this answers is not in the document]
AleksanderSklorz: FYI, now approach is completely changed so different styles are applied

dominikiwanekhyland (approved):

PR #9411, opened 2024-03-07
AAE 21001 enable AAE e2es

rafalszmit:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation
> - [x] Other... Please describe: enable e2e

removing C290040, test like this exist in hxp as C286598

> - [ ] Yes
> - [x] No

**Other information**:


e2e/protractor.excludes.json
```diff
@@ -25,8 +25,6 @@
   "C272812": "https://alfresco.atlassian.net/browse/ACS-6425",
   "C274704": "https://alfresco.atlassian.net/browse/ACS-6425",
   "C311425": "https://alfresco.atlassian.net/browse/ACS-6425",
-  "C290040": "https://alfresco.atlassian.net/browse/ACS-6930",
-  "C286598": "https://alfresco.atlassian.net/browse/ACS-6930",
```
acichon-h, line 29: Citing PR description: `removing C290040, test like this exist in hxp as C286598`. I understand that as:
- there's a test `C286598` in `hxp-frontend-apps`, which covers `C290040`
- `C290040` is then redundant
- the conclusion is that `C290040` is to be removed

What this PR does:
- removes `C290040`
- enables `C286598`

I couldn't find `C286598` in hxp-frontend-apps and even if I missed it, I don't think we should have two different tests with the same id, even if they are in different repos.

rafalszmit: I think I made some mistake with copy paste of the test numbers.

- Removing C290040 from ADF -> in hxp its C633731
- Was enabling C286598 here, but need some more investigation as it still fails.

acichon-h: Please ask for rereview once it's passing or you remove enabling it of code from the PR

PR #9423, opened 2024-03-12
[ACS-6150] Add category selector dialog

nikita-web-ua:

> - [ ] Bugfix
> - [X] Feature
> - [ ] Build related changes
> - [ ] Documentation

Folder rules 'Link category' action has a text input that requires the user to enter the category ID manually, which is inconvenient. With this dialog you could select it from a list.

Category selector dialog was added to select one or multiple categories.
adf-category-selector-dialog

> - [ ] Yes
> - [X] No

**Other information**:
https://alfresco.atlassian.net/browse/ACS-6150


lib/content-services/src/lib/category/categories-management/categories-management.component.ts
```diff
@@ -159,13 +163,16 @@ export class CategoriesManagementComponent implements OnInit, OnDestroy {
...
-            .subscribe(() => {
-                this.categories = [];
-                this.categoryNameControlVisible = false;
-                this.categoryNameControlVisibleChange.emit(false);
-            });
+            if (this.classifiableChanged) {
+                this.classifiableChanged
+                    .pipe(takeUntil(this.onDestroy$))
+                    .subscribe(() => {
+                        console.log('classifiableChanged');
```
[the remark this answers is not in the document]
nikita-web-ua: ty

docs/docassets/images/adf-category-selector-dialog.png
AleksanderSklorz, line 1: I have question - do we need that gray background for existing categories? I think it was not looking like that in original design? Was not that changed by mistake during refactoring of properties panel?

nikita-web-ua: It is only for single selection mode, when a category already selected

lib/content-services/src/lib/dialogs/category-selector.dialog.html
```diff
@@ -0,0 +1,28 @@
...
+
+<mat-dialog-content class="adf-dialog-content">
+    <adf-categories-management
+            (categoriesChange)="categories = $event"
+            [categoryNameControlVisible]="true"
+            [managementMode]="categoriesManagementMode"
+            [multiSelect]="multiSelect">
+    </adf-categories-management>
+</mat-dialog-content>
+<mat-dialog-actions align="end" class="acc-categories-dialog-actions">
```
AleksanderSklorz, line 13: I don't see usages of acc-categories-dialog-actions class so can be removed. Also it has acc prefix which is incorrect as we are in ADF

nikita-web-ua: right, removed

docs/docassets/images/adf-category-selector-dialog.png
AleksanderSklorz: ok, I thought in initial design we just make it not clickable. But sure, we can keep that :)

lib/content-services/src/lib/dialogs/category-selector.dialog.html
AleksanderSklorz, line 1: Can we remove dialog for categories from alfresco-applications and reuse this one? It looks very similarly so we could avoid repeating

AleksanderSklorz (approved):

lib/content-services/src/lib/dialogs/category-selector.dialog.html
AleksanderSklorz: we discussed that on slack

MichalKinas (commented):
Overall PR looks good to me, I only wonder should we change the background with the list of categories to plain white for the dialog?

nikita-web-ua:
> Overall PR looks good to me, I only wonder should we change the background with the list of categories to plain white for the dialog?

It is white:) In the example screenshot there is single selection mode and a category already selected, so the rest of the list is disabled.

MichalKinas:
> > Overall PR looks good to me, I only wonder should we change the background with the list of categories to plain white for the dialog?
> 
> It is white:) In the example screenshot there is single selection mode and a category already selected, so the rest of the list is disabled.

Ah okay, thanks for the explanation, all good then :)

MichalKinas (approved):

AleksanderSklorz:
> Overall PR looks good to me, I only wonder should we change the background with the list of categories to plain white for the dialog?

Yeah @MichalKinas I had the same doubts and I added the same comment here: https://github.com/Alfresco/alfresco-ng2-components/pull/9423#discussion_r1521166900

I can see that even Tags have actually that gray background: 
image


but in initial design when we were implementing that there was not gray background if I remember properly, see attached designs and Akash's video after verification in comments in that task: 
https://alfresco.atlassian.net/browse/ACS-4118

docs/docassets/images/adf-category-selector-dialog.png
AleksanderSklorz, line 1: @nikita-web-ua because you are fixing that gray background - could you please also replace that screenshot with updated screenshot with white background? In other case in the future we will get bad reference how it was looking in past

nikita-web-ua: @AleksanderSklorz I fixed background for other use cases, but added new styles for this new dialog, so no need to change the screenshot

AleksanderSklorz: So we are keeping gray background for this new dialog?
@nikita-web-ua @MichalKinas

nikita-web-ua: I can remove it for this one as well if you want, but again it is grey only when there is single selection and a category selected to make it more clear for the user.

MichalKinas: Hmm the other option we might go with is changing the cursor when category is selected already instead of changing the background, potentially it might be more consistent with other places then, maybe we should discuss this with Agnieszka to decide?

MichalKinas: For example we could use `cursor: not-allowed;` there if category is already selected

AleksanderSklorz: I like Michał's suggestion. Also other alternatives: 
- we can still select next category so then if you select next one it will replace this which you selected previously or
- we can close dialog after first selection or,
- we can hide list after first selection 

But yeah, that can be discussed with Agnieszka.

AgnieszkaOpilska: Let's implement the behavior and layout that would be as consistent with the rest of the app as possible. I would keep white background + I'm thinking about one of these 2 solutions:
- changing the cursor when selecting more categories is forbidden 
- we can still select next category so then if you select next one it will replace this which you selected previously
I think the second option is more user friendly. I don't know is simple change of the cursor will make it clear enough to the user, that they still can remove the category and click another one.

PR #9454, opened 2024-03-21
[ACS-7338] Simple search input component

DenysVuika:

> - [x] Tests for the changes have been added (for bug fixes / features)> - [x] Docs have been added / updated (for bug fixes / features)

> - [ ] Bugfix
> - [x] Feature
> - [ ] Build related changes
> - [ ] Documentation

introduce a minimalistic search input to be used with blank ADF-based applications when building Search experience

> - [ ] Yes
> - [x] No

**Other information**:
https://alfresco.atlassian.net/browse/ACS-7338


lib/content-services/src/lib/search/components/search-input/search-input.component.ts
```diff
@@ -0,0 +1,114 @@
...
+        }
+
+        const words = userInput.split(' ');
+
+        if (words.length > 1) {
+            const separator = words.some(this.isOperator) ? ' ' : ' AND ';
+
+            return words
+                .map((term) => {
+                    if (this.isOperator(term)) {
```
MichalKinas, line 80: I think it can be done in one line like ` return this.isOperator(term) ? term : this.formatFields(fields, term);`

lib/content-services/src/lib/search/components/search-input/search-input.component.ts
```diff
@@ -0,0 +1,114 @@
...
+        const input = event.target as HTMLInputElement;
+        const searchTerm = input.value;
+
+        const query = this.formatSearchQuery(searchTerm, this.fields);
+        if (query) {
+            this.changed.emit(decodeURIComponent(query));
+        }
+    }
+
+    private formatSearchQuery(userInput: string, fields = ['cm:name']) {
```
MichalKinas, line 58: Return type is missing

lib/content-services/src/lib/search/components/search-input/search-input.component.ts
```diff
@@ -0,0 +1,114 @@
...
+    value = '';
+
+    @Input()
+    label = 'SEARCH.INPUT.LABEL';
+
+    @Input()
+    placeholder = 'SEARCH.INPUT.PLACEHOLDER';
+
+    @Input()
+    fields: string[] = ['cm:name'];
```
AleksanderSklorz, line 43: no need to set type (type deduced based on ['cm:name'])

lib/content-services/src/lib/search/components/search-input/search-input.component.ts
```diff
@@ -0,0 +1,114 @@
...
+            this.changed.emit(decodeURIComponent(query));
+        }
+    }
+
+    private formatSearchQuery(userInput: string, fields = ['cm:name']) {
+        if (!userInput) {
+            return null;
+        }
+
+        if (/^http[s]?:\/\//.test(userInput)) {
```
AleksanderSklorz, line 63: Intellij hints that this regex can be a bit simplified: 
image

lib/content-services/src/lib/search/components/search-input/search-input.component.ts
```diff
@@ -0,0 +1,105 @@
...
+            this.changed.emit(decodeURIComponent(query));
+        }
+    }
+
+    private formatSearchQuery(userInput: string, fields = ['cm:name']): string {
+        if (!userInput) {
+            return null;
+        }
+
+        if (/^http?:\/\//.test(userInput)) {
```
AleksanderSklorz, line 63: @DenysVuika is that wanted change? Earlier you had http[s] now you have http. Intellij suggested using https because it was fine to skip [] because then '?' is used only for 's' letter without need for []. 

If that is fine to keep just http here then fine :)

AleksanderSklorz (approved):

lib/content-services/src/lib/search/components/search-input/search-input.component.ts
DenysVuika: the code is taken from here https://github.com/Alfresco/alfresco-content-app/blob/develop/projects/aca-content/src/lib/components/search/search-results/search-results.component.ts#L222 and I believe it's testing for `http` or `https` ?

AleksanderSklorz: yes and I think that intellij complains that http[s]? is the same like https? so no need to wrap [s] around s because regex will work in the same way (in both it allows to use http or https). 

But now you skipped s at all in that if, you have just http

DenysVuika: fixed to `https?` as per suggestion https://stackoverflow.com/a/62948640

PR #9794, opened 2024-06-11
AAE-22947 Dropdown widget code should be refactored

wiktord2000:

> - [X] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://hyland.atlassian.net/browse/AAE-22947

> - [ ] Yes
> - [X] No ?

**Other information**:

### DIFF
--- a/lib/core/src/lib/form/components/widgets/core/form-field-validator.spec.ts
+++ b/lib/core/src/lib/form/components/widgets/core/form-field-validator.spec.ts
@@ -16,7 +16,6 @@
  */
 
 import { ErrorMessageModel } from './error-message.model';
-import { FormFieldOption } from './form-field-option';
 import { FormFieldTypes } from './form-field-types';
 import {
     FixedValueFieldValidator,
@@ -68,17 +67,20 @@ describe('FormFieldValidator', () => {
         it('should fail (display error) for dropdown with empty value', () => {
             const field = new FormFieldModel(new FormModel(), {
                 type: FormFieldTypes.DROPDOWN,
-                value: '<empty>',
-                options: [{ id: 'empty', name: 'Choose option...' }],
+                value: null,
+                options: [
+                    { id: 'empty', name: 'Choose option...' },
+                    { id: 'none_empty', name: 'None empty' }
+                ],
                 hasEmptyValue: true,
                 required: true
             });
 
-            field.emptyOption = { id: '<empty>' } as FormFieldOption;
-            expect(validator.validate(field)).toBe(false);
+            expect(field.value).toEqual({ id: 'empty', name: 'Choose option...' });
+            expect(validator.validate(field)).toBeFalsy();
 
-            field.value = '<non-empty>';
-            expect(validator.validate(field)).toBe(true);
+            field.value = { id: 'noneEmpty', name: 'None empty' };
+            expect(validator.validate(field)).toBeTruthy();
         });
 
         it('should fail (display error) for multiple type dropdown with zero selection', () => {
@@ -107,11 +109,12 @@ describe('FormFieldValidator', () => {
                 type: FormFieldTypes.DROPDOWN,
                 value: null,
                 required: true,
-                options: [{ id: 'one', name: 'one' }],
-                selectionType: 'multiple'
+                options: [{ id: 'id_dog', name: 'Dog' }],
+                selectionType: 'multiple',
+                hasEmptyValue: false
             });
 
-            expect(validator.validate(field)).toBe(false);
+            expect(validator.validate(field)).toBeFalsy();
         });
 
         it('should fail (display error) for dropdown with empty object', () => {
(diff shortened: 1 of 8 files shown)


wiktord2000:
This PR includes some code improvements, especially I decided to store directly {id: 'option_id', name: 'Option'} during option selection instead of keeping only 'option_id' and next finding  corresponding option from options if needed. Indeed we had inconsistency between multiple and single mode. When selecting multiple we store array of objects e.g. [{id: 'option_id', name: 'Option'}, {id: 'option_id2', name: 'Option2'}] and in single only the id of selected value. I tried to simplify this code but I'm not sure whether I changed to much, so please try to review in detail.

lib/core/src/lib/form/components/widgets/core/form-field-validator.spec.ts
```diff
@@ -65,63 +64,72 @@ describe('FormFieldValidator', () => {
...
                 required: true,
-                options: [{ id: 'one', name: 'one' }],
-                selectionType: 'multiple'
+                options: [{ id: 'id_dog', name: 'Dog' }],
+                selectionType: 'multiple',
+                hasEmptyValue: false
             });
 
-            expect(validator.validate(field)).toBe(false);
+            expect(validator.validate(field)).toBeFalsy();
```
mauriziovitale, line 117: are u sure about this ? 

toBe(false) checks if the value is strictly equal to false. It will only pass if the value is exactly false.
toBeFalsy() checks if the value is a falsy value. In JavaScript, falsy values include false, 0, '' (empty string), null, undefined, and NaN.

lib/core/src/lib/form/components/widgets/core/form-field.model.ts
```diff
@@ -306,29 +307,34 @@ export class FormFieldModel extends FormWidgetModel {
         /*
          This is needed due to Activiti issue related to reading dropdown values as value string
          but saving back as object: { id: <id>, name: <name> }
+         Side note: Probably not valid anymore
```
mauriziovitale, line 310: what is a comment of a comment? please remove it if not valid.

lib/core/src/lib/form/components/widgets/core/form-field.model.ts
```diff
@@ -230,16 +232,15 @@ export class FormFieldModel extends FormWidgetModel {
             this.leftLabels = form.json.leftLabels || false;
         }
 
-        const emptyOption = Array.isArray(this.options) ? this.options.find(({ id }) => id === 'empty') : undefined;
-        if (this.hasEmptyValue === undefined) {
-            this.hasEmptyValue = json?.hasEmptyValue ?? !!emptyOption;
-        }
+        this.updateForm();
```
mauriziovitale, line 235: was this already inside the get value? please test this code very well it could break several things

lib/core/src/lib/form/components/widgets/core/form-field.model.ts
```diff
@@ -306,29 +307,34 @@ export class FormFieldModel extends FormWidgetModel {
...
-                        value = emptyOption.id;
-                    }
-                } else {
-                    if (value?.id && value?.name) {
-                        value = value.id;
-                    }
-                }
+            if (this.hasEmptyValue && value === null) {
+                value = this.emptyValueOption;
+                return value;
```
mauriziovitale, line 315: why do u return assign and return value instead of just returning this.emptyOption?

lib/process-services-cloud/src/lib/form/components/widgets/dropdown/dropdown-cloud.widget.ts
```diff
@@ -285,7 +294,7 @@ export class DropdownCloudWidgetComponent extends WidgetComponent implements OnI
     }
 
     private resetValue() {
-        this.field.value = '';
+        this.field.value = null;
```
mauriziovitale, line 297: this sounds very scary. until now we were resetting the field value with an empty string ( so it will never be null ) and now u are passing null, why?

lib/core/src/lib/form/components/widgets/core/form-field-validator.spec.ts
wiktord2000: Makes sense

lib/core/src/lib/form/components/widgets/core/form-field.model.ts
wiktord2000: Thx, forgot to change it

lib/process-services-cloud/src/lib/form/components/widgets/dropdown/dropdown-cloud.widget.ts
wiktord2000: I thought that since value is indeed of type **FormFieldOption** (not id - string) I can switch to null

lib/core/src/lib/form/components/widgets/core/form-field.model.ts
wiktord2000: Need more info, don't get it

lib/core/src/lib/form/components/widgets/core/form-field.model.ts
wiktord2000: Still not 100% sure that we can remove this comment, but indeed I asked a lots of people and they can't recall this drawback and in runtime we can't observe this behavior as far as I know.

PR #9799, opened 2024-06-11
[MNT-23194] add long validators, correct filename typos

g-jaskowski:

> - [ ] Bugfix
> - [x] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://hyland.atlassian.net/browse/MNT-23194
Types d:int and d:long types have the exact same validators, causing d:long to have identical restrictions as d:int.

Added d:long type validators so the properties of this type accept wider range of number than Integer range.

> - [ ] Yes
> - [x] No

**Other information**:


lib/content-services/src/lib/content-metadata/services/property-groups-translator.service.spec.ts
```diff
@@ -428,5 +429,27 @@ describe('PropertyGroupTranslatorService', () => {
...
         });
+
+        it('should translate properly the value attribute for d:long', () => {
+            property.dataType = 'd:long';
+
+            propertyValues = { 'FAS:PLAGUE': '1024' };
+            const cardViewGroup = service.translateToCardViewGroups(propertyGroups, propertyValues, null);
+
+            const cardViewProperty: CardViewLongItemModel = cardViewGroup[0].properties[0] as CardViewLongItemModel;
+            expect(cardViewProperty instanceof CardViewLongItemModel).toBeTruthy();
```
swapnil-verma-gl, line 440: I'm not so sure if this assertion is really needed. The cardViewProperty variable is explicitly defined as a type of CardViewLongItemModel, and then the value being assigned to it during initialising is also type casted as CardViewLongItemModel. So having an assertion to verify its type is not validating the code in the component, rather the code in the test case

lib/core/src/lib/card-view/models/card-view-longitem.model.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+        expect(itemModel.isValid('21')).toBe(true, 'For "21" it should be true');
+        expect(itemModel.isValid('21.0')).toBe(true, 'For "21.0" it should be true');
+        expect(itemModel.isValid('2e1')).toBe(true, 'For "2e1" it should be true');
+        expect(itemModel.isValid('2g1')).toBe(false, 'For "2g1" it should be false');
+        expect(itemModel.isValid(21.3)).toBe(false, 'For 21.3 it should be false');
+        expect(itemModel.isValid('21.3')).toBe(false, 'For "21.3" it should be false');
+        expect(itemModel.isValid('text')).toBe(false, 'For "text" it should be false');
+    });
+
+    it('should validate based on defined constraints', () => {
```
swapnil-verma-gl, line 59: Maybe another test case/assertion that should validate a value between the minimum and maximum, and return true

lib/core/src/lib/card-view/validators/card-view-item-long.validator.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export class CardViewItemLongValidator implements CardViewItemValidator {
+    message = 'CORE.CARDVIEW.VALIDATORS.LONG_VALIDATION_ERROR';
+
+    isValid(value: any | any[]): boolean {
+        if (Array.isArray(value)) {
+            return value.every(this.isLongNumber);
+        }
+
+        return value === '' || (!isNaN(value) && this.isLongNumber(value) && this.isNotEmpty(value));
```
swapnil-verma-gl, line 28: Do we really need the isNotEmpty() check? If the passed value was a blank string, wouldn't the method return true, on the basis of the initial value === '' check

lib/content-services/src/lib/content-metadata/services/property-groups-translator.service.spec.ts
g-jaskowski: @swapnil-verma-gl You're right, checking instanceof is pointless with type defined. My bad, it was copy-pasted. Fixed similar occurrences in the whole file.

lib/core/src/lib/card-view/models/card-view-longitem.model.spec.ts
g-jaskowski: added assertion with correct length

lib/core/src/lib/card-view/models/card-view-longitem.model.ts
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+import { CardViewItem } from '../interfaces/card-view-item.interface';
+import { DynamicComponentModel } from '../../common/services/dynamic-component-mapper.service';
+import { CardViewTextItemModel } from './card-view-textitem.model';
+import { CardViewIntItemProperties } from '../interfaces/card-view.interfaces';
+import { CardViewItemLongValidator, CardViewItemPositiveLongValidator } from '../validators/card-view.validators';
+
+export class CardViewLongItemModel extends CardViewTextItemModel implements CardViewItem, DynamicComponentModel {
+    type: string = 'long';
```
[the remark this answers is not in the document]
g-jaskowski: removed in this and alleother card-view models

lib/core/src/lib/card-view/validators/card-view-item-only-positive-long.validator.ts
```diff
@@ -0,0 +1,34 @@
...
+
+    isValid(value: any | any[]): boolean {
+        if (Array.isArray(value)) {
+            return value.every(this.isPositiveNumber);
+        }
+
+        return value === '' || (!isNaN(value) && this.isPositiveNumber(value));
+    }
+
+    private isPositiveNumber(value: any): boolean {
```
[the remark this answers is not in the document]
g-jaskowski: seems to be working correctly as number, changed

g-jaskowski:
Added method allowing different notations of integer number like 1e2 or 1.0 accepted by validators to be used in query. Previously accepted value, but threw a query error. Values like 1.1 are not accepted for d:int and d:long at validators level.

swapnil-verma-gl (approved):

PR #9852, opened 2024-06-21
AAE-23950 Apply reactive forms for date&time and date widgets

tomgny:
…ring specified display format

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

> - [x] Yes
> - [x] No

**Other information**:


lib/core/src/lib/form/components/widgets/date-time/date-time.widget.ts
```diff
@@ -42,68 +43,115 @@ import { WidgetComponent } from '../widget.component';
...
     styleUrls: ['./date-time.widget.scss'],
-    imports: [NgIf, TranslateModule, MatFormFieldModule, MatInputModule, MatDatetimepickerModule, FormsModule, ErrorWidgetComponent],
+    imports: [NgIf, TranslateModule, MatFormFieldModule, MatInputModule, MatDatetimepickerModule, ReactiveFormsModule, ErrorWidgetComponent],
     encapsulation: ViewEncapsulation.None
 })
-export class DateTimeWidgetComponent extends WidgetComponent implements OnInit {
+export class DateTimeWidgetComponent extends WidgetComponent implements OnInit, OnDestroy {
     minDate: Date;
     maxDate: Date;
+    datetimeInputControl: FormControl<Date>;
```
kathrine0, line 52: either provide default values for the properties or declare them with `?`

lib/core/src/lib/form/components/widgets/date-time/date-time.widget.ts
```diff
@@ -42,68 +43,115 @@ import { WidgetComponent } from '../widget.component';
...
 
-    @Input()
-    value: any = null;
+    private datetimeChangesSubscription: Subscription;
```
kathrine0, line 54: why not do takeUntil(this.onDestroy$) instead? Moreover this also should be either initiated with value or declared with `?`

lib/core/src/lib/form/components/widgets/date-time/date-time.widget.ts
```diff
@@ -42,68 +43,115 @@ import { WidgetComponent } from '../widget.component';
     ],
     templateUrl: './date-time.widget.html',
     styleUrls: ['./date-time.widget.scss'],
-    imports: [NgIf, TranslateModule, MatFormFieldModule, MatInputModule, MatDatetimepickerModule, FormsModule, ErrorWidgetComponent],
+    imports: [NgIf, TranslateModule, MatFormFieldModule, MatInputModule, MatDatetimepickerModule, ReactiveFormsModule, ErrorWidgetComponent],
     encapsulation: ViewEncapsulation.None
 })
-export class DateTimeWidgetComponent extends WidgetComponent implements OnInit {
+export class DateTimeWidgetComponent extends WidgetComponent implements OnInit, OnDestroy {
```
kathrine0, line 49: This should be implemented as Angular custom form control with ControlValueAccessor , but I suppose it's not possible with the base WidgetComponent? 
ref. https://blog.angular-university.io/angular-custom-form-controls/

lib/core/src/lib/form/components/widgets/date-time/date-time.widget.ts
```diff
@@ -42,68 +43,115 @@ import { WidgetComponent } from '../widget.component';
     ],
     templateUrl: './date-time.widget.html',
     styleUrls: ['./date-time.widget.scss'],
-    imports: [NgIf, TranslateModule, MatFormFieldModule, MatInputModule, MatDatetimepickerModule, FormsModule, ErrorWidgetComponent],
+    imports: [NgIf, TranslateModule, MatFormFieldModule, MatInputModule, MatDatetimepickerModule, ReactiveFormsModule, ErrorWidgetComponent],
```
kathrine0, line 46: please make it onPush

lib/core/src/lib/form/components/widgets/date-time/date-time.widget.ts
```diff
@@ -42,68 +43,115 @@ import { WidgetComponent } from '../widget.component';
...
-        }
+    private resetErrors(): void {
+        this.updateValidationSummary('');
+    }
 
-        this.value = DateFnsUtils.getDate(this.field.value);
-        this.onFieldChanged(this.field);
+    private initDateAdapter(): void {
+        if (this.field?.dateDisplayFormat) {
+            const dateAdapter = this.dateAdapter as AdfDateFnsAdapter;
```
kathrine0, line 136: Instead of type casting maybe you can define the type when you inject DateAdapter?

lib/core/src/lib/form/components/widgets/date-time/date-time.widget.ts
```diff
@@ -42,68 +43,115 @@ import { WidgetComponent } from '../widget.component';
...
+            dateAdapter.displayFormat = this.field.dateDisplayFormat;
+
+            const dateTimeAdapter = this.dateTimeAdapter as AdfDateTimeFnsAdapter;
```
kathrine0, line 139: Instead of type casting maybe you can define the type when you inject DateTimeAdapter?

lib/core/src/lib/form/components/widgets/date/date.widget.ts
```diff
@@ -60,52 +61,112 @@ export class DateWidgetComponent extends WidgetComponent implements OnInit, OnDe
...
+            this.field.markAsInvalid();
+        } else {
+            this.resetErrors();
+            this.field.markAsValid();
+        }
+    }
+
+    private handleErrors(errors: ValidationErrors): void {
+        const errorAttributes = new Map<string, string>();
+        switch (true) {
```
kathrine0, line 114: For me it would be clearer to use if else syntax here rather than switch(true), but it's a personal preference so the change is up to you

lib/process-services-cloud/src/lib/form/components/form-cloud.component.ts
```diff
@@ -421,6 +422,7 @@ export class FormCloudComponent extends FormBaseComponent implements OnChanges,
             this.displayModeOn.emit(this.displayModeService.findConfiguration(this.displayMode, this.displayModeConfigurations));
         }
 
+        this.changeDetector.detectChanges();
```
kathrine0, line 425: Why do you need to call the detectChanges() here?

lib/process-services-cloud/src/lib/form/components/widgets/date/date-cloud.widget.ts
```diff
@@ -53,74 +78,138 @@ export class DateCloudWidgetComponent extends WidgetComponent implements OnInit,
...
-     * Current date value.
-     * The value is always stored in the format `dd-MM-yyyy`,
-     * but displayed in the UI component using `dateDisplayFormat`
-     */
-    @Input()
-    value: any = null;
+    dateInputControl: FormControl<Date>;
 
-    private onDestroy$ = new Subject<boolean>();
+    private dateChangesSubscription: Subscription;
```
kathrine0, line 83: Please use onDestroy$ subject rather than Subscription. Also, the onDestroy$ subject can be of type void

lib/process-services-cloud/src/lib/form/components/widgets/date/date-cloud.widget.ts
```diff
@@ -53,74 +78,138 @@ export class DateCloudWidgetComponent extends WidgetComponent implements OnInit,
...
     startAt: Date = null;
 
-    /**
-     * Current date value.
-     * The value is always stored in the format `dd-MM-yyyy`,
-     * but displayed in the UI component using `dateDisplayFormat`
-     */
-    @Input()
-    value: any = null;
+    dateInputControl: FormControl<Date>;
```
kathrine0, line 81: initialize value here or in the constructor or make it optional (with ?)

lib/process-services-cloud/src/lib/form/components/widgets/date/date-cloud.widget.ts
```diff
@@ -53,74 +78,138 @@ export class DateCloudWidgetComponent extends WidgetComponent implements OnInit,
...
 
-    private onDestroy$ = new Subject<boolean>();
+    private dateChangesSubscription: Subscription;
 
-    constructor(public formService: FormService, private dateAdapter: DateAdapter<Date>) {
+    constructor(public readonly formService: FormService, private readonly dateAdapter: DateAdapter<Date>) {
```
kathrine0, line 85: can you use the new `inject()` syntax please?

lib/core/src/lib/form/components/widgets/date-time/date-time.widget.ts
tomgny: we need to define this control at the ngOnInit level, because then we have access to the initial value which is the default value of the control. so in that case, shouldn't we use exclamation mark `!` instead making it opitonal by `?`?

kathrine0: You can define the control here and patch it in the ngOnInit when the value is defined. 
Also actually it's another problem, because afaik, value is an input. So to make it 100% we should patch the form every time the value is changed.... given that the base implementation is correct 😅

lib/process-services-cloud/src/lib/form/components/form-cloud.component.ts
tomgny: currently, in case where some field is empty and required at the initial state, we need to update the form state. Because its initial state and no action was performed, form doesn't now that empty required field is invalid, so we need to do it manually at the widgets level (in the ngOnInit). And because we need to do it manually, we change the state after it was checked and we get the Expression has been change after it was checked error. It should be solved once we move the whole form to be reactive instead of manipulate it manually as we currently do.

lib/core/src/lib/form/components/widgets/date/date.widget.ts
tomgny: i think in the future we can implement general mat-errors handler component, that should handle&display errors. For now, we use the validationSummary and custom error-widget component. So once we implement the new component, we can get rid of this code from each widgets.
Issue -> https://hyland.atlassian.net/browse/AAE-24762

lib/core/src/lib/form/components/widgets/date-time/date-time.widget.ts
tomgny: currently, when we add the type at the inject adapter level, we get this error:

`Type 'DateAdapter<any, any>' is missing the following properties from type 'AdfDateFnsAdapter': displayFormat, formatsts(2739)`
so for now, i would prefer to skip additional refactoring around these adapters, we can consider it as a future improvement. is it ok?
Issue -> https://hyland.atlassian.net/browse/AAE-24763

lib/core/src/lib/form/components/widgets/date-time/date-time.widget.ts
tomgny: same as below, Issue -> https://hyland.atlassian.net/browse/AAE-24763

lib/core/src/lib/form/components/widgets/date-time/date-time.widget.ts
tomgny: Agree, but it would be better to do it in separate task, issue -> https://hyland.atlassian.net/browse/AAE-24759

lib/core/src/lib/form/components/widgets/date-time/date-time.widget.ts
tomgny: As we discussed, we should refactor it more and remove extending WidgetComponent. 
Issue -> https://hyland.atlassian.net/browse/AAE-24760

lib/core/src/lib/form/components/widgets/date-time/date-time.widget.ts
tomgny: replaced by onDestroy subject

lib/core/src/lib/form/components/widgets/date-time/date-time.widget.ts
tomgny: defined at declaration level, and added patch in the onInit.
about onChange, i think its not needed (i checked it), because we don't update `this.field` during component lifecycle, so ngOnChanges is not called for this input. Its only used to initialize default values

lib/core/src/lib/form/components/widgets/core/form.model.ts
```diff
@@ -191,8 +191,10 @@ export class FormModel implements ProcessFormModel {
             return;
         }
 
-        if (!field.validate()) {
-            this.markAsInvalid();
+        if (!FormFieldTypes.isReactiveType(field.type)) {
+            if (!field.validate()) {
```
kathrine0, line 195: This can be one if
```suggestion
            if (!FormFieldTypes.isReactiveType(field.type) && !field.validate()) {
```

kathrine0 (approved):

PR #9861, opened 2024-06-24
[ACS-8113] UX bug - Checkbox selections checked state should be Hyland blue not green- edit summary

jacekpluta:
…d blue not green- edit summary

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://hyland.atlassian.net/browse/ACS-8113

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/core/src/lib/button/button.component.html
+++ b/lib/core/src/lib/button/button.component.html
@@ -1,0 +1,120 @@
+<ng-container [ngSwitch]="variant">
+    <ng-container *ngSwitchDefault>
+        <button
+            mat-button
+            [id]="uid"
+            [title]="tooltip"
+            [color]="color"
+            [disabled]="disabled"
+            [disableRipple]="disableRipple"
+            [attr.aria-label]="ariaLabel"
+            [attr.aria-hidden]="ariaHidden"
+            [attr.data-automation-id]="testId"
+        >
+            <ng-container [ngTemplateOutlet]="buttonContent"></ng-container>
+        </button>
+    </ng-container>
+
+    <ng-container *ngSwitchCase="'raised'">
+        <button
+            mat-raised-button
+            [id]="uid"
+            [title]="tooltip"
+            [color]="color"
+            [disabled]="disabled"
+            [disableRipple]="disableRipple"
+            [attr.aria-label]="ariaLabel"
+            [attr.aria-hidden]="ariaHidden"
+            [attr.data-automation-id]="testId"
+        >
+            <ng-container [ngTemplateOutlet]="buttonContent"></ng-container>
+        </button>
+    </ng-container>
+
+    <ng-container *ngSwitchCase="'stroked'">
+        <button
+            mat-stroked-button
+            [id]="uid"
+            [title]="tooltip"
+            [color]="color"
+            [disabled]="disabled"
+            [disableRipple]="disableRipple"
+            [attr.aria-label]="ariaLabel"
+            [attr.aria-hidden]="ariaHidden"
+            [attr.data-automation-id]="testId"
+        >
+            <ng-container [ngTemplateOutlet]="buttonContent"></ng-container>
+        </button>
+    </ng-container>
+
+    <ng-container *ngSwitchCase="'flat'">
+        <button
+            mat-flat-button
+            [id]="uid"
+            [title]="tooltip"
+            [color]="color"
+            [disabled]="disabled"
+            [disableRipple]="disableRipple"
+            [attr.aria-label]="ariaLabel"
+            [attr.aria-hidden]="ariaHidden"
+            [attr.data-automation-id]="testId"
+        >
+            <ng-container [ngTemplateOutlet]="buttonContent"></ng-container>
+        </button>
+    </ng-container>
+
+    <ng-container *ngSwitchCase="'icon'">
+        <button
+            mat-icon-button
+            [id]="uid"
+            [title]="tooltip"
+            [color]="color"
+            [disabled]="disabled"
+            [disableRipple]="disableRipple"
+            [attr.aria-label]="ariaLabel"
+            [attr.aria-hidden]="ariaHidden"
+            [attr.data-automation-id]="testId"
+        >
+            <mat-icon *ngIf="icon">{{ icon }}</mat-icon>
+            <ng-container *ngIf="!icon" [ngTemplateOutlet]="buttonContent"></ng-container>
+        </button>
+    </ng-container>
+
+    <ng-container *ngSwitchCase="'fab'">
+        <button
+            mat-fab
+            [id]="uid"
+            [title]="tooltip"
+            [color]="color"
+            [disabled]="disabled"
+            [disableRipple]="disableRipple"
+            [attr.aria-label]="ariaLabel"
+            [attr.aria-hidden]="ariaHidden"
+            [attr.data-automation-id]="testId"
+        >
+            <mat-icon *ngIf="icon">{{ icon }}</mat-icon>
+            <ng-container *ngIf="!icon" [ngTemplateOutlet]="buttonContent"></ng-container>
+        </button>
+    </ng-container>
+
+    <ng-container *ngSwitchCase="'mini-fab'">
+        <button
+            mat-mini-fab
+            [id]="uid"
+            [title]="tooltip"
+            [color]="color"
+            [disabled]="disabled"
+            [disableRipple]="disableRipple"
+            [attr.aria-label]="ariaLabel"
+            [attr.aria-hidden]="ariaHidden"
+            [attr.data-automation-id]="testId"
+        >
+            <mat-icon *ngIf="icon">{{ icon }}</mat-icon>
+            <ng-container *ngIf="!icon" [ngTemplateOutlet]="buttonContent"></ng-container>
+        </button>
+    </ng-container>
+</ng-container>
+
+<ng-template #buttonContent>
+    <ng-content></ng-content>
+</ng-template>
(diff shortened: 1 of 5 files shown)


docs/content-services/components/document-list.component.md
```diff
@@ -98,19 +98,20 @@ Displays the documents from a repository.
...
-| ready | [`EventEmitter`](https://angular.io/api/core/EventEmitter)`<`[`NodePaging`](https://github.com/Alfresco/alfresco-js-api/blob/develop/src/api/content-rest-api/docs/NodePaging.md)`>` | Emitted when the Document List has loaded all items and is ready for use |
-| columnsVisibilityChanged | [`EventEmitter`](https://angular.io/api/core/EventEmitter)`<{ [columnId: string]: boolean } \| undefined>` | Emitted when columns visibility change |
-| columnsWidthChanged | [`EventEmitter`](https://angular.io/api/core/EventEmitter)`<{ [columnId: string]: number } \| undefined>` | Emitted when columns width change |
-| columnsOrderChanged | [`EventEmitter`](https://angular.io/api/core/EventEmitter)`< string[] \| undefined>` | Emitted when columns order change |
+| Name                      | Type                                                                                                                                                                                  | Description                                                                                                                                                                                             |
```
AleksanderSklorz, line 101: Formatting of that table does not look good. Could you correct?
image

docs/core/components/datatable.component.md
```diff
@@ -448,15 +448,16 @@ Learn more about styling your datatable: [Customizing the component's styles](#c
...
-| rowDblClick         | [`EventEmitter`](https://angular.io/api/core/EventEmitter)`<`[`DataRowEvent`](../../../lib/core/src/lib/datatable/data/data-row-event.model.ts)`>`              | Emitted when the user double-clicks a row.              |
-| showRowActionsMenu  | [`EventEmitter`](https://angular.io/api/core/EventEmitter)`<`[`DataCellEvent`](../../../lib/core/src/lib/datatable/components/data-cell.event.ts)`>`            | Emitted before the actions menu is displayed for a row. |
-| showRowContextMenu  | [`EventEmitter`](https://angular.io/api/core/EventEmitter)`<`[`DataCellEvent`](../../../lib/core/src/lib/datatable/components/data-cell.event.ts)`>`            | Emitted before the context menu is displayed for a row. |
+| Name                      | Type                                                                                                                                                            | Description                                             |
```
AleksanderSklorz, line 451: Here similarly. Please compare with table above that there formatting looks better: 
image

docs/core/components/datatable.component.md
```diff
@@ -448,15 +448,16 @@ Learn more about styling your datatable: [Customizing the component's styles](#c
...
+|---------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------|---------------------------------------------------------|
+| columnOrderChanged        | [`EventEmitter`](https://angular.io/api/core/EventEmitter)`<`[`DataColumn`](../../../lib/core/src/lib/datatable/data/data-column.model.ts)`<>[]>`               | Emitted when the column order is changed.               |
+| columnsWidthChanged       | [`EventEmitter`](https://angular.io/api/core/EventEmitter)`<`[`DataColumn`](../../../lib/core/src/lib/datatable/data/data-column.model.ts)`<>[]>`               | Emitted when the column width is changed.               |
+| selectedItemsCountChanged | [`EventEmitter`](https://angular.io/api/core/EventEmitter)`<`[`DataColumn`](../../../lib/core/src/lib/datatable/data/data-column.model.ts)`<>[]>`               | Emitted when the item row count is changed.             |
```
AleksanderSklorz, line 455: I think here you should have number instead of DataColumn

lib/content-services/src/lib/breadcrumb/breadcrumb.component.html
```diff
@@ -57,7 +57,7 @@
         </a>
 
         <div *ngSwitchDefault class="adf-breadcrumb-item-current" aria-current="location">
-            {{ item.name | translate }}
+            {{selectedRowItemsCount < 1 ? (item.name | translate) : ('BREADCRUMB.HEADER.SELECTED' | translate: { count: selectedRowItemsCount }) }}
```
AleksanderSklorz, line 60: could you please add space before selectedRowItemsCount? Like: 
{{ selectedRowItemsCount

AleksanderSklorz, line 60: I think no need to repeat translate pipe twice. Can be combined into: 
{{ (selectedRowItemsCount < 1 ? item.name : 'BREADCRUMB.HEADER.SELECTED') | translate: { count: selectedRowItemsCount } }}

lib/content-services/src/lib/breadcrumb/breadcrumb.component.spec.ts
```diff
@@ -300,4 +300,26 @@ describe('Breadcrumb', () => {
...
         expect(component.route[3].name).toBe('test-name');
     });
+
+    it('should set title based on selectedRowItemsCount', () => {
+        component.transform = (transformNode) => {
+            transformNode.id = 'test-id';
+            transformNode.name = 'test-name';
+            return transformNode;
+        };
+        (component.folderNode as any) = {
```
AleksanderSklorz, line 310: It can be just: 
component.folderNode = {
  path: { elements: [{ id: 'element-1-id', name: 'element-1-name' }] }
} as Node; 

Will be better than using any.

lib/content-services/src/lib/breadcrumb/breadcrumb.component.spec.ts
```diff
@@ -300,4 +300,26 @@ describe('Breadcrumb', () => {
...
+            path: { elements: [{ id: 'element-1-id', name: 'element-1-name' }] }
+        };
+        component.ngOnChanges();
+        fixture.detectChanges();
+
+        const titleElement = fixture.debugElement.nativeElement.querySelector('.adf-breadcrumb-item-current');
```
AleksanderSklorz, line 316: You repeat these 2 fragments: 
const titleElement = fixture.debugElement.nativeElement.querySelector('.adf-breadcrumb-item-current');

and 

titleElement.textContent.trim()

because the same two fragments are below. Could you please create some reusable function like (something like getBreadcrumbTitle or something and put these 2 fragments there so then that function can just return this string which you can use for expect?

lib/content-services/src/lib/breadcrumb/breadcrumb.component.ts
```diff
@@ -72,6 +82,9 @@ export class BreadcrumbComponent implements OnInit, OnChanges, OnDestroy {
     @Input()
     maxItems: number;
 
+    @Input()
```
AleksanderSklorz, line 85: missing comment which documents that input like for other inputs

lib/content-services/src/lib/document-list/components/document-list.component.ts
```diff
@@ -832,6 +836,10 @@ export class DocumentListComponent extends DataTableSchema implements OnInit, On
         this.columnsWidthChanged.emit(this.columnsWidths);
     }
 
+    onSelectedItemsCountChanged(count: number) {
```
AleksanderSklorz, line 839: could you please write unit test inside unit tests for that component to test if this is function calls emit for selectedItemsCountChanged event?

lib/core/src/lib/datatable/components/datatable/datatable.component.html
```diff
@@ -187,6 +189,7 @@
                 <mat-checkbox [id]="'select-file-' + idx"
                               [checked]="row.isSelected"
                               [attr.aria-checked]="row.isSelected"
+                              [class.adf-checkbox-hidden]="hoveredRowIndex !== idx && !row.isSelected"
```
AleksanderSklorz, line 192: Do we need set this class based on hoveredRowIndex? I see that mat-checkbox is inside of adf-datatable-row so can't we have use :hover in scss file for row and then inside that put styles for checkbox which makes it visible?

docs/content-services/components/document-list.component.md
AleksanderSklorz: oh, right, looks better then, thanks

lib/content-services/src/lib/breadcrumb/breadcrumb.component.html
AleksanderSklorz: @jacekpluta you changed it to number but link still indicates data column model

docs/core/components/datatable.component.md
AleksanderSklorz: @jacekpluta you changed it to number but link still indicates data column model

lib/content-services/src/lib/breadcrumb/breadcrumb.component.spec.ts
```diff
@@ -33,6 +33,9 @@ describe('Breadcrumb', () => {
     });
     let documentListComponent: DocumentListComponent;
 
+    const getBreadcrumbActionText = (fx: ComponentFixture<BreadcrumbComponent>): string =>
+        fx.debugElement.nativeElement.querySelector('.adf-breadcrumb-item-current').textContent.trim();
```
AleksanderSklorz, line 37: This can be simplified even more. You don't need to pass fx (fixture) as parameter because it is defined in whole describe so you have direct access to it. We often use that directly instead of passing by parameter because these tests are for breadcrumb so you won't have any other fixture so no need for parameter.

lib/content-services/src/lib/breadcrumb/breadcrumb.component.ts
AleksanderSklorz: Yes, especially for ADF as these files are open source so it's helpful for external people to have these documentations.

AleksanderSklorz (approved):

PR #9873, opened 2024-06-25
HXCS 3856 File rotation in ADF Viewer

chndn004:

> - [ ] Bugfix
> - [x] Feature
> - [ ] Build related changes
> - [ ] Documentation

The current behavior is that the rotation functionality in the ADF Viewer is limited when in read-only mode. Users are unable to rotate documents within the ADF Viewer when they are in read-only mode. This limitation can be problematic for users who need to view scanned documents that may be displayed in a rotated format.

Jira Card- https://hyland.atlassian.net/browse/HXCS-3856

The new behavior, after implementing the fix using the new flag `allowRotateInReadOnly`, allows users to rotate documents within the ADF Viewer even when they are in read-only mode. This enhancement provides a better user experience as it enables users to easily read scanned documents that may initially appear in a rotated format.

> - [ ] Yes
> - [x] No

**Other information**:

Here how rotating works in readonly mode

https://github.com/Alfresco/alfresco-ng2-components/assets/32226301/e6d8314d-ea9b-4abb-a61f-b5ff3e1e28ba


richardsd (commented):
I'm ok with this solution, but it might be too specific for the `rotate` action use case. 

What if next we want to display other actions that are available in the viewer, lets say, crop for example?

Why can't we display the toolbar actions, and hide the `save` and `cancel` buttons if the editor is in `readOnly`?

lib/core/src/lib/viewer/components/img-viewer/img-viewer.component.ts
```diff
@@ -54,6 +54,9 @@ export class ImgViewerComponent implements AfterViewInit, OnChanges, OnDestroy {
     @Input()
     readOnly = true;
 
+    @Input()
+    allowRotateInReadOnly = false;
```
wideLandscape, line 58: does this value change somewhere?

lib/core/src/lib/viewer/components/viewer-render/viewer-render.component.ts
```diff
@@ -95,6 +95,10 @@ export class ViewerRenderComponent implements OnChanges, OnInit, OnDestroy {
     @Input()
     readOnly = true;
 
+    /** Enable rotation functionality in read-only mode  */
+    @Input()
+    allowRotateInReadOnly = false;
```
wideLandscape, line 100: does this value change somewhere?

lib/core/src/lib/viewer/components/viewer.component.ts
```diff
@@ -189,6 +189,10 @@ export class ViewerComponent<T> implements OnDestroy, OnInit, OnChanges {
     @Input()
     readOnly = true;
 
+    /** Enable rotation functionality in read-only mode  */
+    @Input()
+    allowRotateInReadOnly = false;
```
wideLandscape, line 194: does this value change somewhere?

wideLandscape (commented):
If the value of `allowRotateInReadOnly` never changes and it's always `false`, why we should use it?

chndn004:
> If the value of `allowRotateInReadOnly` never changes and it's always `false`, why we should use it?

The `@Input() allowRotateInReadOnly` is introduced to provide flexibility in controlling the availability of the rotate functionality in read-only mode. While the default value is set to false, this input can be overridden to true in specific use cases where rotation should be allowed even when the component is in read-only mode.

For instance, in the document view within HXW, we will pass `[allowRotateInReadOnly]=true` to the adf-viewer component, as shown below:

```
<adf-viewer
    id="hxp-viewer"
    #viewer
    *ngIf="document"
    [allowFullScreen]="true"
    [allowRightSidebar]="true"
    [showViewer]="true"
    [overlayMode]="true"
    [blobFile]="blobFile$ | async"
    [allowRotateInReadOnly]="true"
    (showViewerChange)="onClose()"
>
</adf-viewer>
```

chndn004:
> What if next we want to display other actions that are available in the viewer, lets say, crop for example?

The readOnlyActions property provides a flexible way to specify which actions are available in read-only mode. By defining readOnlyActions: { rotate: true, crop: false }, we can easily configure which actions should be displayed without modifying the core logic of the component. This approach ensures that future actions, like cropping, can be added or removed from the read-only mode by simply updating the readOnlyActions property.  Please let me know if I can proceed on this.

> Why can't we display the toolbar actions, and hide the `save` and `cancel` buttons if the editor is in `readOnly`?

The save and cancel buttons appear after clicking on the rotate or crop button, which is the current behavior of the application. You can check [here](https://github.com/Alfresco/alfresco-ng2-components/blob/ddf68405118446597c56e1f06919736df93a89d8/lib/core/src/lib/viewer/components/img-viewer/img-viewer.component.ts#L194).

lib/content-services/src/lib/viewer/components/alfresco-viewer.component.html
```diff
@@ -24,6 +24,7 @@
     [urlFile]="urlFileContent"
     [tracks]="tracks"
     [readOnly]="readOnly"
+    [allowRotateInReadOnly]="allowRotateInReadOnly"
```
wideLandscape, line 27: can we be less verbose without losing meaning?
`rotateInReadOnly`
`alwaysRotate`
wdyt?

wideLandscape (approved):
left a nitpick comment

lib/core/src/lib/viewer/components/img-viewer/img-viewer.component.html
```diff
@@ -58,7 +58,7 @@
 
     </adf-toolbar>
 
-    <adf-toolbar class="adf-secondary-toolbar" *ngIf="!readOnly && isEditing">
+    <adf-toolbar class="adf-secondary-toolbar" *ngIf="isEditing && (!readOnly || allowRotateInReadOnly)">
```
Gabez0r, line 61: Why do we want to display the cancel and the save buttons if we in readonly mode?

Gabez0r (commented):
I don't think this approach will scale well in the future. If later we want to allow the same capability for the crop action, or for any other action we come to add, we'd have to add a new boolean property for each action in a few components. I would consider a single property with an object mapping an action to whether it should or not be displayed in read only.

chndn004:
> I don't think this approach will scale well in the future. If later we want to allow the same capability for the crop action, or for any other action we come to add, we'd have to add a new boolean property for each action in a few components. I would consider a single property with an object mapping an action to whether it should or not be displayed in read only.

This is true. We can try this-

```
@Input()
readOnlyActions: { [key: string]: boolean } = {
    rotate: true,
    crop: false
};
```

Please let me know if i can proceed with this.

lib/core/src/lib/viewer/components/img-viewer/img-viewer.component.html
chndn004: This is comes from product side. [MMF](https://hyland.atlassian.net/browse/HXPM-823)

Screenshot 2024-06-27 at 7 19 16 PM

Gabez0r: Then, if we want to make it permanent and save the changes, we will not be in readOnly mode, right?

chndn004: Currently, the adf-viewer operates in two modes:

Read-only: No action toolbar or button toolbar is visible.
Editable: Both toolbars are visible, but the buttons become visible only after starting an action like rotating or cropping.

Our MMF requires activating only the rotate button with the save feature while keeping the viewer in read-only mode. We cannot directly make the viewer editable, as this would also activate cropping, which is not desired in this scenario.

Gabez0r: Enabling "the save feature" seems just like the opposite of being read-only mode, is it not? From what I read, it seems to me that we want to be in editable mode, but also to be able to control which actions can be displayed in this mode, i.e., we'd like to disable the crop action for hxviewer which would use the adf-viewer to edit a document. Is this the case?

chndn004: You're right that enabling the save feature seems contrary to the read-only mode. Instead, we could have the viewer in editable mode but control which actions are available, which would work and be meaningful. However, I am concerned about how this change might impact existing applications that use the viewer in editable mode.

Gabez0r: There's probably more than one way to achieve this. A way would be to add a configuration property to control which actions are enabled on edit mode and, by default, it would have the current configuration, to avoid introducing breaking changes. On HxViewer, this property could be changed to only show the rotate action.

lib/content-services/src/lib/viewer/components/alfresco-viewer.component.ts
```diff
@@ -215,6 +215,10 @@ export class AlfrescoViewerComponent implements OnChanges, OnInit, OnDestroy {
     nodeEntry: NodeEntry;
     tracks: Track[] = [];
     readOnly: boolean = true;
+    allowedActions: { [key: string]: boolean } = {
```
chndn004, line 218: I've implemented the new `allowedActions` property which lets us specify which actions should be available when not in read-only mode.

For `hxviewer`, we can configure it like this:

```
<adf-viewer
    ...
    [readOnly]="false"
    [allowedActions]="{ rotate: true, crop: false }"
    ...>
</adf-viewer>
```

This setup ensures that in `hxviewer`, users can rotate but not crop, even when the viewer is not in read-only mode.

lib/content-services/src/lib/viewer/components/alfresco-viewer.component.html
chndn004: Please check [here] (https://github.com/Alfresco/alfresco-ng2-components/pull/9873#discussion_r1657456365)

lib/core/src/lib/viewer/components/img-viewer/img-viewer.component.html
chndn004: Please check [here] (https://github.com/Alfresco/alfresco-ng2-components/pull/9873#discussion_r1657456365)

lib/content-services/src/lib/viewer/components/alfresco-viewer.component.ts
Gabez0r, line 218: Because this property controls the visibility of some actions only on edit mode, perhaps the name could be more specific, like `allowedEditActions`? We also have other actions that are always available on readonly mode, like the zoom in/out and reset, so just "allowedActions" could be misleading.

Gabez0r (commented):
Well done. We should probably also update the docs `docs/core/components/viewer.component.md`.

As note, if you are not doing it yet, testing it with hxviewer before merging would be great to avoid going back and forth with public components.

lib/content-services/src/lib/viewer/components/alfresco-viewer.component.ts
chndn004: Done

chndn004:
> Well done. We should probably also update the docs `docs/core/components/viewer.component.md`.

Done.

> As note, if you are not doing it yet, testing it with hxviewer before merging would be great to avoid going back and forth with public components.

I'm unsure how to integrate the updated `adf-viewer` into `hxviewer` locally without creating ADF release. Could you please provide guidance or suggest a plan for implementing this?

richardsd (commented):
It looks good 💪

Gabez0r:
> I'm unsure how to integrate the updated adf-viewer into hxviewer locally without creating ADF release. Could you please provide guidance or suggest a plan for implementing this?

richardsd:
> > I'm unsure how to integrate the updated adf-viewer into hxviewer locally without creating ADF release. Could you please provide guidance or suggest a plan for implementing this?
> 
> @richardsd, do you know how to do this? Is this possible like we do it for workspace-hxp?

@Gabez0r It should be enough to launch the application with the adf target (assuming ADF repository is side-by-side with hxp-frontend-apps).

```bash
npx nx run hxviewer:serve:adf
```

Folder structure example:

- repositories
-- alfresco-ng2-components
-- hxp-frontend-apps

Gabez0r (approved):

richardsd (approved):

DenysVuika (approved):

PR #9904, opened 2024-07-03
[ACS-8081] People facet - update the query on apply hit

nikita-web-ua:

> - [X] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://hyland.atlassian.net/browse/ACS-8081
The search query is updated when options are selected, causing the list of options to be updated based on the results.

The search query is updated when you click the Apply button, allowing you to select multiple options before returning the result.

> - [ ] Yes
> - [X] No

**Other information**:

### DIFF
--- a/lib/content-services/src/lib/search/components/search-filter-chips/search-facet-chip-tabbed/search-facet-tabbed-content.component.spec.ts
+++ b/lib/content-services/src/lib/search/components/search-filter-chips/search-facet-chip-tabbed/search-facet-tabbed-content.component.spec.ts
@@ -23,7 +23,6 @@ import { FacetField } from '../../../models/facet-field.interface';
 import { SearchFacetFiltersService } from '../../../services/search-facet-filters.service';
 import { NO_ERRORS_SCHEMA, SimpleChange } from '@angular/core';
 import { SearchFacetTabbedContentComponent } from './search-facet-tabbed-content.component';
-import { of } from 'rxjs';
 import { HarnessLoader } from '@angular/cdk/testing';
 import { TestbedHarnessEnvironment } from '@angular/cdk/testing/testbed';
 import { MatTabGroupHarness, MatTabHarness } from '@angular/material/tabs/testing';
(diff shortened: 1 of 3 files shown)


MichalKinas (approved):

lib/content-services/src/lib/search/components/search-filter-chips/search-facet-chip-tabbed/search-facet-tabbed-content.component.spec.ts
```diff
@@ -189,15 +190,21 @@ describe('SearchFacetTabbedContentComponent', () => {
...
         component.reset();
-        expect(queryBuilder.update).toHaveBeenCalled();
+        expect(queryBuilderUpdateSpy).toHaveBeenCalled();
         expect(component.updateDisplayValue).toHaveBeenCalled();
     });
+
+    it('should not call queryBuilder.update on options change', () => {
+        expect(queryBuilderUpdateSpy.calls.count()).toBe(2);
+        component.onOptionsChange([{ value: 'test' }], 'field');
+        expect(queryBuilderUpdateSpy.calls.count()).toBe(2);
```
swapnil-verma-gl, line 208: I guess the logic behind the test is that since the number of calls to the function before onOptionsChange and after onOptionsChange is the same, then the function has not been called. However, I'm curious as to why the method is being called twice. I would have expected it to have been called 0 times (ideal) or once (maybe its needed during init).
Also, can you check and see if we can use toHaveBeenCalledBefore () here instead. That would ensure that if in the future the number of calls to queryBuilder.update change for whatever reason, then we won't have to update this test again. I guess the code might look something like this (untested) - 
```
it('should not call queryBuilder.update on options change', () => {
        spyOn(component, 'onOptionsChange').and.callThrough();
        component.onOptionsChange([{ value: 'test' }], 'field');
        expect(queryBuilderUpdateSpy).toHaveBeenCalledBefore(component.onOptionsChange);
});
```

nikita-web-ua: rewrote the test, thanks

swapnil-verma-gl (approved):

dominikiwanekhyland (approved):

DenysVuika (approved):

PR #9941, opened 2024-07-11
ACS-8393: actions service (lightweight pub/sub)

DenysVuika:

> - [ ] Bugfix
> - [x] Feature
> - [ ] Build related changes
> - [ ] Documentation

Lightweight pub/sub service for UI components and applications

> - [ ] Yes
> - [x] No

**Other information**:
https://hyland.atlassian.net/browse/ACS-8393

### DIFF
--- a/lib/core/src/lib/services/action.service.ts
+++ b/lib/core/src/lib/services/action.service.ts
@@ -1,0 +1,42 @@
+/*!
+ * @license
+ * Copyright © 2005-2024 Hyland Software, Inc. and its affiliates. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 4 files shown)


lib/core/src/lib/services/action.service.ts
```diff
@@ -0,0 +1,42 @@
...
+import { BehaviorSubject, Observable } from 'rxjs';
+import { filter } from 'rxjs/operators';
+
+export interface Action {
+    type: string;
+    payload?: any;
+}
+
+@Injectable({ providedIn: 'root' })
+export class ActionService {
```
DaryaBalvanovich, line 28: Could you please explain why do we need it and how it should simplify redux usage?

DenysVuika: The applications that do not use Redux can still use the event bus service if needed... the NgRx library has a learning curve and not all the teams are using or willing to use that, the ADF has no dependencies on NgRx as well

DenysVuika:
I will move this PR to ACA for now

PR #9946, opened 2024-07-12
AAE-22948 Radio buttons should not display errors / call apis when they're in readOnly state

wiktord2000:

> - [X] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

AAE-22948

> - [ ] Yes
> - [X] No

**Other information**:

### DIFF
--- a/lib/core/src/lib/form/components/widgets/core/form-field.model.spec.ts
+++ b/lib/core/src/lib/form/components/widgets/core/form-field.model.spec.ts
@@ -124,6 +124,22 @@ describe('FormFieldModel', () => {
             expect(field.value).toEqual('id_one');
         });
 
+        it('should add value (selected options) to field options if NOT present (multiple selection)', () => {
+            const selectedOptions = [
+                { id: 'id_one', name: 'One' },
+                { id: 'id_two', name: 'Two' }
+            ];
+            const field = new FormFieldModel(new FormModel(), {
+                type: FormFieldTypes.DROPDOWN,
+                options: [],
+                value: selectedOptions,
+                selectionType: 'multiple'
+            });
+
+            expect(field.options).toEqual(selectedOptions);
+            expect(field.value).toEqual(selectedOptions);
+        });
+
         it('should assign "empty" option as value if value is null and "empty" option is present in options', () => {
             const field = new FormFieldModel(new FormModel(), {
                 type: FormFieldTypes.DROPDOWN,
@@ -531,56 +547,43 @@ describe('FormFieldModel', () => {
     });
 
     describe('radio buttons field model instantiation', () => {
+        const mockOptions = [
+            { id: 'opt1', name: 'Option 1' },
+            { id: 'opt2', name: 'Option 2' }
+        ];
+
         describe('should parse and resolve selected option id in case of', () => {
             it('string - id', () => {
                 const field = new FormFieldModel(new FormModel(), {
                     type: FormFieldTypes.RADIO_BUTTONS,
-                    options: [
-                        { id: 'opt1', name: 'Option 1' },
-                        { id: 'opt2', name: 'Option 2' }
-                    ],
+                    options: mockOptions,
                     value: 'opt2'
                 });
 
                 expect(field.value).toBe('opt2');
-                expect(field.options).toEqual([
-                    { id: 'opt1', name: 'Option 1' },
-                    { id: 'opt2', name: 'Option 2' }
-                ]);
+                expect(field.options).toEqual(mockOptions);
             });
 
             it('string - name', () => {
                 const field = new FormFieldModel(new FormModel(), {
                     type: FormFieldTypes.RADIO_BUTTONS,
-                    options: [
-                        { id: 'opt1', name: 'Option 1' },
-                        { id: 'opt2', name: 'Option 2' }
-                    ],
+                    options: mockOptions,
                     value: 'Option 1'
                 });
 
                 expect(field.value).toBe('opt1');
-                expect(field.options).toEqual([
-                    { id: 'opt1', name: 'Option 1' },
-                    { id: 'opt2', name: 'Option 2' }
-                ]);
+                expect(field.options).toEqual(mockOptions);
             });
 
             it('object with id and name', () => {
                 const field = new FormFieldModel(new FormModel(), {
                     type: FormFieldTypes.RADIO_BUTTONS,
-                    options: [
-                        { id: 'opt1', name: 'Option 1' },
-                        { id: 'opt2', name: 'Option 2' }
-                    ],
+                    options: mockOptions,
                     value: { id: 'opt2', name: 'Option 2' }
                 });
 
                 expect(field.value).toBe('opt2');
-                expect(field.options).toEqual([
-                    { id: 'opt1', name: 'Option 1' },
-                    { id: 'opt2', name: 'Option 2' }
-                ]);
+                expect(field.options).toEqual(mockOptions);
             });
 
             it('object with id, name and options', () => {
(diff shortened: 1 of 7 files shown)


lib/core/src/lib/form/components/widgets/core/form-field.model.spec.ts
```diff
@@ -277,7 +305,8 @@ describe('FormFieldModel', () => {
         const currentDate = new Date();
         const expectedDateFormat = `${DateFnsUtils.formatDate(currentDate, 'yyyy-MM-dd')}T00:00:00.000Z`;
 
-        expect(field.value).toEqual(currentDate);
+        expect(field.value.getDate()).toEqual(currentDate.getDate());
+        expect(field.value.getHours()).toEqual(currentDate.getHours());
```
wiktord2000, line 309: It is the thing mentioned by Tom [here](https://github.com/Alfresco/alfresco-ng2-components/pull/9939/files) (flaky unit test)

wiktord2000: Needed to pass units (not related with actual issue)

lib/core/src/lib/form/components/widgets/core/form-field.model.spec.ts
```diff
@@ -112,67 +112,95 @@ describe('FormFieldModel', () => {
...
+            it('when object', () => {
+                const field = new FormFieldModel(new FormModel(), {
+                    type: FormFieldTypes.DROPDOWN,
+                    options: [],
+                    value: { id: 'delayed-option-id', name: 'Delayed option' }
+                });
+                expect(field.value).toBe('delayed-option-id');
+            });
+        });
     });
```
wiktord2000, line 204: Reorganization of dropdown tests - make them similar to radio ones, indeed describe block added and **should parse and leave dropdown value as is** changed to **should leave not resolved value (in case of delayed options)** other tests remained the same

lib/core/src/lib/form/components/widgets/core/form-field.model.spec.ts
```diff
@@ -501,41 +530,112 @@ describe('FormFieldModel', () => {
...
+            });
 
-    it('should parse and leave radio button value as is', () => {
-        const field = new FormFieldModel(new FormModel(), {
-            type: FormFieldTypes.RADIO_BUTTONS,
-            options: [],
-            value: 'deferred-radio'
+            expect(field.options).toEqual([{ id: 'opt1', name: 'Option 1' }]);
+            expect(field.value).toEqual('opt1');
         });
```
wiktord2000, line 638: Similar what we do for dropdown

wiktord2000: When we will preview completed forms with radio buttons from older projects we will see at least selected option, the newly created will store full list of options

lib/core/src/lib/form/components/widgets/core/form-field.model.ts
```diff
@@ -414,7 +427,9 @@ export class FormFieldModel extends FormWidgetModel {
             }
             case FormFieldTypes.RADIO_BUTTONS: {
                 const radioButton: FormFieldOption = this.options.find((opt) => opt.id === this.value);
-                this.form.values[this.id] = radioButton || null;
+
+                this.form.values[this.id] = radioButton ? { ...radioButton, options: this.options } : { id: null, name: null, options: this.options };
+
                 break;
```
wiktord2000, line 433: The most important part, related to this comment:
image

wiktord2000: We used **options** word when defining object representing radio buttons value but we can change it if needed (can be slightly confusing I guess) e.g. allOptions, optionContext, optionsOnSubmit

lib/core/src/lib/form/components/widgets/core/form-field.model.ts
```diff
@@ -341,14 +341,27 @@ export class FormFieldModel extends FormWidgetModel {
...
+            if (matchingOption) {
+                return matchingOption.id;
+            }
+
+            if (this.isValidOption(value)) {
+                this.addOption({ id: value.id, name: value.name });
+                return value.id;
+            }
+
+            return value;
```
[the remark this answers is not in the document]
wiktord2000: It's 8 years old code (some Activity comments there which made it more unclear what we want to achieve) and I'm not sure what was the reason of returning just value and not caring about its type. After investigation (spent more then month of delving into dropdown/radio buttons widgets), can say they this type should be **string** representing selected **id or name/label** (some comments suggest it).

Part of **parseValue** method only called only once on constructor:
image

What we can suppose as well, we want to keep this value (id or name of selected option) in case we will receive full list of options later e.g. from rest api call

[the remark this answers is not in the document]
wiktord2000: Indeed our current UI for Radio Buttons Widget is prepared for **selected option id** (just string) but from requested json we can receive:
- selected option **id (string)**
- selected option **name (string)**
- full object of selected option **object { id: string, name: string }**
- full object of selected option + options **{ id: string, name: string, options: [{ id: string, name: string }, { ... }] } -> when my PR will be merged**

Based on what I know so far.

tomgny (commented):
what about the mapping (input/output)? do these changes affect it (and can break it)?

wiktord2000:
> what about the mapping (input/output)? do these changes affect it (and can break it)?

as discussed the risk should be small with this change we will receive **{ id: ... , name: ..., options: [{ ... }{ ... }] }** as an output of RadioButtons widget (which can be mapped e.g. to the json process variable), instead of: **{ id: ... , name: ... }**. We just extending information so hopefully we won't brake anything, we assume that we ware using json only to target id or name e.g. ${radioJsonVar.id} / ${radioJsonVar.name} (Expression tab)

lib/core/src/lib/form/components/widgets/core/form-field.model.ts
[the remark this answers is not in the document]
wiktord2000: I grater detail we do a lots of checks in the scope of this function: 
```
function parseValue(json: any): any {
    if (json.type === FormFieldTypes.DROPDOWN) { return parsedValueForDropdown }
    if (json.type === FormFieldTypes.RADIO_BUTTONS) { return parsedValueForRadio }
    if (this.isDateField(json) || this.isDateTimeField(json)) { return paredValueForDate } 
    if (this.isCheckboxField(json)) { return parsedValueForCheckbox }
}
```
For sure this code should be split e.g. 
`if (json.type === FormFieldTypes.DROPDOWN) { return parseDropdownValue(json.value) }`

where

`function parseDropdownValue(value: any): ParsedDropdownValue { ... }`

with this in place we could improve parseValue function as follows:
```
function parseValue(json: any): ParsedDropdownValue | ParsedRadioValue | ... {
    if (json.type === FormFieldTypes.DROPDOWN) { return parseDropdownValue(json.value) }
    if (json.type === FormFieldTypes.DROPDOWN) { return parseRadioValue(json.value) }
    ...
}
```

But for now I won't cluttered the code with these changes, nonetheless it's a good point in regards of refactoring.

wiktord2000:
FYI: I had small sync with @DenysVuika about radio / dropdown files without *-cloud suffix, and he mentioned that these files also should be updated accordingly.

lib/core/src/lib/form/components/widgets/core/form-field.model.ts
wiktord2000: **Update:** 
This convention will be followed only by radio buttons of **rest** type, for **manual** ones we will stick with the old implementation. 

Reason: 
- Manual radio can't store **options** property (runtime issue when submitting form) (no problems in case of **rest** type)
- Actually storing options for **manual** type doesn't have any sense, when specific process is triggered the static options list is fixed (releasing new version of the project can't impact this list e.g. when we will change options)
- Static options don't have to be requested in separate call, we have them out of the box

mauriziovitale (approved):

kathrine0 (approved):

tomgny (approved):

PR #9959, opened 2024-07-18
AAE-24018 Readonly dropdown with the rest config don't fetch options when form rule change it to not readonly

wiktord2000:

> - [X] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

AAE-24018

https://github.com/user-attachments/assets/df688bc9-262e-476a-88a8-5abfb04dcb02

**Note:** We have form rule here: Enable dropdown when Text field has '1', otherwise is disabled. 

https://github.com/user-attachments/assets/9e7ae15c-a15c-44ae-b4f1-6c8615acd525

**Note:** We have form rule here: Enable dropdown when Text field has '1', otherwise is disabled. 

> - [ ] Yes
> - [X] No

**Other information**:

### DIFF
--- a/lib/process-services/src/lib/form/widgets/dropdown/dropdown.widget.ts
+++ b/lib/process-services/src/lib/form/widgets/dropdown/dropdown.widget.ts
@@ -52,7 +52,7 @@ export class DropdownWidgetComponent extends WidgetComponent implements OnInit {
     }
 
     ngOnInit() {
-        if (this.field?.restUrl) {
+        if (this.field?.restUrl && !this.field?.form?.readOnly) {
             if (this.field.form.taskId) {
                 this.getValuesByTaskId();
             } else {
@@ -101,5 +101,9 @@ export class DropdownWidgetComponent extends WidgetComponent implements OnInit {
 
     showRequiredMessage(): boolean {
         return (this.isInvalidFieldRequired() || this.field.value === 'empty') && this.isTouched();
+    }
+
+    get isReadOnlyField(): boolean {
+        return this.field.readOnly;
     }
 }
--- a/lib/process-services/src/lib/form/widgets/dropdown/dropdown.widget.ts
+++ b/lib/process-services/src/lib/form/widgets/dropdown/dropdown.widget.ts
@@ -52,7 +52,7 @@ export class DropdownWidgetComponent extends WidgetComponent implements OnInit {
     }
 
     ngOnInit() {
-        if (this.field?.restUrl && !this.field?.form?.readOnly) {
+        if (this.field?.restUrl && !this.isReadOnlyForm()) {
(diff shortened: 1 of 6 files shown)


wiktord2000:
**Note:** With the current code is default to call api only when field is enabled by the applied rule (it would be ideal I guess), so the call is done at the form init even the dropdown is disabled.

wiktord2000:
The actual issue refers to cloud widgets but I guess I have to update regular ones as well, I can remove the changes if I'am mistaken in **dropdown.widget.*** files

kathrine0:
> **Note:** With the current code is difficult to call api only when field is not readonly by the applied rule (it would be ideal I guess), so the call is done at the form init even the dropdown is readonly at that time.

Can't we track the readonly value change and query the api only then? Why is it not possible?

wiktord2000:
> > **Note:** With the current code is difficult to call api only when field is not readonly by the applied rule (it would be ideal I guess), so the call is done at the form init even the dropdown is readonly at that time.
> 
> Can't we track the readonly value change and query the api only then? Why is it not possible?

After small investigation, we relay on **@Input() field** from inherited component class -> [WidgetComponent](https://github.com/Alfresco/alfresco-ng2-components/blob/eda03ed525044e1b337d8c185eebeb09462d73cf/lib/core/src/lib/form/components/widgets/widget.component.ts#L52C1-L52C27), and readonly state is determined by **field.readOnly** property (when this rule will be fulfilled this value will change), I tried to use **ngOnChanges** hook on that but seems that we still relay on the same reference hence we can't track its changes. The benefit of current implementation is that we always call once when we will call every time readonly has changed it could lead to multiple api calls, which are not needed

lib/process-services/src/lib/form/widgets/dropdown/dropdown.widget.ts
```diff
@@ -52,7 +52,7 @@ export class DropdownWidgetComponent extends WidgetComponent implements OnInit {
     }
 
     ngOnInit() {
-        if (this.field?.restUrl) {
+        if (this.field?.restUrl && !this.field?.form?.readOnly) {
```
wiktord2000, line 55: Will move **this.field?.form?.readOnly** to the function as cloud one does

wiktord2000:
```
    private isReadOnlyForm(): boolean {
        return !!this.field?.form?.readOnly;
    }
```

kathrine0 (approved):

lib/process-services/src/lib/form/widgets/dropdown/dropdown.widget.ts
wiktord2000: Done

PR #9977, opened 2024-07-23
AAE-23358: Fix allowing to bypass required form on start event by cli…

kathrine0:
…cking the start button before form is loaded

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://hyland.atlassian.net/browse/AAE-23358

> - [ ] Yes
> - [x] No

**Other information**:

Before:

https://github.com/user-attachments/assets/174a6362-4163-4263-87c5-c287e9c76c27

After:

https://github.com/user-attachments/assets/5d063e9a-2ea6-47e4-bc45-74158fbebdb6


lib/process-services-cloud/src/lib/process/start-process/components/start-process-cloud.component.ts
```diff
@@ -115,39 +116,65 @@ export class StartProcessCloudComponent implements OnChanges, OnInit, OnDestroy
...
-        private processNameCloudPipe: ProcessNameCloudPipe
-    ) {}
+    formCloud?: FormModel;
+    processForm = new FormGroup({
+        processInstanceName: new FormControl('', [
+            Validators.required,
+            Validators.maxLength(this.getMaxNameLength()),
+            Validators.pattern('^[^\\s]+(\\s+[^\\s]+)*$')
+        ]),
+        processDefinition: new FormControl('', [Validators.required, this.processDefinitionNameValidator()])
```
mauriziovitale, line 140: did u pass empty by purpose instead of this.processDefinitionName ?

mauriziovitale: we could remove 
```
this.processDefinition.setValue(this.processDefinitionName);

**Review by mauriziovitale — APPROVED** [review]

File: lib/process-services-cloud/src/lib/process/start-process/components/start-process-cloud.component.ts (hunk 1 above)
Comment (kathrine0) [reply lib/process-services-cloud/src/lib/process/start-process/components/start-process-cloud.component.ts:140]: The form is now initiated on the class level and `processDefinitionName` comes from an Input. So at the moment of the initialization, the `processDefinitionName` doesn't have a value yet

File: lib/process-services-cloud/src/lib/process/start-process/components/start-process-cloud.component.spec.ts [hunk 2]
```diff
@@ -968,27 +977,27 @@ describe('StartProcessCloudComponent', () => {
...
             await fixture.whenStable();
 
             const startButton = fixture.debugElement.query(By.css('#button-start'));
             expect(startButton).not.toBeNull();
-            expect(component.disableStartButton()).toBeFalse();
+            expect(component.disableStartButton).toBeFalse();
             expect((startButton.nativeElement as HTMLButtonElement).disabled).toBeFalse();
         });
 
         it('start process button should be disabled when isLoading is true', async () => {
```
wiktord2000, line 990: Please update test name, isLoading is no longer used

wiktord2000: Please check other test names as well

lib/process-services-cloud/src/lib/process/start-process/components/start-process-cloud.component.ts
```diff
@@ -254,13 +280,7 @@ export class StartProcessCloudComponent implements OnChanges, OnInit, OnDestroy
 
         this.startProcessCloudService
```
wiktord2000, line 281: When the API call for definitions fails, will we display the running spinner infinitely?

lib/process-services-cloud/src/lib/process/start-process/components/start-process-cloud.component.ts
```diff
@@ -115,39 +116,65 @@ export class StartProcessCloudComponent implements OnChanges, OnInit, OnDestroy
...
+
+    private readonly startProcessCloudService = inject(StartProcessCloudService);
+    private readonly processNameCloudPipe = inject(ProcessNameCloudPipe);
+
+    get isProcessFormValid(): boolean {
+        if (this.hasForm && this.isFormCloudLoaded) {
+            return (this.formCloud ? !Object.keys(this.formCloud.values).length : false) || this.formCloud?.isValid || this.isProcessStarting;
+        } else {
+            return this.processForm.valid || this.isProcessStarting;
+        }
```
wiktord2000, line 151: else block is not needed in this case

lib/process-services-cloud/src/lib/process/start-process/components/start-process-cloud.component.ts
```diff
@@ -115,39 +116,65 @@ export class StartProcessCloudComponent implements OnChanges, OnInit, OnDestroy
...
-    ) {}
+    formCloud?: FormModel;
+    processForm = new FormGroup({
+        processInstanceName: new FormControl('', [
+            Validators.required,
+            Validators.maxLength(this.getMaxNameLength()),
+            Validators.pattern('^[^\\s]+(\\s+[^\\s]+)*$')
+        ]),
+        processDefinition: new FormControl('', [Validators.required, this.processDefinitionNameValidator()])
+    });
```
wiktord2000, line 141: Is it the better approach to no use FormBuilder service?

lib/process-services-cloud/src/lib/process/start-process/components/start-process-cloud.component.ts
kathrine0: Probably yes 🤷‍♀️. As already talked with Maurizio, this component really needs to be rewritten. I only focused on a bug fix itself and some minor non-breaking improvements

lib/process-services-cloud/src/lib/process/start-process/components/start-process-cloud.component.ts
```diff
@@ -254,13 +280,7 @@ export class StartProcessCloudComponent implements OnChanges, OnInit, OnDestroy
...
         this.startProcessCloudService
             .getProcessDefinitions(this.appName)
-            .pipe(
-                tap(() => {
-                    this.processDefinitionLoaded = true;
-                    this.loading$.next(false);
-                }),
-                takeUntil(this.onDestroy$)
-            )
+            .pipe(takeUntil(this.onDestroy$))
```
wiktord2000, line 283: I was wondering (when working on my PR), whether we can add take(1) / takeFirst() in this case

lib/process-services-cloud/src/lib/process/start-process/components/start-process-cloud.component.ts
kathrine0: It depends. FormBuilder gives you more possibilities (e.g. nonNullableForm, simplified grouping) but for basic usage you can simply use new FormGroup.
https://angular.dev/guide/forms/reactive-forms#create-a-formgroup-instance 
https://angular.dev/guide/forms/reactive-forms#using-the-formbuilder-service-to-generate-controls

lib/process-services-cloud/src/lib/process/start-process/components/start-process-cloud.component.ts
kathrine0: Are you sure about it? I took the original implementation and only refactored it to getter. As mentioned before, I don't want to introduce any breaking change.

lib/process-services-cloud/src/lib/process/start-process/components/start-process-cloud.component.ts
wiktord2000: thx, sure, resolved

lib/process-services-cloud/src/lib/process/start-process/components/start-process-cloud.component.ts
wiktord2000: Actually, we do here if(sth){ return }, else { return } so I'm 100% sure it won't break anything but it's up to you, you can leave it as well

wiktord2000: (old imp)
image

wiktord2000: it's code refactor thing

lib/process-services-cloud/src/lib/process/start-process/components/start-process-cloud.component.ts
wiktord2000: Makes sense, we will save one inject() with this approach so we'll benefit on that, resolved

lib/process-services-cloud/src/lib/process/start-process/components/start-process-cloud.component.ts
kathrine0: Ah, you don't mean removing `return this.processForm.valid || this.isProcessStarting;` but you mean to remove only `else { }`. Sure I could but what for? ;) It's a cosmetic change and as I said already, what I'd love to do here is to create a v2 of this component

lib/process-services-cloud/src/lib/process/start-process/components/start-process-cloud.component.ts
kathrine0: I think it should work, in fact for one-time API calls it's event better than takeUntil

lib/process-services-cloud/src/lib/process/start-process/components/start-process-cloud.component.ts
wiktord2000: Sure, maybe it's only my thing that i don't like use else with direct return, resolved

lib/process-services-cloud/src/lib/process/start-process/components/start-process-cloud.component.ts
wiktord2000: but I wondering what will happen when we will replace takeUntil(this.onDestroy$) with only take(1) at the time we will lose component from the view and don't receive response yet, guessing that this subscribing will be completed only when we receive something not when component is out of the view, so probably even we will use take(1) we should still keep takeUntil(this.onDestroy$).
(gpt)
image

kathrine0: This is actually interesting topic that you brought up. Usually I was using first() with httpClient calls (and even that is redundant as http calls usually produce one result and complete), however there are cases that it would be better to cancel the subscription with onDestroy$.
I came across this discussion on stackoverflow, I encourage you to take a look to see various edge cases people describe when unsubscribing is a better option: https://stackoverflow.com/questions/35042929/is-it-necessary-to-unsubscribe-from-observables-created-by-http-methods-to-avoid

wiktord2000: Thank you for reminder and ref, I just forgot that some of the Angular services can automatically unsubscribe after one value being received. Like a HttpClient does, but as it was touched in discussion, we still should relay at least on takeUntil(onDestroy$), resolved

PR #10051, opened 2024-08-06
[ACS-8502] missing alignment for menu and search button

dominikiwanekhyland:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://hyland.atlassian.net/browse/ACS-8502

> - [ ] Yes
> - [ ] No

**Other information**:

### DIFF
--- a/lib/core/src/lib/search-text/search-text-input.component.scss
+++ b/lib/core/src/lib/search-text/search-text-input.component.scss
@@ -9,7 +9,7 @@
         display: none;
     }
 
-    .adf-search-animation-container {
+    .adf-search-container-transition {
         display: flex;
         align-items: center;
     }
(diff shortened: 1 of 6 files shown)


DaryaBalvanovich (approved):

lib/core/src/lib/search-text/search-text-input.component.scss
```diff
@@ -5,15 +5,16 @@
 }
 
 .adf-search-button {
-    left: -13px;
+    left: -20px;
+    top: 16px;
```
MichalKinas, line 9: This will affect any other app using this component let's not change it in this way, if we want to fix it it should be fixed by fixing alignement for the whole component not only by moving an icon

dominikiwanekhyland: image
@MichalKinas so it's actually an icon misplaced inside the component, so I think that in other apps using adf it would be visible as broken

PR #10074, opened 2024-08-09
[ACS-7600] Legal hold - Hold interface and documentation fixes 

tamaragruszka:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

required fields were made as optional

required fields are required

> - [ ] Yes
> - [x] No

**Other information**:


lib/js-api/src/api/gs-core-rest-api/model/hold.ts
```diff
@@ -16,9 +16,9 @@
  */
 
 export interface Hold {
-    id?: string;
-    name?: string;
-    reason?: string;
+    id: string;
+    name: string;
+    reason: string;
```
MichalKinas, line 21: Should reason be required? I think it's not returned by API

lib/js-api/src/api/gs-core-rest-api/docs/Hold.md
```diff
@@ -4,19 +4,19 @@
 
 ```ts
 export interface Hold {
```
DenysVuika, line 6: You can remove the whole typescript block, putting the source of the interface in the docs does not make sense

lib/js-api/src/api/gs-core-rest-api/model/hold.ts
tamaragruszka: when we are creating new hold it is required, so I think it should be also in model

MichalKinas: As far as I can see `HoldBody` is used for creating holds so I don't think it should be required especially that api does not return it

g-jaskowski (approved):

lib/js-api/src/api/gs-core-rest-api/model/hold.ts
```diff
@@ -15,10 +15,8 @@
  * limitations under the License.
  */
 
-export interface Hold {
+import { HoldBody } from './holdBody';
+
+export interface Hold extends HoldBody {
```
MichalKinas, line 20: I'm not sure I like this since this approach makes `reason` here required but it will always be null when hold is fetched from api which feels weird

lib/js-api/src/api/gs-core-rest-api/model/hold.ts
```diff
@@ -15,10 +15,8 @@
...
     id: string;
-    name: string;
-    reason: string;
-    description?: string;
-    selected?: boolean;
```
MichalKinas, line 23: This field is not used anymore?

lib/js-api/src/api/gs-core-rest-api/docs/Hold.md
```diff
@@ -1,22 +1,10 @@
...
-|-----------------|------------|---------------|-----------------------------------|
-| **id**          | **string** |               | Hold id                           |
-| **name**        | **string** |               | Hold name                         |
-| **reason**      | **string** |               | Hold reason                       |
-| **description** | **string** |               | Additional information for a hold |
+| Name            | Type       | Default value | Description                                  |
+|-----------------|------------|---------------|----------------------------------------------|
+| **id**          | **string** |               | Hold id                                      |
+| **name**        | **string** |               | Hold name                                    |
+| **reason**      | **string** |               | Hold reason                                  |
```
MichalKinas, line 9: Should be optional as well

lib/js-api/src/api/gs-core-rest-api/model/hold.ts
MichalKinas: Okay sounds good

MichalKinas (approved):

nikita-web-ua (approved):

lib/js-api/src/api/gs-core-rest-api/model/hold.ts
DaryaBalvanovich: align to create separate interface for `getNodeAssignedHolds`. In other places `reason` is mandatory

lib/js-api/src/api/gs-core-rest-api/docs/Hold.md
DaryaBalvanovich: done

MichalKinas (approved):

lib/js-api/src/api/gs-core-rest-api/model/hold.ts
DaryaBalvanovich: discussed, split interfaces

PR #10112, opened 2024-08-21
[ACS-8602] [Legal Hold] The tabs names in "Manage Legal Holds" window are not matching the design

DaryaBalvanovich:

https://hyland.atlassian.net/browse/ACS-8602

the issue accrues after angular migration. because of `::before` space above title was too big.

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

title is centered

title aligned to the left

> - [ ] Yes
> - [ ] No

**Other information**:

### DIFF
--- a/lib/core/src/lib/dialogs/dialog/dialog.component.scss
+++ b/lib/core/src/lib/dialogs/dialog/dialog.component.scss
@@ -114,14 +114,22 @@ $dialog-padding: 24px;
     }
 
     .adf-dialog-header {
-        position: relative;
         display: flex;
-        justify-content: space-between;
         align-items: center;
         padding-bottom: $dialog-padding;
 
         &.adf-centered-header {
             align-items: flex-start;
+        }
+
+        .adf-dialog-title-container {
+            flex: 1;
+
+            &.adf-centered-title {
+                display: flex;
+                flex-direction: column;
+                align-items: center;
+            }
         }
 
         .adf-dialog-title {
@@ -168,10 +176,6 @@ $dialog-padding: 24px;
 
     .adf-dialog-header::after {
         bottom: 0;
-    }
-
-    #{$mat-dialog-title}::before {
-        display: none;
     }
 
     .adf-dialog-actions {
@@ -277,14 +281,10 @@ $dialog-padding: 24px;
     color: var(--theme-primary-color);
 }
 
-// TODO: Update after migration to Angular 15
-/* stylelint-disable-next-line selector-class-pattern */
-.mat-dialog-actions .mat-button-base + .mat-button-base {
+#{$mat-dialog-actions} #{$mat-button-base} + #{$mat-button-base} {
     margin-left: 16px;
 }
 
-// TODO: Update after migration to Angular 15
-/* stylelint-disable-next-line selector-class-pattern */
-.mat-dialog-container {
+#{$mat-dialog-container} #{$mat-dialog-surface} {
     border-radius: 8px;
 }
--- a/lib/core/src/lib/dialogs/dialog/dialog.component.scss
+++ b/lib/core/src/lib/dialogs/dialog/dialog.component.scss
@@ -1,3 +1,5 @@
+@import 'styles/mat-selectors';
+
 $dialog-large-l-width: 1075px;
 $dialog-large-l-height: 907px;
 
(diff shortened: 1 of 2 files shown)


g-jaskowski (approved):

lib/core/src/lib/dialogs/dialog/dialog.component.scss
```diff
@@ -168,6 +170,10 @@ $dialog-padding: 24px;
         bottom: 0;
     }
 
+    #{$mat-dialog-title}::before {
```
swapnil-verma-gl, line 173: I guess the long term goal is to get rid of mat-selectors completely along with the corresponding scss selector file. So instead of targeting mat-dialog-title, you can instead use the 'adf-dialog-header' class and align the text that way. Also, instead of using the ::before pseudo selector and setting its display property to none, a better approach would be to remove the justify-content property from the adf-dialog-header css, and give the adf-dialog-title a flex value of 1. So, all in all, the css would look something like 

```css
.adf-dialog-header {
        position: relative;
        display: flex;
        /* justify-content: space-between; <-Remove this */
        align-items: center;
        padding-bottom: $dialog-padding;

       ...

        .adf-dialog-title {
            font-size: large;
            font-weight: 200;
            margin: 0;
            flex: 1    /* <-Add this */
        }
        ...
}
```

lib/core/src/lib/dialogs/dialog/dialog.component.scss
```diff
@@ -1,3 +1,5 @@
+@import 'styles/mat-selectors';
```
DenysVuika, line 1: this should not be used, we are trying to get rid of this asap

DaryaBalvanovich: used to replace old selectors

lib/core/src/lib/dialogs/dialog/dialog.component.scss
DaryaBalvanovich: applied suggestion, thank you!

lib/core/src/lib/dialogs/dialog/dialog.component.scss
MichalKinas: @DenysVuika so did we already decide what to use instead?

DenysVuika: @MichalKinas just using proper container styling, in most of the cases there are ways to avoid touching mat components, in the case we definitely need that I suggest creating custom components that wrap that (i.e. we have button, progress already, can have adf-input if needed). But in this PR there's nothing from the workarounds needed

lib/core/src/lib/dialogs/dialog/dialog.component.scss
```diff
@@ -271,14 +281,10 @@ $dialog-padding: 24px;
     color: var(--theme-primary-color);
 }
 
-// TODO: Update after migration to Angular 15
-/* stylelint-disable-next-line selector-class-pattern */
-.mat-dialog-actions .mat-button-base + .mat-button-base {
+#{$mat-dialog-actions} #{$mat-button-base} + #{$mat-button-base} {
```
DenysVuika, line 284: this can be removed as well, you can have something like `.adf-dialog-actions`, just introduce a custom class for the actions container, and you can style children easy, i.e. 

```css
.adf-dialog-actions {
  display: flex;
  column-gap: 16px
}
```

for the dialog surface can be found something as well.
if you can't access the parent dialog class - the library allows passing a custom dialog panel class when invoking a dialog

DaryaBalvanovich: added, but `column-gap: 8px` as we still have margin `8px`

lib/core/src/lib/dialogs/dialog/dialog.component.scss
DaryaBalvanovich: removed `mat-selector`. Not sure that now it looks ideal because this overrides now looks more hacking and harder to understand:

- We use `column-gap` but we use `8px` to get `16px` in total with margin from button
- And we use `overflow: hidden` to apply **new** `border-radius` but actually `border-radius` defined in another place: `mdc-dialog__surface`

DenysVuika (approved):

PR #10142, opened 2024-08-28
ACS-8227: remove internal mime type icon pipe

DenysVuika:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

- improve Viewer code not to require extra pipes
- fix Viewer mime type alt (should be text for screen readers, not the value)
- remove the mime type pipe (internal one)

> - [ ] Yes
> - [x] No

**Other information**:
https://hyland.atlassian.net/browse/ACS-8227


lib/core/src/lib/viewer/components/viewer.component.html
```diff
@@ -43,8 +43,8 @@
                         <mat-icon>navigate_before</mat-icon>
                     </button>
                     <img class="adf-viewer__mimeicon"
-                         [alt]="originalMimeType || mimeType"
-                         [src]="(originalMimeType || mimeType) | adfMimeTypeIcon"
+                         [alt]="'ADF_VIEWER.ARIA.MIME_TYPE_ICON' | translate"
+                         [src]="mimeTypeIconUrl"
```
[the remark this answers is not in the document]
DenysVuika: Not fully true, the `adfMimeTypeIcon` was never implemented as a pure pipe, so was not giving any performance improvements compared to the getter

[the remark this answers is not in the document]
DenysVuika: @swapnil-verma-gl I reviewed the whole implementation and it's absolutely useless. There's a second property "originalMimeType" introduced, even propagated to the viewer... however the fix is a 1 liner in code, no need for all that code

```typescript
const nodeMimeType = nodeData?.content?.mimeType;
const renditionMimeType = nodeRendition.mimeType;
mimeType = nodeMimeType || renditionMimeType;
```

docs/core/components/viewer.component.md
swapnil-verma-gl, line 1: Can we also remove the entry for originalMimeType from [class properties](https://github.com/Alfresco/alfresco-ng2-components/blob/be65588bb874d7035a9592a59cc163debdf143c1/docs/core/components/viewer.component.md?plain=1#L77) section?

swapnil-verma-gl (dismissed):

lib/core/src/lib/viewer/components/viewer.component.ts
```diff
@@ -300,7 +301,7 @@ export class ViewerComponent<T> implements OnDestroy, OnInit, OnChanges {
     ) {}
 
     ngOnChanges(changes: SimpleChanges) {
-        const { blobFile, urlFile } = changes;
+        const { blobFile, urlFile, mimeType } = changes;
 
         if (blobFile?.currentValue) {
             this.mimeType = blobFile.currentValue.type;
```
swapnil-verma-gl, line 307: Looking at this code, should we also handle getting mimeTypeIcon from the mimeType property as well? Is there a possibility of a case where the mimeTypeIcon shows a different icon compared to the value of the mimeType property?

lib/core/src/lib/viewer/components/viewer.component.html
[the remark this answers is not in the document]
swapnil-verma-gl: Yes, this looks much cleaner now. We should also make sure to add a note about the removal of the originalMimeType property, for customers who might be this component in their own apps/extensions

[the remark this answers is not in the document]
DenysVuika: I think that was the internal solution we did in the last release, and that release was internal, so customers didn't have chances reusing it

lib/core/src/lib/viewer/components/viewer.component.ts
DenysVuika: this component has no `mimeTypeIcon` property... we have a calculated `mimeTypeIconUrl` that gets the path to SVG file based on the mimeType

swapnil-verma-gl: I meant the mimeType input property mentioned at line num [209](https://github.com/Alfresco/alfresco-ng2-components/pull/10142/files/be65588bb874d7035a9592a59cc163debdf143c1#diff-8edc4da1fb1bec4de886be767b73f594e2c448e23cf2858260ba7f60acce88b3L209). Based on the code in line num 306 (where this comment thread has started), in addition to being changed via inputProperty from the parent, the mimeType variable will also change if the blobFile input property changes. However, the proposed code of updating the mimeTypeIcon only takes into account changes being made to the mimeType input property by the parent (via changes.mimeType). So I am wondering, if there can be a case, where only the blobFile property changes. The code here would also update the mimeType variable, but the mimeTypeIcon variable would not get updated (since the mimeType input property was not explicitly changed by the parent component)

DenysVuika: ok, got it, updating the url also on the blobFile change

lib/core/src/lib/viewer/components/viewer.component.ts
```diff
@@ -300,15 +301,20 @@ export class ViewerComponent<T> implements OnDestroy, OnInit, OnChanges {
...
+            this.mimeTypeIconUrl = this.thumbnailService.getMimeTypeIcon(mimeType.currentValue);
```
swapnil-verma-gl, line 308: I guess it should use either blobFile.currentValue.type or this.mimeType, instead of mimeType, since we want to cover the case where only the blobFile property was changed. But other than that, everything looks good to me.

swapnil-verma-gl (approved):

lib/core/src/lib/viewer/components/viewer.component.ts
DenysVuika: yes, updated the code, thanks

```
this.mimeTypeIconUrl = this.thumbnailService.getMimeTypeIcon(blobFile.currentValue.type);
```

PR #10143, opened 2024-08-28
ACS-8221: remove internal localized role pipe

DenysVuika:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

- remove internal localize role pipe
- improve the user-role-column component implementation to translate the roles
- add missing unit tests

> - [ ] Yes
> - [x] No

**Other information**:
https://hyland.atlassian.net/browse/ACS-8221


lib/content-services/src/lib/permission-manager/components/user-role-column/user-role-column.component.ts
```diff
@@ -38,13 +38,13 @@ import { LocalizedRolePipe } from '@alfresco/adf-core';
                 (keyup.arrowup)="$event.stopPropagation()"
             >
                 <mat-option *ngFor="let role of roles" [value]="role.role">
-                    {{ role.label | adfLocalizedRole }}
+                    {{ i18nRoleValue(role.label) | translate }}
```
swapnil-verma-gl, line 41: Is there a way in which we don't have to call a method in the template? Maybe using ngIf? Having a method  call in template is causing the method to be called repeatedly for each change detection cycle, which might cause performance issues in the long run

DenysVuika: same as other pipes I am cleaning up, they are not "pure", so zero performance optimizations... this code is just getting rid of the useless pipe, but not does any performance work. I would suggest we review the pipes later in terms of the performance, as the scope of this PR is just to remove the pipe.

DenysVuika: For this PR, I've changed the way the options are localised, so should improve performance.

lib/content-services/src/lib/permission-manager/components/user-role-column/user-role-column.component.ts
```diff
@@ -80,10 +85,33 @@ export class UserRoleColumnComponent {
...
+        }
+        return value;
+    }
+
+    ngOnChanges(changes: SimpleChanges) {
+        if (changes.value) {
+            this.i18nValue = this.i18nRoleValue(changes.value.currentValue);
+        }
+
+        if (changes.roles) {
```
swapnil-verma-gl, line 112: Is the roles array something that will actively change while this component is 'alive' ? I'm asking because while the code will work without any issues, it does make it so that the roles input property is no longer directly accessed anywhere in this component, only indirectly accessed in ngOnChanges. 
If it won't change, then I guess shifting this code to ngOnInit/ngAfterViewInit or something similar should work too, and access the roles object directly
If it does change, then I guess we can leave it as proposed (although it does feel a bit weird to me to be honest)

lib/content-services/src/lib/permission-manager/components/user-role-column/user-role-column.component.ts
swapnil-verma-gl: Yes, this looks much better now. Thanks!!

lib/content-services/src/lib/permission-manager/components/user-role-column/user-role-column.component.ts
DenysVuika: my intention was to change the source, where the roles are coming from, but this "roles" input is referenced with a bunch of other components that have it coming from the http responses. So no clue how often it's going to change. NgOnChanges is a good place as it is fired only when "roles" are changed

swapnil-verma-gl (approved):

g-jaskowski (approved):

PR #10200, opened 2024-09-12
[ACS-8749] View More tags button no longer shows when all tags are removed from a node

swapnil-verma-gl:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

View more button was not getting hidden when all tags were removed from a node

View more button now gets hidden

> - [ ] Yes
> - [x] No

**Other information**:
https://hyland.atlassian.net/browse/ACS-8749


lib/core/src/lib/dynamic-chip-list/dynamic-chip-list.component.html
```diff
@@ -26,7 +26,7 @@
     <button
         data-automation-id="adf-dynamic-chip-list-view-more-button"
         mat-button
-        [hidden]="!limitChipsDisplayed"
+        [hidden]="chipsToDisplay.length === 0 || !limitChipsDisplayed"
```
MichalKinas, line 29: Maybe we could go with `< 2`? Since it still doesn't make too much sense to show it when there is only one tag

swapnil-verma-gl: The `chipsToDisplay` property only contains the chips that are to be currently displayed, not all chips that are present. By default, it will always have a length of at least 1, so that at least a single chip is shown. So having a check against `< 2` will cause problems as it will hide the button in case the container is so small that only a single chip can be shown. See below video  - 

https://github.com/user-attachments/assets/97b75adf-7f16-4885-99e7-f9967496ab70

g-jaskowski (approved):

swapnil-verma-gl:
Closing this PR in favour of #10221. This PR has got some rebase issue, and is showing more files than what were originally changed.

PR #10226, opened 2024-09-17
[ACS-8065][ADF] Unify components in TagModule

tamaragruszka:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://hyland.atlassian.net/browse/ACS-8065
https://hyland.atlassian.net/browse/ACS-5168 - labels

all components in tag module use dynamic chip component

> - [ ] Yes
> - [x] No

**Other information**:


lib/content-services/src/lib/tag/tags-creator/tags-creator.component.ts
```diff
@@ -15,23 +15,24 @@
...
 import { MatButtonModule } from '@angular/material/button';
+import { MatChipsModule } from '@angular/material/chips';
 import { MatIconModule } from '@angular/material/icon';
+import { MatInputModule } from '@angular/material/input';
 import { MatListModule } from '@angular/material/list';
 import { MatProgressSpinnerModule } from '@angular/material/progress-spinner';
+import { TranslateModule } from '@ngx-translate/core';
+import { EMPTY, forkJoin, Observable, Subject, timer } from 'rxjs';
+import { debounce, distinctUntilChanged, finalize, first, map, takeUntil, tap } from 'rxjs/operators';
+import { AdfStringsToChipsPipe } from '../../../../../core/src/lib/dynamic-chip-list/adf-strings-to-chips.pipe';
```
DenysVuika, line 32: incorrect import

lib/core/src/lib/dynamic-chip-list/adf-strings-to-chips.pipe.ts
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+import { Pipe, PipeTransform } from '@angular/core';
+import { Chip } from './chip';
+
+@Pipe({
+    name: 'adfStringsToChips',
+    standalone: true
+})
+export class AdfStringsToChipsPipe implements PipeTransform {
```
DenysVuika, line 25: we don't need this pipe to be in the core, please just use this code in the component instead

lib/core/src/lib/dynamic-chip-list/dynamic-chip-list.component.ts
```diff
@@ -118,6 +120,7 @@ export class DynamicChipListComponent implements OnChanges, OnInit, AfterViewIni
         if (changes.chips) {
             this.initialChips = this.chips;
             this.chipsToDisplay = this.initialChips;
+            this.changeDetectorRef.detectChanges();
```
DenysVuika, line 123: is this some sort of hack/workaround?

tamaragruszka: missed that, should be removed

lib/core/src/lib/dynamic-chip-list/adf-strings-to-chips.pipe.ts
tamaragruszka: transforming other way will need much more problematic changes, that was the cleanest way to do it

DenysVuika: you can do the same in the component, what's the problem moving this function there? this pipe does zero performance saving as it's not even marked as pure

tamaragruszka: Pipes are pure by default, so no need mark it manually. Also `map` function is returning brand new array.

In this scenario I am not changing any variable data type, so the change happens while passing data between components. There is already a lot of different variables for different operations.

The only way is to add input field for string array, which will need additional checks on every array change, which will cause additional calls on change.

lib/content-services/src/lib/content-metadata/components/content-metadata/content-metadata.component.scss
```diff
@@ -84,18 +84,8 @@ $panel-properties-height: 56px !default;
...
-                }
-
-                &.adf-creator-with-existing-tags-panel {
-                    background: var(--adf-theme-background-dialog-color);
-                }
-            }
-
-            [hidden] {
-                visibility: hidden;
+            #{$mat-chip} {
```
swapnil-verma-gl, line 87: Is it necessary to use this selector here? Can we have a class or id that we are adding directly in the html/ts? The plan is to eventually get rid of the mat-selectors.scss file, so this will have to be changed at some point.

tamaragruszka: yeah, it was possible :)

lib/content-services/src/lib/content-metadata/components/content-metadata/content-metadata.component.ts
```diff
@@ -218,6 +214,12 @@ export class ContentMetadataComponent implements OnChanges, OnInit, OnDestroy {
         return this._tags;
     }
 
+    get tagsToDisplay(): Chip[] {
```
AleksanderSklorz, line 217: Can we don’t do mapping in getter? Getter is called in html and we should avoid calling functions in html (so getter should not call map function). Instead this mapping should be done in place where tags are assigned.

lib/content-services/src/lib/tag/tags-creator/tags-creator.component.ts
```diff
@@ -91,8 +92,10 @@ export class TagsCreatorComponent implements OnInit, OnDestroy {
     @Input()
     set tags(tags: string[]) {
         this._tags = [...tags];
+        this.tagsToDisplay = this._tags.map((tag) => {
```
AleksanderSklorz, line 95: Can we use the getter for tags here? If some field has getter then we should refer to it by getter.

lib/content-services/src/lib/tag/tags-creator/tags-creator.component.ts
```diff
@@ -104,6 +107,8 @@ export class TagsCreatorComponent implements OnInit, OnDestroy {
         return this._tags;
     }
 
+    tagsToDisplay: Chip[] = [];
```
AleksanderSklorz, line 110: This field should be defined below inputs and outputs. You defined it between inputs and outputs.

lib/content-services/src/lib/content-metadata/components/content-metadata/content-metadata.component.ts
tamaragruszka: we could, but that would need changes in component to the scratch, the original property is passed to the others and that would need separate task

lib/content-services/src/lib/tag/tags-creator/tags-creator.component.ts
AleksanderSklorz: not addressed, I meant here something like: 
this.tags.map instead of this._tags.map, can you just remove this _ character as we have getter which we can use?

lib/core/src/lib/dynamic-chip-list/dynamic-chip-list.component.ts
```diff
@@ -119,10 +125,10 @@ export class DynamicChipListComponent implements OnChanges, OnInit, AfterViewIni
             this.initialChips = this.chips;
             this.chipsToDisplay = this.initialChips;
             if (this.limitChipsDisplayed && this.chipsToDisplay.length) {
-                setTimeout(() => {
-                    this.calculateChipsToDisplay();
-                    this.changeDetectorRef.detectChanges();
-                });
+                // setTimeout(() => {
```
AleksanderSklorz, line 128: commented code, please don't leave commented code

lib/core/src/lib/dynamic-chip-list/dynamic-chip-list.component.ts
AleksanderSklorz, line 1: can we have some unit tests for changes in that file?

lib/content-services/src/lib/tag/tags-creator/tags-creator.component.ts
AleksanderSklorz: ah sorry @tamaragruszka, I see you fixed this for that file but the same thing is in different file, here: 
image

Can you fix?

AleksanderSklorz: it's lib/content-services/src/lib/content-metadata/components/content-metadata/content-metadata.component.ts file

lib/content-services/src/lib/content-metadata/components/content-metadata/content-metadata.component.ts
```diff
@@ -214,6 +211,13 @@ export class ContentMetadataComponent implements OnChanges, OnInit, OnDestroy {
         return this._assignedTags;
     }
 
+    set tags(tags: string[]) {
+        this._tags = tags;
+        this.tagsToDisplay = this._tags.map((tag) => {
```
AleksanderSklorz, line 216: eslint issue: 
image

lib/content-services/src/lib/tag/tags-creator/tags-creator.component.ts
```diff
@@ -91,13 +91,13 @@ export class TagsCreatorComponent implements OnInit, OnDestroy {
     @Input()
     set tags(tags: string[]) {
         this._tags = [...tags];
+        this.tagsToDisplay = this.tags.map((tag) => {
```
AleksanderSklorz, line 94: eslint issue: 
image

tamaragruszka: I don't have any issue there

lib/content-services/src/lib/content-metadata/components/content-metadata/content-metadata.component.ts
tamaragruszka: I don't have any issue there

AleksanderSklorz: I think something wrong with your settings (for example IDE settings) because I can see that this rule is added in eslint so that eslint issue is valid and should be fixed. Note that it's warning, not error, so maybe you have some settings which skip warnings but it should be fixed as it's in our configuration for eslint
image

lib/content-services/src/lib/tag/tags-creator/tags-creator.component.ts
AleksanderSklorz: same here, please fix

lib/content-services/src/lib/content-metadata/components/content-metadata/content-metadata.component.ts
tamaragruszka: I wonder if that can be something about IDE version, because it wasn't straight change and adding there braces wasn't intuitive

AleksanderSklorz (approved):

MichalKinas (approved):

PR #10248, opened 2024-09-25
AAE-22724 Fix json widget display

DiogoABastos:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://hyland.atlassian.net/browse/AAE-22724

JSON widget is now displayed correctly both in form editor and preview views

> - [ ] Yes
> - [x] No

**Other information**:


lib/core/src/lib/form/components/widgets/json/json.widget.html
```diff
@@ -0,0 +1,8 @@
+<div class="adf-json-widget">
+    <div class="adf-json-widget-label">
```
DenysVuika, line 2: you can probably go without this div as it does not do much, just use <label> instead

lib/core/src/lib/form/components/widgets/json/json.widget.scss
```diff
@@ -0,0 +1,11 @@
+.adf {
+    &-json-widget {
+        width: 100%;
+
+        &-label {
+            .adf-label {
+                -webkit-font-smoothing: antialiased;
```
DenysVuika, line 7: I think we have postcss already for all kinds of scenarios, you don't need to target some webkit styles unless it's a bugfix

DiogoABastos: I was not getting the same label style as in other widgets and added that rule

lib/core/src/lib/form/components/widgets/json/json.widget.html
DiogoABastos: I have. followed the same approach as in the text-widget. I have tried removing the div that wraps around the label but it does break the component. We can remove this div but this would require further style rules to the label and button

DenysVuika (approved):

PR #10250, opened 2024-09-26
[ACS-8834] KR UI page on Auth token expire page get cancel popup

jacekpluta:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://hyland.atlassian.net/browse/ACS-8833

Issue has been fixed

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/core/src/lib/dialogs/unsaved-changes-dialog/unsaved-changes.guard.spec.ts
+++ b/lib/core/src/lib/dialogs/unsaved-changes-dialog/unsaved-changes.guard.spec.ts
@@ -20,15 +20,47 @@ import { MatDialog, MatDialogModule, MatDialogRef } from '@angular/material/dial
 import { Observable, Subject } from 'rxjs';
 import { UnsavedChangesDialogComponent } from './unsaved-changes-dialog.component';
 import { UnsavedChangesGuard } from './unsaved-changes.guard';
+import { AuthenticationService, AuthGuardService, NoopAuthModule } from '@alfresco/adf-core';
 
 describe('UnsavedChangesGuard', () => {
     let guard: UnsavedChangesGuard;
+    let authService: AuthenticationService;
+    let authGuardService: AuthGuardService;
+
+    const expectGuardToBe = (condition: boolean, done: DoneFn, checkUnsaved?: boolean) => {
+        (guard.canDeactivate() as Observable<boolean>).subscribe((allowed) => {
+            if (checkUnsaved) {
+                allowed = guard.unsaved;
+            }
+            condition ? expect(allowed).toBeTrue() : expect(allowed).toBeFalse();
+            done();
+        });
+    };
 
     beforeEach(() => {
         TestBed.configureTestingModule({
-            imports: [MatDialogModule]
+            imports: [MatDialogModule, NoopAuthModule],
+            providers: [
+                {
+                    provide: AuthenticationService,
+                    useValue: {
+                        isLoggedIn: () => true
+                    }
+                },
+                {
+                    provide: AuthGuardService,
+                    useValue: {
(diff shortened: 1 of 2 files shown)


lib/core/src/lib/dialogs/unsaved-changes-dialog/unsaved-changes.guard.spec.ts
```diff
@@ -17,20 +17,52 @@
...
+                { provide: RedirectAuthService, useValue: { onLogin: EMPTY, onTokenReceived: of() } }
+            ]
         });
         guard = TestBed.inject(UnsavedChangesGuard);
+        authService = TestBed.inject(AuthenticationService);
+        TestBed.inject(AuthGuardService);
+        appConfigService = TestBed.inject(AppConfigService);
     });
 
+    const expectGuardToBe = (condition: boolean, done: DoneFn, checkUnsaved?: boolean) => {
```
AleksanderSklorz, line 54: move this const before beforeEach

lib/core/src/lib/dialogs/unsaved-changes-dialog/unsaved-changes.guard.spec.ts
```diff
@@ -17,20 +17,52 @@
...
+        (guard.canDeactivate() as Observable<boolean>).subscribe((allowed) => {
+            if (checkUnsaved) {
```
AleksanderSklorz, line 56: we can avoid repeating condition with expect and calling done function by doing following: 
```ts
if (checkUnsaved) {
  allowed = guard.unsaved;
}
condition ? expect(allowed).toBeTrue() : expect(allowed).toBeFalse();
done();

File: lib/core/src/lib/dialogs/unsaved-changes-dialog/unsaved-changes.guard.spec.ts [hunk 3]
```diff
@@ -43,63 +75,70 @@ describe('UnsavedChangesGuard', () => {
...
 
-            (guard.canDeactivate() as Observable<boolean>).subscribe((result) => {
-                expect(result).toBeFalse();
-                done();
+                expectGuardToBe(true, done);
+                afterClosed$.next(true);
+            });
+
+            it('should return true when authGuardBaseService.withCredentials returns true', (done) => {
+                spyOn(appConfigService, 'get').and.returnValue(false);
```
AleksanderSklorz, line 137: We should avoid mocking inner dependencies. Instead, please mock out withCredentials property from AuthGuardService which is used by UnsavedChangesGuard. It would look something like: 
```ts
spyOnProperty(authGuardService, 'withCredentials').and.returnValue(false);
```

Please note I used spyOnProperty to mock value returned by withCredentials getter.

lib/core/src/lib/dialogs/unsaved-changes-dialog/unsaved-changes.guard.spec.ts
```diff
@@ -17,20 +17,52 @@
...
     let guard: UnsavedChangesGuard;
+    let authService: AuthenticationService;
+    let appConfigService: AppConfigService;
 
     beforeEach(() => {
         TestBed.configureTestingModule({
-            imports: [MatDialogModule]
+            imports: [MatDialogModule, HttpClientTestingModule],
+            providers: [
+                UrlHelperService,
```
AleksanderSklorz, line 37: I think you don't need these all providers so you can remove them all. Also you don't need HttpClientTestingModule. Instead you can just import NoopAuthModule in imports array and it should work as well as it's special module for mocked auth things

AleksanderSklorz (approved):

PR #10262, opened 2024-09-30
Release 7.0.0-alpha.3

MichalKinas:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation
> - [x] Other... Please describe: Release PR

> - [ ] Yes
> - [x] No

**Other information**:


g-jaskowski (approved):

nikita-web-ua (approved):

AleksanderSklorz (approved):

lib/core/package.json
```diff
@@ -33,8 +33,8 @@
     "@angular/material": ">=14.1.2",
     "@angular/router": ">=14.1.3",
     "@mat-datetimepicker/core": ">=10.1.1",
-    "@alfresco/js-api": ">=8.0.0-alpha.2-0",
-    "@alfresco/adf-extensions": ">=7.0.0-alpha.2-0",
+    "@alfresco/js-api": ">=8.0.0-alpha.3",
+    "@alfresco/adf-extensions": ">=7.0.0-alpha.3",
```
VitoAlbano, line 37: Shouldn't this keep having the -0 to catch also the prerelease versions generated after?

MichalKinas: Well that depends on the approach we want to take since after the release we can bump the version to e.g. `7.0.0-alpha.4` so that new alphas will be produced towards the new version, then we can update peer deps in this PR or if we don't want to bump it and keep producing `7.0.0-alpha.3-xxx` then I will update them in this PR, which approach do you think is better?

VitoAlbano (approved):

PR #10266, opened 2024-09-30
AAE-26155 Added feature for refreshing filter

tomaszhanaj:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://hyland.atlassian.net/browse/AAE-26155

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/process-services-cloud/src/lib/process/process-filters/components/process-filters-cloud.component.ts
+++ b/lib/process-services-cloud/src/lib/process/process-filters/components/process-filters-cloud.component.ts
@@ -43,14 +43,6 @@ export class ProcessFiltersCloudComponent implements OnInit, OnChanges, OnDestro
     @Input()
     showIcons: boolean = false;
 
-    /** (optional) The property indicates that the filter has been refreshed by an external action */
-    @Input()
-    set refreshedFilterKey(value: string[]) {
-        if (value?.length) {
-            this.updatedFiltersSet.delete(value[0]);
-        }
-    }
-
     /** Emitted when a filter is being selected based on the filterParam input. */
     @Output()
     filterSelected = new EventEmitter<ProcessFilterCloudModel>();
@@ -92,6 +84,7 @@ export class ProcessFiltersCloudComponent implements OnInit, OnChanges, OnDestro
             this.getFilters(this.appName);
         }
         this.initProcessNotification();
+        this.getFilterKeysAfterExternalRefreshing();
     }
 
     ngOnChanges(changes: SimpleChanges) {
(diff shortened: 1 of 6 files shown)


lib/process-services-cloud/src/lib/process/process-filters/components/process-filters-cloud.component.ts
```diff
@@ -43,6 +43,14 @@ export class ProcessFiltersCloudComponent implements OnInit, OnChanges, OnDestro
     @Input()
     showIcons: boolean = false;
 
+    /** (optional) Toggles showing an icon by the side of each filter */
+    @Input()
+    set refreshedFilterKey(value: string[]) {
+        if (value?.length) {
```
VitoAlbano, line 49: it's not clear to me what is happening here : We get as input an array. but there is always one element and we delete the set .
1 - I would avoid having operation on the set for an input attributes as this might lead to some hidden bugs. 
2 - If we want a value why don't we just accept a value?

lib/process-services-cloud/src/lib/task/task-filters/components/base-task-filters-cloud.component.ts
```diff
@@ -46,7 +46,7 @@ export abstract class BaseTaskFiltersCloudComponent implements OnDestroy {
     error: EventEmitter<any> = new EventEmitter<any>();
 
     counters$: { [key: string]: Observable<number> } = {};
-    updatedCounters: string[] = [];
+    updatedCountersSet = new Set<string>();
```
VitoAlbano, line 49: HAve you switched this to a Set to better handling adding/removing?

VitoAlbano (changes requested):

lib/process-services-cloud/src/lib/task/task-filters/components/task-filters-cloud.component.ts
```diff
@@ -32,6 +32,14 @@ import { TaskCloudEngineEvent } from '../../../models/engine-event-cloud.model';
     encapsulation: ViewEncapsulation.None
 })
 export class TaskFiltersCloudComponent extends BaseTaskFiltersCloudComponent implements OnInit, OnChanges {
+    /** (optional) Toggles showing an icon by the side of each filter */
+    @Input()
+    set refreshedFilterKey(value: string[]) {
```
DenysVuika, line 37: this api does not make much sense without explanation

lib/process-services-cloud/src/lib/process/process-filters/components/process-filters-cloud.component.ts
VitoAlbano: @tomaszhanaj This is exactly my point, why instead of passing an array we pass the plain value as it's the thing that we are interested in? we can filter on the parent component to have it as plain string and have angular do the job of triggering the change when the value ACTUAL changes if what i got is correct

PR #10270, opened 2024-10-03
[MNT-24628] Handle difference between config object and browser storage ticket

tiagosalvado10:

> - [X] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

Following the steps from [MNT-24628](https://hyland.atlassian.net/browse/MNT-24628) we end up having different tickets in the client config object and browser local storage - where the ticket from config object is causing a 401 error. If the ticket from browser storage is used the request is successful.

The function `AlfrescoApiClient.getAlfTicket(string)` is now checking if tickets are different, if so, returns the ticket from browser storage and emits an event to initialize the API client and auth objects with the new ticket. 

The second time the function is called, the config object will contain the correct ticket (i.e. equals to the storage one) and the request will work.

> - [ ] Yes
> - [X] No

**Other information**:

[MNT-24628]: https://hyland.atlassian.net/browse/MNT-24628?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tiagosalvado10:
Added unit tests.

lib/js-api/src/alfrescoApi.ts
```diff
@@ -240,6 +249,19 @@ export class AlfrescoApi implements Emitter, AlfrescoApiType {
         });
     }
 
+    ticketMismatchListeners() {
```
DenysVuika, line 252: missing null/undefined checks, I suggest checking the clients like `this.contentClient?.off(...)` and for other cases

tiagosalvado10: I've added null/undefined checks as requested.

DenysVuika (approved):

PR #10307, opened 2024-10-17
[ACS-8634] "Manage Searches" - a full page list of saved searches

dominikiwanekhyland:

> - [ ] Bugfix
> - [x] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://hyland.atlassian.net/browse/ACS-8634

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/content-services/src/lib/common/services/saved-searches.service.spec.ts
+++ b/lib/content-services/src/lib/common/services/saved-searches.service.spec.ts
@@ -15,7 +15,7 @@
  * limitations under the License.
  */
 
-import { fakeAsync, flush, TestBed } from '@angular/core/testing';
+import { TestBed } from '@angular/core/testing';
 import { AlfrescoApiService } from '../../services/alfresco-api.service';
 import { NodeEntry } from '@alfresco/js-api';
 import { SavedSearchesService } from './saved-searches.service';


lib/core/src/lib/styles/_mat-selectors.scss
```diff
@@ -125,6 +125,7 @@ $mat-button-touch-target: '.mat-mdc-button-touch-target';
 $mat-evolution-chip-text-label: '.mdc-evolution-chip__text-label';
 $cdk-overlay-pane: '.cdk-overlay-pane';
 $cdk-drag-preview: '.cdk-drag-preview';
+$cdk-drag-placeholder: '.cdk-drag-placeholder';
```
swapnil-verma-gl, line 128: I think the overall plan is to get rid of mat-selectors.scss file completely. So if you are depending on this selector somewhere in your design, it would be better to modify that code now to no longer use the selector, to avoid future effort

lib/core/src/lib/datatable/components/datatable/datatable.component.ts
```diff
@@ -1010,6 +1022,20 @@ export class DataTableComponent implements OnInit, AfterContentInit, OnChanges,
         return !drop.getSortedItems()[index].disabled;
     }
 
+    onDragDrop(droppedEvent: CdkDragDrop<any>): void {
```
swapnil-verma-gl, line 1025: Can you add a unit test for the drag functionality?

lib/core/src/lib/styles/_mat-selectors.scss
dominikiwanekhyland: and what would be the correct approach to use that selector? unfortunately, I can't replace it by ours, as it's internal cdk selector and it's impossible without it

swapnil-verma-gl: I replaced the occurrence of cdk-drag-placeholder with adf-datatable-row, and it seems to produce a similar layout to what is being produced right now. However, I would suggest to make sure that the styling does indeed match the required design

swapnil-verma-gl:
something I noticed while testing. When doing a new search, the loading bar is no longer visible. This does not seem to happen on the current ACADEV env. See video below - 
Uploading no-search-loading.mp4…

lib/core/src/lib/datatable/components/datatable/datatable.component.ts
MichalKinas: Done

lib/core/src/lib/styles/_mat-selectors.scss
MichalKinas: Removed the selector

PR #10374, opened 2024-11-08
[MNT-24698] Add aspect name to the title when there are duplicated aspect titles

tiagosalvado10:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

When there are aspects with the same title in a custom model, the UI metadata panels will also have the same title. This situation creates a conflict which prevents the panels from being expanded/collapsed.

When setting the title for each metadata panel, it is now verified if the aspect title is duplicated, if so the aspect name is concatenated to the title, i.e, title (name) - This not only resolves the conflict but also allows to better distinguish aspects with the same title.

> - [ ] Yes
> - [x] No

**Other information**:

Further information can be found in this [Jira comment](https://hyland.atlassian.net/browse/MNT-24698?focusedCommentId=2205471)

### DIFF
--- a/lib/content-services/src/lib/content-metadata/services/content-metadata.service.ts
+++ b/lib/content-services/src/lib/content-metadata/services/content-metadata.service.ts
@@ -91,7 +91,11 @@ export class ContentMetadataService {
             const title = propertyGroup.title;
             const name = propertyGroup.name;
             if (title) {
-                propertyGroup.title = propertyGroupsTitles[title] ? `${title} (${name})` : title;
+                if (propertyGroupsTitles[title]) {
+                    propertyGroup.title = name ? `${title} (${name})` : title;
+                } else {
+                    propertyGroup.title = title;
+                }
                 propertyGroupsTitles[title] = title;
             } else {
                 propertyGroup.title = name;
--- a/lib/content-services/src/lib/content-metadata/services/content-metadata.service.ts
+++ b/lib/content-services/src/lib/content-metadata/services/content-metadata.service.ts
@@ -86,19 +86,20 @@ export class ContentMetadataService {
     }
 
     setTitleToNameIfNotSet(propertyGroups: OrganisedPropertyGroup[]): OrganisedPropertyGroup[] {
-        const propertyGroupsTitles = {};
+        const propertyGroupsTitles = [];
         propertyGroups.map((propertyGroup) => {
             const title = propertyGroup.title;
             const name = propertyGroup.name;
             if (title) {
-                if (propertyGroupsTitles[title]) {
+                if (propertyGroupsTitles.includes(title)) {
                     propertyGroup.title = name ? `${title} (${name})` : title;
                 } else {
                     propertyGroup.title = title;
                 }
-                propertyGroupsTitles[title] = title;
+                propertyGroupsTitles.push(title);
             } else {
                 propertyGroup.title = name;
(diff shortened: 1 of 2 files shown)


lib/content-services/src/lib/content-metadata/services/content-metadata.service.ts
```diff
@@ -87,13 +86,25 @@ export class ContentMetadataService {
     }
 
     setTitleToNameIfNotSet(propertyGroups: OrganisedPropertyGroup[]): OrganisedPropertyGroup[] {
+        const propertyGroupsTitles = {};
```
swapnil-verma-gl, line 89: Instead of a JSON object containing data like 
```
propertyGroupTitles = {
   aspectTitle: 'aspectTitle',
   someOtherTitle: 'someOtherTitle'
}
```
we can instead have an array based collection, like
```
propertyGroupTitles = [ 'aspectTitle', 'someOtherTitle' ]
```
and then use `propertyGroupTitles.includes(title)` in the if check. That way, the entire logic would be much more readable on first glance

lib/content-services/src/lib/content-metadata/services/content-metadata.service.ts
```diff
@@ -87,13 +86,25 @@ export class ContentMetadataService {
...
+            const name = propertyGroup.name;
+            if (title) {
+                if (propertyGroupsTitles[title]) {
+                    propertyGroup.title = name ? `${title} (${name})` : title;
+                } else {
+                    propertyGroup.title = title;
+                }
+                propertyGroupsTitles[title] = title;
+            } else {
+                propertyGroup.title = name;
```
swapnil-verma-gl, line 101: I think we should also be keeping track of cases where the name property is used for a title. For e.g., there might be a situation where there are 2 aspects, where one does not have a title, rather a name field, with the same value as the title field of the other aspect. This may be defined as follows - 
```
aspect1Group = {
        name: 'Aspect:Custom',
        properties: [...]
};

aspect2Group = {
        name: 'test:aspect2',
        title: 'Aspect:Custom',
        properties: [...]
};
```
In this case, when creating the aspect panels, for the first aspect, the code will execute the outer most else case, and the name field will be used for the title (`Aspect:Custom`). For the second aspect, the code will execute the first if block. Within it, it will check if it is duplicating any of the current aspects, which it WONT find, since the data is only being tracked for the if block, not the else block. It will then simply use the title field for the header (`Aspect:Custom`), thereby duplicating the entries, and reproducing the issue

lib/content-services/src/lib/content-metadata/services/content-metadata.service.ts
tiagosalvado10: @swapnil-verma-gl thanks for pointing that out. I've applied the suggested change and added a new unit test to cover it.

lib/content-services/src/lib/content-metadata/services/content-metadata.service.ts
tiagosalvado10: @swapnil-verma-gl thanks for pointing that out. I've applied the suggested change and added a new unit test to cover it.

swapnil-verma-gl: Thanks, this looks much better now.

swapnil-verma-gl (approved):

PR #10375, opened 2024-11-08
[ACS-8921][ACA] Data Table / The input field in "Filter by Name" dialog is enormous in comparison with the UI design

dominikiwanekhyland:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://hyland.atlassian.net/browse/ACS-8921

Filter is fixed

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/content-services/src/lib/search/components/search-text/search-text.component.scss
+++ b/lib/content-services/src/lib/search/components/search-text/search-text.component.scss
@@ -10,37 +10,5 @@
             min-width: unset;
             border-radius: 50%;
         }
-
-        #{$mat-form-field-wrapper},
-        #{$mat-form-field-wrapper}#{$mat-text-field-focused} {
-            padding: 0;
-
-            #{$mat-notched-outline} {
-                #{$mat-notched-outline-trailing},
-                #{$mat-notched-outline-notch} {
-                    border-width: 0;
-                    border-top: 0 solid;
-                    border-left: 0 solid;
-                    border-right: 0 solid;
-                    border-bottom: 1px solid var(--adf-theme-foreground-secondary-text-color);
-
-                    #{$mat-floating-label} {
-                        left: 0;
-                    }
-                }
-            }
-        }
-    }
-
-    #{$mat-text-field-focused} {
-        #{$mat-notched-outline-trailing},
-        #{$mat-notched-outline-notch},
-        #{$mat-notched-outline-leading} {
-            border-bottom-width: 2px;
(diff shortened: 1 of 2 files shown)


lib/content-services/src/lib/search/components/search-text/search-text.component.scss
```diff
@@ -3,25 +3,38 @@
...
-        #{$mat-notched-outline-trailing},
-        #{$mat-notched-outline-notch} {
-            border-top: 0 solid;
-            border-left: 0 solid;
-            border-right: 0 solid;
-            border-bottom: 1px solid var(--adf-theme-foreground-secondary-text-color);
-
-            #{$mat-floating-label} {
-                left: 0;
+        #{$mat-form-field-wrapper},
```
swapnil-verma-gl, line 16: We  want to switch away from mat selectors. Since this is a component specific css file, is it not possible to use something like classes?

dominikiwanekhyland: it's inner element of angular material, so we can't add any custom class to it. It's mostly reformatting, as it was already added here

MichalKinas (changes requested):
There are some issues present still, for example when entering the value input is not aligned properly
Zrzut ekranu 2024-11-12 o 09 47 49

swapnil-verma-gl:
I just took a look at the JIRA, and there is a much simpler way of resolving this bug. All that is needed, is to remove the `appearance=outline` property from the input text field in HTML (located [here](https://github.com/Alfresco/alfresco-ng2-components/blob/ed8349e564f8392624a55b3ecc59615b70614f08/lib/content-services/src/lib/search/components/search-text/search-text.component.html#L3)). That changes the UI of the text field to something more consistent with the rest of the application. Additionally, it also allows us to remove the mat-selectors from the scss file as well. I tried on my local with the proposed change, and a completely empty scss file and it looked fine, except for some minor padding/spacing issues, which should not be too hard to fix.

On a side note, I remember this bug being fixed during the ng15 upgrade as well (see [ACS-8429](https://hyland.atlassian.net/browse/ACS-8429)). Don't know what caused it to re-appear, could be a miss during rebase. But regardless, we should probably make sure not to miss it during the ng17 upgrade

MichalKinas (approved):

g-jaskowski (approved):

PR #10386, opened 2024-11-12
AAE-10239 showRowContextMenu-datatable-action-called-multiple-times

mkrbr:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://hyland.atlassian.net/browse/AAE-10239
The showRowContextMenu event is emitted for every change on the component.

The showRowContextMenu event is emitted when the context menu on a cell is opened.

> - [ ] Yes
> - [x] No

adf-context-menu requires an action instead of an array of items.

**Other information**:


lib/content-services/src/lib/tree/components/tree.component.html
```diff
@@ -44,7 +44,7 @@
                 [attr.data-automation-id]="'node_' + node.id"
                 *matTreeNodeDef="let node"
                 matTreeNodePadding
-                [adf-context-menu]="contextMenuOptions"
+                [adf-context-menu]="getContextMenuOptions"
```
fkindgen, line 47: Do you tried `OnPush` change detection to avoid mulitple calls? https://hyland.atlassian.net/wiki/spaces/TECH/pages/1293321363/FE+code+best+practices#Use-OnPush-Change-Detection-Strategy

mkrbr: This would also work but as it was on component level I was not sure, if I could update it.

lib/core/src/lib/context-menu/context-menu.directive.ts
```diff
@@ -27,7 +27,7 @@ import { ContextMenuOverlayService } from './context-menu-overlay.service';
 export class ContextMenuDirective {
     /** Items for the menu. */
     @Input('adf-context-menu')
-    links: any[];
+    getActions: () => any[];
```
DudaRobert, line 30: Are you sure this directive is not exported? Otherwise, this would be an unnecessary breaking change.

DudaRobert (approved):

lib/core/src/lib/context-menu/context-menu.directive.ts
mkrbr: The directive is exported. What would be your suggestion? I needed to adjust the return type to make sure we set the actions when right before the context menu is opened, like it was requested in the ticket.

DudaRobert (approved):

fkindgen (approved):

PR #10392, opened 2024-11-15
AAE-27343 Listen to form rules changes from reactive widgets

tomgny:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://hyland.atlassian.net/browse/AAE-27343

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/core/src/lib/form/components/form-field/form-field.component.ts
+++ b/lib/core/src/lib/form/components/form-field/form-field.component.ts
@@ -20,6 +20,7 @@ import {
     Component,
     ComponentFactory,
     ComponentRef,
+    DestroyRef,
     inject,
     Input,
     NgModule,
@@ -33,6 +34,8 @@ import { FormRenderingService } from '../../services/form-rendering.service';
 import { WidgetVisibilityService } from '../../services/widget-visibility.service';
 import { FormFieldModel } from '../widgets/core/form-field.model';
 import { FieldStylePipe } from '../../pipes/field-style.pipe';
+import { FormFieldTypes } from '../widgets/core/form-field-types';
+import { takeUntilDestroyed } from '@angular/core/rxjs-interop';
 
 declare const adf: any;
 
@@ -62,6 +65,7 @@ export class FormFieldComponent implements OnInit, OnDestroy {
 
     private readonly formRenderingService = inject(FormRenderingService);
     private readonly visibilityService = inject(WidgetVisibilityService);
+    private readonly destroyRef = inject(DestroyRef);
     private readonly compiler = inject(Compiler);
 
     ngOnInit() {
@@ -88,12 +92,27 @@ export class FormFieldComponent implements OnInit, OnDestroy {
                     instance.fieldChanged.subscribe((field) => {
                         if (field && this.field.form) {
                             this.visibilityService.refreshVisibility(field.form);
-                            field.form.onFormFieldChanged(field);
+                            this.triggerFormFieldChanged(field);
                         }
                     });
+
+                    if (FormFieldTypes.isReactiveType(instance?.field?.type)) {
+                        this.updateReactiveFormControlOnFormRulesEvent(instance);
+                    }
                 }
             }
         }
+    }
+
+    private updateReactiveFormControlOnFormRulesEvent(instance: any): void {
+        instance?.formService.formRulesEvent.pipe(takeUntilDestroyed(this.destroyRef)).subscribe(() => {
+            instance?.updateReactiveFormControl();
+            this.triggerFormFieldChanged(instance.field);
+        });
+    }
+
+    private triggerFormFieldChanged(field: FormFieldModel): void {
+        field.form.onFormFieldChanged(field);
     }
 
     ngOnDestroy() {
--- a/lib/core/src/lib/form/components/widgets/date-time/date-time.widget.ts
+++ b/lib/core/src/lib/form/components/widgets/date-time/date-time.widget.ts
@@ -87,9 +87,9 @@ export class DateTimeWidgetComponent extends WidgetComponent implements OnInit {
 
     private updateFormControlState(): void {
         this.datetimeInputControl.setValidators(this.isRequired() ? [Validators.required] : []);
-        if (this.field?.readOnly || this.readOnly) {
-            this.datetimeInputControl.disable({ emitEvent: false });
-        }
+        this.field?.readOnly || this.readOnly
+            ? this.datetimeInputControl.disable({ emitEvent: false })
+            : this.datetimeInputControl.enable({ emitEvent: false });
 
         this.datetimeInputControl.updateValueAndValidity({ emitEvent: false });
     }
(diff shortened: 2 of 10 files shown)


lib/core/src/lib/form/components/form-field/form-field.component.ts
```diff
@@ -88,14 +92,29 @@ export class FormFieldComponent implements OnInit, OnDestroy {
...
+                        this.updateReactiveFormControlOnFormRulesEvent(instance);
+                    }
                 }
             }
         }
     }
 
+    private updateReactiveFormControlOnFormRulesEvent(instance: any): void {
+        instance?.formService.formRulesEvent.pipe(takeUntilDestroyed(this.destroyRef)).subscribe(() => {
+            instance?.updateReactiveFormControl();
```
BSekula, line 109: Can we add an interface to `instance`? what I mean is adding something like: 

```
interface ReactiveFormWidget {
   updateReactiveFormControl: (): void;
   formService: Type
}
```

and implement it in reactive widgets

```
export class DateTimeWidgetComponent extends WidgetComponent implements OnInit, ReactiveFormWidget {..
```

So `updateReactiveFormControl` could be easily traced?

tomgny: good point, will add it

lib/core/src/lib/form/components/widgets/date/date.widget.ts
```diff
@@ -81,8 +80,17 @@ export class DateWidgetComponent extends WidgetComponent implements OnInit, OnDe
...
+        this.updateFormControlState();
+        this.validateField();
+    }
+
+    private setFormControlValue(): void {
         this.dateInputControl.setValue(this.field.value, { emitEvent: false });
+    }
+
+    private updateFormControlState(): void {
         this.dateInputControl.setValidators(this.isRequired() ? [Validators.required] : []);
```
BSekula, line 94: Can we use custom validators in date widget? 🤔

tomgny: as default option we use only required from default validators. for other cases, like min/max constraints, we handle it in the handleErrors method, and its needed to be compatible with existing form structure + other non-reactive form widgets.

lib/core/src/lib/form/components/form-field/form-field.component.ts
```diff
@@ -88,14 +92,29 @@ export class FormFieldComponent implements OnInit, OnDestroy {
                     instance.fieldChanged.subscribe((field) => {
                         if (field && this.field.form) {
                             this.visibilityService.refreshVisibility(field.form);
-                            field.form.onFormFieldChanged(field);
+                            this.triggerFormFieldChanged(field);
```
VitoAlbano, line 95: is it worth to have a method just for a single line of code?

lib/core/src/lib/form/components/form-field/form-field.component.ts
```diff
@@ -88,14 +92,29 @@ export class FormFieldComponent implements OnInit, OnDestroy {
...
                         }
                     });
+
+                    if (FormFieldTypes.isReactiveType(instance?.field?.type)) {
```
VitoAlbano, line 99: I was expecting this to be called when something has changed into the form field so inside the fieldChanged stream

lib/core/src/lib/form/components/form-field/form-field.component.ts
```diff
@@ -88,14 +92,29 @@ export class FormFieldComponent implements OnInit, OnDestroy {
...
+                        this.updateReactiveFormControlOnFormRulesEvent(instance);
+                    }
                 }
             }
         }
     }
 
+    private updateReactiveFormControlOnFormRulesEvent(instance: any): void {
+        instance?.formService.formRulesEvent.pipe(takeUntilDestroyed(this.destroyRef)).subscribe(() => {
```
VitoAlbano, line 108: Reducing the check to only the form rules might miss some other parts where we update the form and the field changed event is triggered

VitoAlbano (changes requested):

lib/core/src/lib/form/components/form-field/form-field.component.ts
tomgny: this method now is used in two places, so  thats why i move it to separate method, but i can agree that maybe its not worth

lib/core/src/lib/form/components/form-field/form-field.component.ts
tomgny: hmmm why? we subscribe to the formRulesEvent when we create this component instance, why do we need to put it under another subscription (nested)? it should listen for these events for the whole lifecycle of the instance

lib/core/src/lib/form/components/form-field/form-field.component.ts
tomgny: on each onFieldChanged call, inside the [base widget.component.ts ](https://github.com/Alfresco/alfresco-ng2-components/blob/develop/lib/core/src/lib/form/components/widgets/widget.component.ts#L106), and also on the all events, we update the value of formRulesEvent subject so it will be always triggered for all the events (and its needed for all events, because in modeling, we can set rules for all available event types, like blur, click etc). i think it should be enough, but maybe im not aware of some corner case

lib/core/src/lib/form/components/form-field/form-field.component.ts
VitoAlbano: If you add an interface (or an abstract class like WidgetComponent) please double check with a custom form widget if it's all good

lib/core/src/lib/form/components/form-field/form-field.component.ts
tomgny: if we do not subscribe for formRules changes, we wont be able to update the state of particular instance until the instance.fieldChanged will be triggered (so for example after blur, or input something event).
to make sure I understand you, this is what you mean, right?

```suggestion
                if (componentType) {
                    this.componentRef = this.container.createComponent(componentType);
                    const instance = this.componentRef.instance;
                    instance.field = this.field;
                    instance.fieldChanged.subscribe((field) => {
                        if (field && this.field.form) {
                            this.visibilityService.refreshVisibility(field.form);
                            this.triggerFormFieldChanged(field);

                            if (FormFieldTypes.isReactiveType(instance?.field?.type)) {
                                instance?.updateReactiveFormControl();
                                this.triggerFormFieldChanged(instance.field);
                            }
                        }
                    });
```
(i tested it and it doesnt work)

lib/core/src/lib/form/components/widgets/date-time/date-time.widget.ts
```diff
@@ -71,8 +70,16 @@ export class DateTimeWidgetComponent extends WidgetComponent implements OnInit,
...
+    updateReactiveFormControl(): void {
+        this.updateFormControlState();
+        this.validateField();
+    }
+
+    private setFormControlValue(): void {
         this.datetimeInputControl.setValue(this.field.value, { emitEvent: false });
+    }
+
+    private updateFormControlState(): void {
```
wiktord2000, line 82:
```suggestion
    private updateFormControlValidators(): void {
```
WDYT?
To me state is really broad term

VitoAlbano (approved):

lib/core/src/lib/form/components/widgets/date-time/date-time.widget.ts
tomgny: i think state in this case better describe what this method does, because it updates also readonly/disabled state (required is also state of the field)

PR #10413, opened 2024-11-20
Release 7.0.0-alpha.5 package-lock.json

swapnil-verma-gl:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation
> - [x] Other... Please describe: Release PR - Pushing missed package-lock.json

> - [ ] Yes
> - [ ] No

**Other information**:


lib/cli/package-lock.json
```diff
@@ -9,7 +9,7 @@
       "version": "7.0.0-alpha.5",
       "license": "Apache-2.0",
       "dependencies": {
-        "@alfresco/js-api": ">=7.8.0",
```
MichalKinas, line 12: Let's rollback the change for the cli, here js-api should have older version as we had some issues with that in the past

package-lock.json
```diff
@@ -366,229 +366,6 @@
         }
       }
     },
-    "node_modules/@angular-devkit/build-angular/node_modules/@esbuild/android-arm": {
```
MichalKinas, line 369: Not sure why do you have any changes here but I assume it depends on operating system so I would roll those change back too

MichalKinas (changes requested):

lib/cli/package-lock.json
swapnil-verma-gl: So should I revert the whole file to the original version? or just this mentioned version?

package-lock.json
swapnil-verma-gl: If we roll back the entire file, we would have the same package-lock.json as before, which causes the release job to fail

lib/cli/package-lock.json
MichalKinas: Whole file please

package-lock.json
MichalKinas: I doubt this part was responsible as the issue was with `Original error: Pruned lock file creation failed. The following package was not found in the root lock file: @alfresco/js-api@>=8.0.0-alpha.5` and this version was not changed across this file

MichalKinas:
Okay I found the real issue - `"@alfresco/js-api": ">=8.0.0-alpha.5",` this line in `package.json` in `cli`

MichalKinas (approved):

PR #10418, opened 2024-11-20
[AAE-22900] Update ObjectDataTableAdapter to support server and clien…

amolodyh-hyland:
Fixes [AAE-22900](https://hyland.atlassian.net/browse/AAE-22900?atlOrigin=eyJpIjoiYmU2M2FhMzFmMjFmNGVjNDkxOGRmMmNhOGI2MGE5Y2YiLCJwIjoiaiJ9). This adds server sorting for the base `ObjectDataTableAdapter` to enable tables in tabs such as "Process Admin" User Tasks, Services Tasks, and Applications List.

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

 [AAE-22900](https://hyland.atlassian.net/browse/AAE-22900?atlOrigin=eyJpIjoiYmU2M2FhMzFmMjFmNGVjNDkxOGRmMmNhOGI2MGE5Y2YiLCJwIjoiaiJ9)

Allows for enabling server sorting where it the adapter does not sort the data. The server data is rendered in the order it comes in.

> - [ ] Yes
> - [x] No

**Other information**:

This a partial fix to the provided story  [AAE-22900](https://hyland.atlassian.net/browse/AAE-22900?atlOrigin=eyJpIjoiYmU2M2FhMzFmMjFmNGVjNDkxOGRmMmNhOGI2MGE5Y2YiLCJwIjoiaiJ9). After this is merged, hxp-frontend-apps will need to be updated by updating the `applications-list-filtered-data-adapter` adapter as well.

[AAE-22900]: https://hyland.atlassian.net/browse/AAE-22900?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[AAE-22900]: https://hyland.atlassian.net/browse/AAE-22900?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/core/src/lib/datatable/data/object-datatable-adapter.ts
+++ b/lib/core/src/lib/datatable/data/object-datatable-adapter.ts
@@ -23,11 +23,14 @@ import { DataSorting } from './data-sorting.model';
 import { DataTableAdapter } from './datatable-adapter';
 import { Subject } from 'rxjs';
 
+export type SortingMode = 'client' | 'server';
+
 // Simple implementation of the DataTableAdapter interface.
 export class ObjectDataTableAdapter implements DataTableAdapter {
     private _sorting: DataSorting;
     private _rows: DataRow[];
     private _columns: DataColumn[];
+    private readonly _sortingMode: SortingMode;
 
     selectedRow: DataRow;
     rowsChanged: Subject<Array<DataRow>>;
@@ -54,9 +57,10 @@ export class ObjectDataTableAdapter implements DataTableAdapter {
         return schema;
     }
 
-    constructor(data: any[] = [], schema: DataColumn[] = []) {
+    constructor(data: any[] = [], schema: DataColumn[] = [], sortingMode: SortingMode = 'client') {
         this._rows = [];
         this._columns = [];
+        this._sortingMode = sortingMode?.toString().toLowerCase() === 'server' ? 'server' : 'client';
 
         if (data && data.length > 0) {
             this._rows = data.map((item) => new ObjectDataRow(item));
@@ -118,6 +122,10 @@ export class ObjectDataTableAdapter implements DataTableAdapter {
 
     setSorting(sorting: DataSorting): void {
         this._sorting = sorting;
+
+        if (this._sortingMode === 'server') {
+            return;
+        }
 
         if (sorting?.key) {
             this._rows.sort((a: DataRow, b: DataRow) => {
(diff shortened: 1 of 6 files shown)


lib/process-services-cloud/src/lib/process/process-list/datatable/process-list-datatable-adapter.ts
```diff
@@ -21,7 +21,7 @@ import { ProcessInstanceCloudListViewModel } from '../models/perocess-instance-c
 
 export class ProcessListDatatableAdapter extends ObjectDataTableAdapter {
     constructor(data: ProcessInstanceCloudListViewModel[], schema: DataColumn<ProcessListDataColumnCustomData>[]) {
```
amolodyh-hyland, line 23: @DenysVuika @eromano Do we want to provide the ability to set the `sortingMode` in these extending classes so we can utilize feature flags in the hxp-frontend-apps projects? Not sure if we want a feature flag for this but if we did, we'd need to be able to provide the sort option here.

Based on the information I've received, we don't want pages using these adapters to use client sorting so the constructor parameter would be redundant. Also, we would have to remove the constructor parameter once the feature flag is removed so it seems architecturally redundant.

lib/core/src/lib/datatable/data/object-datatable-adapter.ts
```diff
@@ -54,9 +57,10 @@ export class ObjectDataTableAdapter implements DataTableAdapter {
         return schema;
     }
 
-    constructor(data: any[] = [], schema: DataColumn[] = []) {
+    constructor(data: any[] = [], schema: DataColumn[] = [], sortingMode: SortingMode = 'client') {
         this._rows = [];
         this._columns = [];
+        this._sortingMode = sortingMode?.toString().toLowerCase() === 'server' ? 'server' : 'client';
```
sdonnell-hyland, line 63: nit: toString() probably isn't necessary

amolodyh-hyland: Just trying to be super careful incase `sortingMode` is not provided as a string. You think that's overkill?

sdonnell-hyland: It doesn't really hurt anything, but I would expect all consumers of ObjectDataTableAdapter to be using TypeScript typings, and not need the additional javascript safety checks.  I spot checked some files and didn't see anywhere else where typings weren't trusted, but I could have easily missed another file that does.

Personally, I would just account for sortingMode being undefined, but if it is defined assume its a SortingMode.

sdonnell-hyland (approved):
Looks good to me

jcain82 (approved):

lib/core/src/lib/datatable/data/object-datatable-adapter.ts
amolodyh-hyland: Sounds good, makes sense to me. I'll update it.

PR #10419, opened 2024-11-21
GH auto: Automated Update of Translations from Crowdin

alfresco-build:


VitoAlbano:
Something weird happened on this PR as some of the values were removed instead of added

atchertchian:
As discussed with @RCorsaletti it seems to me that Crowdin changed their export behavior: now untranslated (or unapproved) 
translations are empty instead of being filled with the source English value (cc @ehsan-2019)

The previous changes did empty all values for en.json files, the trick used to let translators change EN reference content was to add an extra English target language on Crowdin side.
I have removed this target language, and this unwanted change is not present anymore.

My understanding is that the FE code supports fallback on en reference file when translation is empty, so I assume the change in export is not a problem.
I have asked crowdin support to confirm the change in behavior.

cc @VitoAlbano

atchertchian:
After all this was linked to a change in export settings on crowdin project side, these changes have been reverted

atchertchian:
The configuration changes have been reverted but the PR does not get updated because there are no changes detected anymore -> closing

PR #10434, opened 2024-11-27
[ACS-8898] Empty context menu on multi-selected libraries

dominikiwanekhyland:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://hyland.atlassian.net/browse/ACS-8898

> - [ ] Yes
> - [x] No

**Other information**:


AleksanderSklorz (approved):

lib/content-services/src/lib/search/services/base-query-builder.service.ts
```diff
@@ -596,6 +596,7 @@ export abstract class BaseQueryBuilderService {
      * @param searchUrl search url to navigate to
      */
     async navigateToSearch(query: string, searchUrl: string) {
+        this.update();
```
MichalKinas, line 599: This line was previously here but I removed it as it was introducing another bug, with it the search execution will be called twice causing some errors in various situations, for example if you search for 'test' you will get the results but then after searching for `zxcv` even though you should not get any results you will get the 'old' search results from previous query, that's why I think we need to find some other solution for this bug

MichalKinas (changes requested):

lib/content-services/src/lib/search/services/search-facet-filters.service.ts
```diff
@@ -68,8 +68,6 @@ export class SearchFacetFiltersService {
             this.responseFacets = null;
         });
 
-        this.queryBuilder.updated.pipe(takeUntilDestroyed()).subscribe((query) => this.queryBuilder.execute(true, query));
```
MichalKinas, line 71: With this change when you change the state of any filter and click apply it won't trigger any search request

MichalKinas (changes requested):

dominikiwanekhyland:
https://github.com/Alfresco/alfresco-content-app/pull/4263

requrires ACA changes

MichalKinas (approved):

PR #10455, opened 2024-12-03
[ACS-8918] search date filter changing between created and modified tabs slides the text inside the tabs too

AleksanderSklorz:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://hyland.atlassian.net/browse/ACS-8918

Text is not sliding when tabs are switched for date filter. 

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/content-services/src/lib/search/components/search-filter-tabbed/search-filter-tabbed.component.ts
+++ b/lib/content-services/src/lib/search/components/search-filter-tabbed/search-filter-tabbed.component.ts
@@ -36,9 +36,9 @@ export class SearchFilterTabbedComponent implements OnInit, OnDestroy {
     selectedIndex: number = 0;
 
     @ViewChild(MatTabGroup)
-    private tabGroup: MatTabGroup;
+    private readonly tabGroup: MatTabGroup;
 
-    private intersectionObserver = new IntersectionObserver(
+    private readonly intersectionObserver = new IntersectionObserver(
         (entries) => {
             if (!entries[0].isIntersecting) {
                 this.tabGroup.selectedIndex = (this.selectedIndex + 1) % this.tabsContents.length;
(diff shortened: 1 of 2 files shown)


nikita-web-ua (approved):
can you resolve the sonar cloud issues please?

AleksanderSklorz:
> can you resolve the sonar cloud issues please?

done

swapnil-verma-gl (approved):

lib/content-services/src/lib/search/components/search-filter-tabbed/search-filter-tabbed.component.ts
```diff
@@ -29,12 +29,39 @@ import { TranslateModule } from '@ngx-translate/core';
...
 
     selectedIndex: number = 0;
 
+    @ViewChild(MatTabGroup)
+    private readonly tabGroup: MatTabGroup;
+
+    private readonly intersectionObserver = new IntersectionObserver(
+        (entries) => {
+            if (!entries[0].isIntersecting) {
+                this.tabGroup.selectedIndex = (this.selectedIndex + 1) % this.tabsContents.length;
```
MichalKinas, line 44: Why do we need to set `this.tabGroup.selectedIndex` twice?

AleksanderSklorz: This is some angular material problem when we use tabs together with some kind of overlaps like on below screenshot. Basically the problem is following: 
When we close this popup then seems like Tabs component becomes broken because after reopening popup Tabs component doesn't know on which tab it was before closing so when we try to switch tab we can see for a moment strange text sliding effect. 
To resolve that problem I wrote code which observes for hiding/showing component and I do single switching of tabs so Tabs component can restore it's state. It seems like only solution for that problem, because we can't just "destroy" tabs component and recreate it because we need to keep previous state. 

image

PR #10560, opened 2025-01-17
[ACS-8824] Disable node properties save button if values are invalid

nikita-web-ua:

  > - [X] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://hyland.atlassian.net/browse/ACS-8824
- Some forbidden symbols ignored by validator 
- Save button is active when form has errors 

- Disable save button if property validation fails 
- Add missing forbidden symbols to validator

> - [ ] Yes
> - [X] No

**Other information**:

### DIFF
--- a/lib/content-services/src/lib/content-metadata/components/content-metadata/content-metadata.component.spec.ts
+++ b/lib/content-services/src/lib/content-metadata/components/content-metadata/content-metadata.component.spec.ts
@@ -306,27 +306,38 @@ describe('ContentMetadataComponent', () => {
             });
 
             it('should disable the save button if metadata has not changed', () => {
-                component.hasMetadataChanged = false;
                 fixture.detectChanges();
                 toggleEditModeForGroup();
                 expect(getGroupSaveButton().disabled).toBeTrue();
             });
 
-            it('should disable the save button if there are invalid properties', () => {
-                component.hasMetadataChanged = true;
-                component.invalidProperties = ['invalidPropertyKey'];
+            it('should disable the save button if there are invalid properties', fakeAsync(() => {
+                updateService.update(
+                    {
+                        key: 'properties.property-key',
+                        isValidValue: false
+                    } as CardViewBaseItemModel,
+                    'updated-value'
+                );
+                tick(500);
                 fixture.detectChanges();
                 toggleEditModeForGroup();
                 expect(getGroupSaveButton().disabled).toBeTrue();
-            });
+            }));
 
-            it('should enable the save button if metadata has changed and there are no invalid properties', () => {
-                component.hasMetadataChanged = true;
-                component.invalidProperties = [];
+            it('should enable the save button if metadata has changed and there are no invalid properties', fakeAsync(() => {
+                updateService.update(
+                    {
+                        key: 'properties.property-key',
+                        isValidValue: true
+                    } as CardViewBaseItemModel,
+                    'updated-value'
+                );
+                tick(500);
                 fixture.detectChanges();
                 toggleEditModeForGroup();
                 expect(getGroupSaveButton().disabled).toBeFalse();
-            });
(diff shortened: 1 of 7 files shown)


swapnil-verma-gl (approved):

lib/content-services/src/lib/content-metadata/components/content-metadata/content-metadata.component.ts
```diff
@@ -549,4 +551,17 @@ export class ContentMetadataComponent implements OnChanges, OnInit {
         }
         return observables;
     }
+
+    private updateInvalidProperties() {
+        if (this.targetProperty?.isValidValue === false) {
```
AleksanderSklorz, line 556: Do we really need to check boolean values like === false and === true? Can't it be like for example: 
if (this.targetProperty?.isValidValue)
?

lib/content-services/src/lib/content-metadata/components/content-metadata/content-metadata.component.ts
```diff
@@ -164,6 +164,7 @@ export class ContentMetadataComponent implements OnChanges, OnInit {
     categoriesManagementMode = CategoriesManagementMode.ASSIGN;
     classifiableChanged = this.classifiableChangedSubject.asObservable();
     editing = false;
+    invalidProperties: string[] = [];
```
AleksanderSklorz, line 167: Looking at updateInvalidProperties function you don't want to add duplicates to array if element already exists in array. Maybe better choice will be to use Set type instead of Array? It will simplify your code in couple ways, for example: 
- you don't need to manually check if array already has element - set checks it automatically and doesn't allow to add duplicates 
- you can remove element from Set based on element (no need to look for index first)

lib/content-services/src/lib/content-metadata/components/content-metadata/content-metadata.component.spec.ts
```diff
@@ -288,6 +289,115 @@ describe('ContentMetadataComponent', () => {
...
+            it('should disable the save button if metadata has not changed', () => {
+                component.hasMetadataChanged = false;
+                fixture.detectChanges();
+                toggleEditModeForGroup();
+                expect(getGroupSaveButton().disabled).toBeTrue();
+            });
+
+            it('should disable the save button if there are invalid properties', () => {
+                component.hasMetadataChanged = true;
+                component.invalidProperties = ['invalidPropertyKey'];
```
AleksanderSklorz, line 317: I won't add this comment to every place but please apply it for each test for invalidProperties. Generally invalidProperties is normal property, it's not Input property. So you shouldn't test it by just setting this field because normally that field won't be set from parent. Instead it will be set inside ngOnInit code. So to test that you should to run change detector or ngOnInit function to set that property. 
Please also change the same for hasMetadataChanged property because I see it's not Input as well.

lib/content-services/src/lib/content-metadata/components/content-metadata/content-metadata.component.ts
nikita-web-ua: Good idea, implemented. Thank you

lib/content-services/src/lib/content-metadata/components/content-metadata/content-metadata.component.spec.ts
nikita-web-ua: done

lib/content-services/src/lib/content-metadata/components/content-metadata/content-metadata.component.ts
nikita-web-ua: I added this statement on purpose as an extra precaution step

AleksanderSklorz (approved):

PR #10571, opened 2025-01-23
[ACS-8582] Introduce UnitTestingUtils class and transform existing tests in core library

MichalKinas:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://hyland.atlassian.net/browse/ACS-8582

Shared unit testing utils class has been introduced to speed up the process of writing new unit tests cases.

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/core/feature-flags/src/lib/directives/features.directive.spec.ts
+++ b/lib/core/feature-flags/src/lib/directives/features.directive.spec.ts
@@ -20,7 +20,7 @@ import { Component } from '@angular/core';
 import { ComponentFixture, TestBed } from '@angular/core/testing';
 import { provideMockFeatureFlags } from '../mocks/features-service-mock.factory';
 import { FeaturesDirective } from './features.directive';
-import { UnitTestingUtils } from '@alfresco/adf-core';
+import { UnitTestingUtils } from '../../../../src/lib/testing/unit-testing-utils';
 
 @Component({
     template: `
@@ -46,6 +46,7 @@ class TestWithDisabledFlagComponent {
 describe('FeaturesDirective', () => {
     let enabledFixture: ComponentFixture<TestWithEnabledFlagComponent>;
     let disabledFixture: ComponentFixture<TestWithDisabledFlagComponent>;
+    let testingUtils: UnitTestingUtils;
 
     beforeEach(async () => {
         TestBed.configureTestingModule({
@@ -66,17 +67,20 @@ describe('FeaturesDirective', () => {
         disabledFixture = TestBed.createComponent(TestWithDisabledFlagComponent);
         disabledFixture.detectChanges();
 
+        testingUtils = new UnitTestingUtils(enabledFixture.debugElement);
+
         await enabledFixture.whenStable();
         await disabledFixture.whenStable();
     });
 
     it('should render the element with enabled features', () => {
-        const enabledFixtureElement = UnitTestingUtils.getByCSS(enabledFixture.debugElement, '#underFeatureFlag');
+        const enabledFixtureElement = testingUtils.getByCSS('#underFeatureFlag');
         expect(enabledFixtureElement).toBeDefined();
         expect(enabledFixtureElement.nativeElement).toBeDefined();
     });
 
     it('should not render the element with disabled features', () => {
-        expect(UnitTestingUtils.getByCSS(disabledFixture.debugElement, '#underFeatureFlag')).toBeNull();
+        testingUtils.setDebugElement(disabledFixture.debugElement);
+        expect(testingUtils.getByCSS('#underFeatureFlag')).toBeNull();
     });
 });
--- a/lib/core/feature-flags/src/lib/directives/features.directive.spec.ts
+++ b/lib/core/feature-flags/src/lib/directives/features.directive.spec.ts
@@ -20,7 +20,7 @@ import { Component } from '@angular/core';
 import { ComponentFixture, TestBed } from '@angular/core/testing';
 import { provideMockFeatureFlags } from '../mocks/features-service-mock.factory';
 import { FeaturesDirective } from './features.directive';
-import { By } from '@angular/platform-browser';
+import { UnitTestingUtils } from '@alfresco/adf-core';
 
 @Component({
     template: `
@@ -71,11 +71,12 @@ describe('FeaturesDirective', () => {
     });
 
     it('should render the element with enabled features', () => {
-        expect(enabledFixture.debugElement.query(By.css('#underFeatureFlag'))).toBeDefined();
-        expect(enabledFixture.debugElement.query(By.css('#underFeatureFlag')).nativeElement).toBeDefined();
+        const enabledFixtureElement = UnitTestingUtils.getByCSS(enabledFixture.debugElement, '#underFeatureFlag');
+        expect(enabledFixtureElement).toBeDefined();
+        expect(enabledFixtureElement.nativeElement).toBeDefined();
     });
 
     it('should not render the element with disabled features', () => {
-        expect(disabledFixture.debugElement.query(By.css('#underFeatureFlag'))).toBeNull();
+        expect(UnitTestingUtils.getByCSS(disabledFixture.debugElement, '#underFeatureFlag')).toBeNull();
     });
 });
--- a/lib/core/feature-flags/src/lib/directives/not-features.directive.spec.ts
+++ b/lib/core/feature-flags/src/lib/directives/not-features.directive.spec.ts
@@ -18,9 +18,9 @@
 import { CommonModule } from '@angular/common';
 import { Component } from '@angular/core';
 import { ComponentFixture, TestBed } from '@angular/core/testing';
-import { By } from '@angular/platform-browser';
 import { provideMockFeatureFlags } from '../mocks/features-service-mock.factory';
 import { NotFeaturesDirective } from './not-features.directive';
+import { UnitTestingUtils } from '@alfresco/adf-core';
 
 @Component({
     template: `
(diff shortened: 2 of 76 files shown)


dominikiwanekhyland (approved):

lib/core/src/lib/testing/unit-testing-utils.ts
```diff
@@ -0,0 +1,465 @@
...
+
+    static getAllByDataAutomationId(debugElement: DebugElement, dataAutomationId: string): DebugElement[] {
+        return UnitTestingUtils.getAllByCSS(debugElement, `[data-automation-id="${dataAutomationId}"]`);
+    }
+
+    static getInnerTextByDataAutomationId(debugElement: DebugElement, dataAutomationId: string): string {
+        return UnitTestingUtils.getByDataAutomationId(debugElement, dataAutomationId).nativeElement.innerText;
+    }
+
+    static getByDirective(debugElement: DebugElement, directive: any): DebugElement {
```
AleksanderSklorz, line 65: instead of any for directive, could you please use ``Type<any>``? It will be a bit safer because it won't allow some values like for example strings.

lib/core/src/lib/testing/unit-testing-utils.ts
```diff
@@ -0,0 +1,465 @@
...
+
+    static getInputByCSS(debugElement: DebugElement, selector: string): HTMLInputElement {
+        return UnitTestingUtils.getByCSS(debugElement, selector)?.nativeElement;
+    }
+
+    static getInputByDataAutomationId(debugElement: DebugElement, dataAutomationId: string): HTMLInputElement {
+        return UnitTestingUtils.getByDataAutomationId(debugElement, dataAutomationId)?.nativeElement;
+    }
+
+    static fillInputByCSS(debugElement: DebugElement, selector: string, value: any): void {
```
AleksanderSklorz, line 130: I think we can set value to string type because we know it should be string as value property of input has string type: 
Screenshot 2025-01-24 at 11 29 10

lib/core/src/lib/testing/unit-testing-utils.ts
```diff
@@ -0,0 +1,465 @@
...
+        const input = UnitTestingUtils.getInputByCSS(debugElement, selector);
+        input.value = value;
+        input.dispatchEvent(new Event('input'));
+    }
+
+    static fillInputByDataAutomationId(debugElement: DebugElement, dataAutomationId: string, value: any): void {
```
AleksanderSklorz, line 136: similar situation like above - we can set value to string because value of input has string type

lib/core/src/lib/testing/unit-testing-utils.ts
```diff
@@ -0,0 +1,465 @@
...
+import { MatErrorHarness, MatFormFieldHarness } from '@angular/material/form-field/testing';
+import { MatInputHarness } from '@angular/material/input/testing';
+import { MatAutocompleteHarness } from '@angular/material/autocomplete/testing';
+import { ComponentFixture } from '@angular/core/testing';
+import { MatTabGroupHarness, MatTabHarness } from '@angular/material/tabs/testing';
+import { MatToolbarHarness } from '@angular/material/toolbar/testing';
+import { MatSnackBarHarness } from '@angular/material/snack-bar/testing';
+import { MatProgressBarHarness } from '@angular/material/progress-bar/testing';
+
+export class UnitTestingUtils {
```
AleksanderSklorz, line 36: I'm wondering on one thing. Actually all functions are static what causes that we need to repeat for each function things like debugElement or loader. I'm wondering if that would not be better if instead of static functions we would have some instance of helper object and instance functions. In that case we can pass whole fixture to helper's constructor and then base on that fixture the helper can get debugElement or loader automatically by itself. No need to pass these values to repetitively for each function. We can eventually leave these parameters as optional for example in case somebody needs to use different debugElement than from fixture but I guess it would remove lot of repeated code because in most cases we will use it from fixture. 

But I'm fine with static approach too, just wondering what will be better :)

lib/core/feature-flags/src/lib/directives/features.directive.spec.ts
```diff
@@ -20,7 +20,7 @@ import { Component } from '@angular/core';
 import { ComponentFixture, TestBed } from '@angular/core/testing';
 import { provideMockFeatureFlags } from '../mocks/features-service-mock.factory';
 import { FeaturesDirective } from './features.directive';
-import { By } from '@angular/platform-browser';
+import { UnitTestingUtils } from '@alfresco/adf-core';
```
swapnil-verma-gl, line 23: You are already in core lib, so import from @alfresco/adf-core is incorrect

lib/core/feature-flags/src/lib/directives/not-features.directive.spec.ts
```diff
@@ -18,9 +18,9 @@
 import { CommonModule } from '@angular/common';
 import { Component } from '@angular/core';
 import { ComponentFixture, TestBed } from '@angular/core/testing';
-import { By } from '@angular/platform-browser';
 import { provideMockFeatureFlags } from '../mocks/features-service-mock.factory';
 import { NotFeaturesDirective } from './not-features.directive';
+import { UnitTestingUtils } from '@alfresco/adf-core';
```
swapnil-verma-gl, line 23: Possibly incorrect import

lib/core/src/lib/avatar/avatar.component.spec.ts
```diff
@@ -15,7 +15,7 @@
  * limitations under the License.
  */
 
-import { AvatarComponent } from '@alfresco/adf-core';
+import { AvatarComponent, UnitTestingUtils } from '@alfresco/adf-core';
```
swapnil-verma-gl, line 18: Possibly incorrect import

lib/core/src/lib/testing/unit-testing-utils.ts
MichalKinas: Good idea, done

lib/core/src/lib/testing/unit-testing-utils.ts
MichalKinas: I like this approach, changed :D

lib/core/feature-flags/src/lib/directives/features.directive.spec.ts
MichalKinas: Right I missed it, fixed

lib/core/feature-flags/src/lib/directives/not-features.directive.spec.ts
MichalKinas: Fixed

lib/core/src/lib/testing/unit-testing-utils.ts
```diff
@@ -0,0 +1,477 @@
...
+
+    getByDataAutomationId(dataAutomationId: string): DebugElement {
+        return this.getByCSS(`[data-automation-id="${dataAutomationId}"]`);
+    }
+
+    getByDataAutomationClass(dataAutomationClass: string): DebugElement {
+        return this.getByCSS(`[data-automation-class="${dataAutomationClass}"]`);
+    }
+
+    getAllByDataAutomationId(dataAutomationId: string): DebugElement[] {
```
AleksanderSklorz, line 73: I'm wondering if there is a sense to have this function because ideally data automation ids should not be repeated. If we have some array then ideally we should have some index/id like: 
data-automation-id="form-input-1"
data-automation-id="form-input-2"
data-automation-id="form-input-3"
....
etc.

Maybe it would have sense to add possibility to change also operator from = to for example ^= (then it search for each element which has data-automation-id starting from for example "form-input")? See doc for others operators: https://www.w3schools.com/css/css_attribute_selectors.asp

lib/core/src/lib/testing/unit-testing-utils.ts
```diff
@@ -0,0 +1,477 @@
...
+        const input = await this.getMatInput();
+        return input.getValue();
+    }
+
+    async getMatInputValueByDataAutomationId(dataAutomationId: string): Promise<string> {
+        const input = await this.getMatInputByDataAutomationId(dataAutomationId);
+        return input.getValue();
+    }
+
+    async sendKeysToMatInput(keys: string[] | TestKey[]): Promise<void> {
```
AleksanderSklorz, line 420: Small thing but what if somebody would like to pass mixed sequence of keys like - characters which end with some special character? Maybe we should change type to the same type which is allowed by sendKeys: 
keys: (string | TestKey)[]
This type allows us to mix strings with TestKey inside single array

lib/core/src/lib/testing/unit-testing-utils.ts
```diff
@@ -0,0 +1,477 @@
...
+        const button = await this.getMatButtonByCSS(selector);
+        await button.click();
+    }
+
+    async clickMatButtonByDataAutomationId(dataAutomationId: string): Promise<void> {
+        const button = await this.getMatButtonByDataAutomationId(dataAutomationId);
+        await button.click();
+    }
+
+    async sendKeysToMatButton(keys: string[] | TestKey[]): Promise<void> {
```
AleksanderSklorz, line 195: we can apply the same comment like https://github.com/Alfresco/alfresco-ng2-components/pull/10571/files#r1931893317 here

lib/core/src/lib/testing/unit-testing-utils.ts
```diff
@@ -0,0 +1,477 @@
...
+import { MatTabGroupHarness, MatTabHarness } from '@angular/material/tabs/testing';
+import { MatToolbarHarness } from '@angular/material/toolbar/testing';
+import { MatSnackBarHarness } from '@angular/material/snack-bar/testing';
+import { MatProgressBarHarness } from '@angular/material/progress-bar/testing';
+
+export class UnitTestingUtils {
+    private debugElement: DebugElement;
+    private loader: HarnessLoader;
+
+    constructor(debugElement?: DebugElement, loader?: HarnessLoader) {
```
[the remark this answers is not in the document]
MichalKinas: Good point, done

lib/core/src/lib/testing/unit-testing-utils.ts
MichalKinas: Done

lib/core/src/lib/testing/unit-testing-utils.ts
MichalKinas: Done

AleksanderSklorz (approved):

PR #10591, opened 2025-01-27
[ACS-9119] Fixed textOrientation getting reset on arabic after browser refresh

swapnil-verma-gl:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

For right-to-left languages (such as arabic), textOrientation would get reset after browser refresh. Additionally, there was additional margin getting added on the right when changing browser window sizes

Fixed textOrientation and margin-right issues

> - [ ] Yes
> - [x] No

**Other information**:
https://hyland.atlassian.net/browse/ACS-9119

### DIFF
--- a/lib/core/src/lib/common/services/user-preferences.service.ts
+++ b/lib/core/src/lib/common/services/user-preferences.service.ts
@@ -247,12 +247,15 @@ export class UserPreferencesService {
 
     private getLanguageByKey(key: string): LanguageItem {
         const defaultLanguage = { key: 'en' } as LanguageItem;
+        let language: LanguageItem;
 
         const customLanguages = this.appConfig.get<Array<LanguageItem>>(AppConfigValues.APP_CONFIG_LANGUAGES_KEY);
         if (customLanguages && Array.isArray(customLanguages)) {
-            return customLanguages.find((language) => key.includes(language.key)) || defaultLanguage;
-        } else {
-            return DEFAULT_LANGUAGE_LIST.find((language) => language.key === key) ?? defaultLanguage;
+            language = customLanguages.find((language) => key.includes(language.key));
         }
+        if (!language) {
+            language = DEFAULT_LANGUAGE_LIST.find((language) => language.key === key) ?? defaultLanguage;
+        }
+        return language;
     }
 }
--- a/lib/core/src/lib/common/services/user-preferences.service.ts
+++ b/lib/core/src/lib/common/services/user-preferences.service.ts
@@ -250,12 +250,10 @@ export class UserPreferencesService {
         let language: LanguageItem;
 
         const customLanguages = this.appConfig.get<Array<LanguageItem>>(AppConfigValues.APP_CONFIG_LANGUAGES_KEY);
-        if (customLanguages && Array.isArray(customLanguages)) {
-            language = customLanguages.find((language) => key.includes(language.key));
+        if (Array.isArray(customLanguages)) {
+            language = customLanguages.find((customLanguage) => key.includes(customLanguage.key));
         }
-        if (!language) {
-            language = DEFAULT_LANGUAGE_LIST.find((language) => language.key === key) ?? defaultLanguage;
-        }
+        language ??= DEFAULT_LANGUAGE_LIST.find((defaultLang) => defaultLang.key === key) ?? defaultLanguage;
         return language;
     }
 }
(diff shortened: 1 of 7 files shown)


dominikiwanekhyland (approved):

AleksanderSklorz:
@swapnil-verma-gl text orientation is kept after page refreshing but I see there is some blank space after that: 

https://github.com/user-attachments/assets/dd23aa93-a426-40e0-b4b6-0209b412e523

lib/core/src/lib/common/services/user-preferences.service.ts
```diff
@@ -246,11 +247,15 @@ export class UserPreferencesService {
 
     private getLanguageByKey(key: string): LanguageItem {
         const defaultLanguage = { key: 'en' } as LanguageItem;
+        let language: LanguageItem;
 
-        const registeredLanguages = this.appConfig.get<Array<LanguageItem>>(AppConfigValues.APP_CONFIG_LANGUAGES_KEY);
-        if (registeredLanguages && Array.isArray(registeredLanguages)) {
-            return registeredLanguages.find((language) => key.includes(language.key)) || defaultLanguage;
+        const customLanguages = this.appConfig.get<Array<LanguageItem>>(AppConfigValues.APP_CONFIG_LANGUAGES_KEY);
+        if (customLanguages && Array.isArray(customLanguages)) {
```
AleksanderSklorz, line 253: no need for both conditions, Array.isArray is enough because if customLanguages will be null or undefined then isArray gives just false because null and undefined aren't arrays.

lib/core/src/lib/common/services/user-preferences.service.ts
```diff
@@ -246,11 +247,15 @@ export class UserPreferencesService {
...
+            language = customLanguages.find((language) => key.includes(language.key));
         }
-        return defaultLanguage;
+        if (!language) {
+            language = DEFAULT_LANGUAGE_LIST.find((language) => language.key === key) ?? defaultLanguage;
```
AleksanderSklorz, line 257: instead of if we can write it a bit shorter: 
```ts
language ??= DEFAULT_LANGUAGE_LIST.find((language) => language.key === key) ?? defaultLanguage;
```
It will assign right part only when language is null or undefined, see doc: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_assignment

AleksanderSklorz:
@swapnil-verma-gl  there are some sonar issues, please check them if can be fixed 
image

lib/core/src/lib/common/services/user-preferences.service.ts
swapnil-verma-gl: The condition was present before as well. However, since the suggestion does add value, and removes unnecessary code, updated the condition

lib/core/src/lib/common/services/user-preferences.service.ts
swapnil-verma-gl: That looks much cleaner now, thanks!!

swapnil-verma-gl:
> @swapnil-verma-gl text orientation is kept after page refreshing but I see there is some blank space after that:
> 
>  Screen.Recording.2025-01-27.at.14.46.49.mov

I thought that the autosize attribute in layour-container-component.html fixed that. However, I see that that is still not the case. I'll push a fix for that. Thanks!

lib/core/src/lib/layout/components/layout-container/layout-container.component.scss
```diff
@@ -37,6 +37,7 @@ adf-layout-container {
 #{$mat-sidenav-content},
 #{$mat-drawer-transition} #{$mat-drawer-content} {
     margin-left: 0 !important;
+    margin-right: 0 !important;
```
AleksanderSklorz, line 40: Just to confirm - there is no other way to remove that space and adding !important in styles is required? (asking because we should to avoid !important)

swapnil-verma-gl: Unfortunately yes. The extra spacing being added was due to margin-right getting added on to the mat-sidenav-content element, via inline CSS. The only way I can think of to have a greater specificity than inline CSS is via the use of !important flag.

AleksanderSklorz (approved):

PR #10596, opened 2025-01-28
[ACS-5184] Show progress spinner on file navigation

nikita-web-ua:

> - [X] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://hyland.atlassian.net/browse/ACS-5184

Navigating to the next file will trigger spinner loading until the next file's data is ready to be rendered.

> - [ ] Yes
> - [X] No

**Other information**:


MichalKinas (approved):

swapnil-verma-gl (approved):

nikita-web-ua:
> Two things:
> 
> 1. If you have pdf then you can see spinner and after that loading bar. Can we show only one thing (like for example only spinner) because it seems a bit strange because spinner means that something is loading so when spinner hides then we expect that everything is loaded but then again loading bar is displayed which also suggest that something is loading?
> 2. Sometimes when I'm switching to image then when spinner hides I can see empty space for a moment which looks like something is not fully loaded yet. Can we somehow resolve that?
> 
>  Screen.Recording.2025-01-28.at.11.59.05.mov

Hi @AleksanderSklorz , mentioned issues should be resolved now

lib/core/src/lib/viewer/components/viewer-render/viewer-render.component.spec.ts
```diff
@@ -476,4 +476,73 @@ describe('ViewerComponent', () => {
...
+        it('should show spinner until content is ready when viewerType is media', () => {
+            component.isLoading = false;
+            component.urlFile = 'some-file.mp4';
+
+            component.ngOnChanges();
+            fixture.detectChanges();
+
+            expect(testingUtils.getByCSS('.adf-viewer-render-main-loader')).not.toBeNull();
+
+            const mediaViewer = testingUtils.getByCSS('adf-media-player');
```
AleksanderSklorz, line 496: instead checking component by tag's name you can use getByDirective function from testingUtils and then get this element by class name (MediaPlayerComponent)

lib/core/src/lib/viewer/components/viewer-render/viewer-render.component.spec.ts
```diff
@@ -476,4 +476,73 @@ describe('ViewerComponent', () => {
...
+        it('should show spinner until content is ready when viewerType is pdf', () => {
+            component.isLoading = false;
+            component.urlFile = 'some-url.pdf';
+
+            component.ngOnChanges();
+            fixture.detectChanges();
+
+            expect(testingUtils.getByCSS('.adf-viewer-render-main-loader')).not.toBeNull();
+
+            const pdfViewer = testingUtils.getByCSS('adf-pdf-viewer');
```
AleksanderSklorz, line 513: same here, please use getByDirective

lib/core/src/lib/viewer/components/viewer-render/viewer-render.component.spec.ts
```diff
@@ -476,4 +476,73 @@ describe('ViewerComponent', () => {
...
+
+        it('should show spinner until content is ready when viewerType is image', () => {
+            component.isLoading = false;
+            component.urlFile = 'some-url.png';
+
+            component.ngOnChanges();
+            fixture.detectChanges();
+            expect(testingUtils.getByCSS('.adf-viewer-render-main-loader')).not.toBeNull();
+
+            const imgViewer = testingUtils.getByCSS('adf-img-viewer');
```
AleksanderSklorz, line 529: same here, please use getByDirective

lib/core/src/lib/viewer/components/viewer-render/viewer-render.component.spec.ts
nikita-web-ua: good point, applied

AleksanderSklorz (approved):

PR #10599, opened 2025-01-29
AAE-30110 Render sections in runtime

tomgny:

> - [ ] Bugfix
> - [x] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://hyland.atlassian.net/browse/AAE-30110

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/core/src/lib/form/components/form-renderer.component.html
+++ b/lib/core/src/lib/form/components/form-renderer.component.html
@@ -38,13 +38,23 @@
                 <section class="adf-grid-list-column-view" *ngIf="currentRootElement?.isExpanded">
                     <div class="adf-grid-list-single-column"
                          *ngFor="let column of currentRootElement?.columns"
-                         [style.width.%]="getColumnWith(currentRootElement)">
+                         [style.width.%]="getColumnWidth(currentRootElement)">
                         <div class="adf-grid-list-column-view-item" *ngFor="let field of column?.fields">
                             <ng-container *ngIf="field.type === 'section'; else formField">
-                                to be implemented
+                                <div class="adf-grid-list-section-single-column"
+                                    [style.visibility]="!field?.isVisible ? 'hidden' : 'visible'"
+                                    [style.display]="!field?.isVisible ? 'none' : 'flex'">
+                                    <div *ngFor="let sectionColumn of field.columns"
+                                        [style.width.%]="getSectionColumnWidth(field.numberOfColumns, sectionColumn.fields)"
+                                    >
+                                        <div *ngFor="let sectionField of sectionColumn.fields" class="adf-grid-list-section-column-view-item">
+                                            <adf-form-field [field]="sectionField"/>
+                                        </div>
+                                    </div>
+                                </div>
                             </ng-container>
                             <ng-template #formField>
-                                <adf-form-field [field]="field" />
+                                <adf-form-field [field]="field"/>
                             </ng-template>
                         </div>
                     </div>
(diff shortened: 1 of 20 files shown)


lib/core/src/lib/form/components/form-renderer.component.html
```diff
@@ -38,13 +38,23 @@
...
                     <div class="adf-grid-list-single-column"
                          *ngFor="let column of currentRootElement?.columns"
-                         [style.width.%]="getColumnWith(currentRootElement)">
+                         [style.width.%]="getColumnWidth(currentRootElement)">
                         <div class="adf-grid-list-column-view-item" *ngFor="let field of column?.fields">
                             <ng-container *ngIf="field.type === 'section'; else formField">
-                                to be implemented
+                                <div class="adf-grid-list-section-single-column"
+                                    [style.visibility]="!field?.isVisible ? 'hidden' : 'visible'"
+                                    [style.display]="!field?.isVisible ? 'none' : 'flex'">
```
wiktord2000, line 46: Why do we need two of them? I mean visibility & display

lib/core/src/lib/form/components/form-renderer.component.scss
```diff
@@ -14,39 +14,55 @@
...
+            &-column-view-item {
+                width: 100%;
+                display: flex;
+                flex-grow: 1;
+                box-sizing: border-box;
+                padding-left: 1%;
+                padding-right: 1%;
+            }
+        }
     }
```
wiktord2000, line 66: Could you include the video how it behaves?

wiktord2000: Or actually we can meet on Monday during call

lib/core/src/lib/form/components/widgets/core/form-field.model.ts
```diff
@@ -282,30 +282,32 @@ export class FormFieldModel extends FormWidgetModel {
...
-                    const col = new ContainerColumnModel();
-
-                    col.fields = (json.fields[currentField] || []).map((field) => new FormFieldModel(form, field));
-                    col.rowspan = json.fields[currentField].length;
-
-                    col.fields.forEach((colFields: any) => {
-                        this.colspan = colFields.colspan > this.colspan ? colFields.colspan : this.colspan;
-                    });
+        Object.keys(fields).forEach((currentField) => {
+            if (Object.prototype.hasOwnProperty.call(fields, currentField)) {
```
wiktord2000, line 292: if (!Object.prototype.hasOwnProperty.call(fields, currentField)) {
    return;
}

(Optional suggestion)

lib/core/src/lib/form/components/widgets/core/form.model.ts
```diff
@@ -340,30 +340,65 @@ export class FormModel implements ProcessFormModel {
...
+        });
+    }
+
+    private isContainerField(field: ContainerModel | FormFieldModel): field is ContainerModel {
+        return field instanceof ContainerModel;
+    }
+
+    private isSectionField(field: ContainerModel | FormFieldModel): field is FormFieldModel {
+        return field.type === FormFieldTypes.SECTION;
+    }
```
wiktord2000, line 371: Interesting syntax, didn't know about that :)

wiktord2000 (approved):

lib/core/src/lib/form/components/widgets/core/form.model.ts
tomgny: this is useful when the parameter can have multiple types (btw we should avoid that but here, the change would be a breaking change)

lib/core/src/lib/form/components/form-renderer.component.html
tomgny: good point, i think display is enough. difference between these two is for visibility we still keep the element in the layout, for display we remove it. and because we rely on one property, setting visibility is redundant, i will remove it from here and also from adf-form-field where we initially had this

PR #10668, opened 2025-02-21
[ACS-9267] a11y testing: Datatable - columns settings menu

nikita-web-ua:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://hyland.atlassian.net/browse/ACS-9267

- resolved a11y issues related to adf-columns-selector menu
- menu accessible by keyboard
- fixed a bug where the menu would instantly close when opened with the Enter key

> - [ ] Yes
> - [X] No

**Other information**:

### DIFF
--- a/lib/content-services/src/lib/document-list/components/document-list.component.spec.ts
+++ b/lib/content-services/src/lib/document-list/components/document-list.component.spec.ts
@@ -1137,7 +1137,7 @@ describe('DocumentList', () => {
     it('should display [empty folder] template ', () => {
         fixture.detectChanges();
         runInInjectionContext(injector, () => {
-            documentList.dataTable = new DataTableComponent(null, null, matIconRegistryMock, domSanitizerMock);
+            documentList.dataTable = new DataTableComponent(null, null, matIconRegistryMock, domSanitizerMock, null);
         });
         expect(documentList.dataTable).toBeDefined();
         expect(fixture.debugElement.query(By.css('adf-empty-list'))).not.toBeNull();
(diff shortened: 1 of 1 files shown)


lib/core/src/lib/datatable/components/columns-selector/columns-selector.component.html
```diff
@@ -33,6 +34,7 @@
             class="adf-columns-selector-search-input"
             data-automation-id="adf-columns-selector-search-input"
             type="text"
+            role="menuitem"
```
AleksanderSklorz, line 37: I think this change causes new accessibility issue in axe tool: 
image

nikita-web-ua: hmm, I don't see this one 
Screenshot 2025-02-21 at 13 31 32

nikita-web-ua: I see it now, it was hidden by Best Practices flag, but removing role="menuitem" from this input will result to the following issue which is marked as critical impact, while the one you've attached has minor impact 
Screenshot 2025-02-21 at 13 39 14

AleksanderSklorz: I think it may be because I see on your screenshot that you have set off Best Practices option in axe. Please set on this option and then you should see that issue @nikita-web-ua

AleksanderSklorz: @nikita-web-ua ok, I got it. Could you please check if there is any way to resolve both issues? If there is no way to fix both then please just add comment on your jira that this fix introduced minor issue and document that we can't fix it because it introduces critical issue. Is that fine?

AleksanderSklorz:
@nikita-web-ua I can see 6 sonar issues on that PR, can we fix them somehow?
image

lib/core/src/lib/datatable/components/columns-selector/columns-selector.component.html
nikita-web-ua: added comment to the ticket

AleksanderSklorz: thank you

AleksanderSklorz (approved):

PR #10711, opened 2025-03-12
AAE-27160 Fix the structure of data returned from service-tasks API

sdonnell-hyland:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

[AAE-27160](https://hyland.atlassian.net/browse/AAE-27160)

Columns in the service task list cannot be sorted.  This is because the sortKey doesn't match column key since there is an additional layer in the data (`entry`).  Attempting to update the sortKey to match, for example `entry.startDate`, allows the columns to sort correctly, but API calls include `entry.startDate` as the column name in query parameters.

During review, I noticed some mock data was actually correct and contained the `entry` node in data.  This change is breaking, but provides consistency with similar components.

Server data is filtered and mapped to exclude the extra `entry` layer.

> - [x] Yes
> - [ ] No

Applications that use this component will need to update the `key` value in their column definitions as well as update the success callback and ng-template that use the data.  For Hyland Experience, this specifically means the app.config.json templates need to be updated for both admin-apa and admin-hxp.

[AAE-27160]: https://hyland.atlassian.net/browse/AAE-27160?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/process-services-cloud/src/lib/task/task-list/components/service-task-list/service-task-list-cloud.component.spec.ts
+++ b/lib/process-services-cloud/src/lib/task/task-list/components/service-task-list/service-task-list-cloud.component.spec.ts
@@ -230,7 +230,7 @@ describe('ServiceTaskListCloudComponent', () => {
 
     describe('component changes', () => {
         beforeEach(() => {
-            component.rows = fakeServiceTask.list.entries;
+            component.rows = fakeServiceTask.list.entries.map((task) => task.entry);
             fixture.detectChanges();
         });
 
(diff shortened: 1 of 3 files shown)


dthornton-hyl (approved):
Approve if others agree with the breaking change.

jcain82 (approved):

DenysVuika (approved):

PR #10743, opened 2025-03-24
[ACS-9406] Add getContentRenditionTypePreview to process-content service

nikita-web-ua:

> - [ ] Bugfix
> - [X] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://hyland.atlassian.net/browse/ACS-9406

getContentRenditionTypePreview  method added to process-content service to get the preview type rendition for a related content file

> - [ ] Yes
> - [X] No

**Other information**:

### DIFF
--- a/lib/process-services/src/lib/form/services/process-content.service.spec.ts
+++ b/lib/process-services/src/lib/form/services/process-content.service.spec.ts
@@ -186,4 +186,15 @@ describe('ProcessContentService', () => {
             done();
         });
     });
+
+    it('should return a Blob as preview', (done) => {
+        const blob = createFakeBlob();
+        spyOn(service, 'getContentRenditionTypePreview').and.returnValue(of(blob));
+        service.getContentRenditionTypePreview(999).subscribe((result) => {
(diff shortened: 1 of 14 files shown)


lib/process-services/src/lib/form/services/process-content.service.spec.ts
```diff
@@ -186,4 +186,15 @@ describe('ProcessContentService', () => {
             done();
         });
     });
+
+    it('should return a Blob as preview', (done) => {
+        const blob = createFakeBlob();
+        spyOn(service, 'getContentRenditionTypePreview').and.returnValue(of(blob));
```
AleksanderSklorz, line 192: This test is incorrect (and above tests too as they are similar). Here getContentRenditionTypePreview function is mocked and then it's tested inside subscribe if response is correct but the response will be always correct (as it's mocked) even when logic of function will be broken. Instead we should mock getRawContent function from contentApi

nikita-web-ua: done

dominikiwanekhyland (approved):

PR #10781, opened 2025-04-10
AAE-30358 Token is not refreshed in WS connection_init when changing pages

alep85:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://hyland.atlassian.net/browse/AAE-30358

> - [ ] Yes
> - [ ] No

**Other information**:


lib/process-services-cloud/src/lib/services/web-socket.service.ts
```diff
@@ -135,9 +141,25 @@ export class WebSocketService {
...
             },
-            attempts: {
-                max: 5,
-                retryIf: (error) => !!error
+            attempts: (count: number, _operation: Operation, error: any) => {
+                if (!error) {
+                    return false;
+                }
+
+                const isUnauthorizedError =
```
VitoAlbano, line 149: I think we can reuse this part to actually catch the error from network when it happens and then try to refresh the token. Same with the retry mechanism instead of having a plain number we can use the retryIf

PR #10795, opened 2025-04-16
Ng19 migration latest

dominikiwanekhyland:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

> - [ ] Yes
> - [ ] No

**Other information**:


package.json
```diff
@@ -162,24 +161,25 @@
...
     "shelljs": "^0.8.5",
     "spdx-license-list": "^6.9.0",
     "storybook": "^8.2.9",
     "stylelint": "16.8.2",
     "stylelint-config-standard-scss": "^13.1.0",
     "ts-node": "^10.9.2",
     "tsconfig-paths": "^4.1.1",
-    "typescript": "5.3.3",
-    "webpack": "5.97.1",
+    "typescript": "5.7.3",
```
AleksanderSklorz, line 180: based on compatibility documentation we can use even higher ts version (we can increase it to 5.8.x - any version before 5.9.x) because Angular is upgraded to 19.2.x 
https://angular.dev/reference/versions#actively-supported-versions

lib/content-services/src/lib/content-metadata/components/content-metadata-card/content-metadata-card.component.ts
```diff
@@ -31,7 +31,6 @@ import { TranslateModule } from '@ngx-translate/core';
 
 @Component({
     selector: 'adf-content-metadata-card',
-    standalone: true,
```
VitoAlbano, line 34: Just as extra info: Careful here, we decided to move fully on standalone and reverting this will create issues along the way for sure @DenysVuika

dominikiwanekhyland: @VitoAlbano thanks, in angular 19 **standalone: true** is default, so I removed those adnotations :)

VitoAlbano: My fear is that this will create a lot of conflicts when rebasing but that should be fine for now

PR #10856, opened 2025-05-14
Fix fieldsCache for nested empty group

krzysztofsk:

> - [ X] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

In my case an empty array is added to fieldsCache when empty group is nested inside a container and field validation is failing.

It ignores the empty array

> - [ ] Yes
> - [ x] No

**Other information**:


lib/core/src/lib/form/components/widgets/core/form.model.ts
```diff
@@ -387,9 +387,11 @@ export class FormModel implements ProcessFormModel {
     }
 
     private handleSingleField(field: FormFieldModel, formFieldModel: FormFieldModel[]): void {
-        formFieldModel.push(field);
-        if (field.fields) {
-            this.processFields(Object.values(field.fields), formFieldModel);
+        if (field instanceof FormFieldModel) {
```
DenysVuika, line 390: missing unit tests

DenysVuika (changes requested):

lib/core/src/lib/form/components/widgets/core/form.model.ts
tomgny, line 390: i would also suggest to move this condition to separate method, like 
    private isFormField(field: ContainerModel | FormFieldModel): field is FormFieldModel {
        return field instanceof FormFieldModel;
    }
    
    and use it at the else if in the processFields method, to be consistent with the use of isContainerField.

tomgny:
we need to check if these changes are compatible with our repositories that are using ng2-components. i will update you about the results here (currently we have different problem which blocks us)

tomgny (approved):
all affected test passed on our private repository, approved

DenysVuika (approved):

tomgny:
@krzysztofsk i cloned your changes to separate PR and merged it -> https://github.com/Alfresco/alfresco-ng2-components/pull/10861
For some reason (probably your branch name or target branch) it was not possible to upload artifacts.
To use implemented changes, you need to use latest version or ADF instead of latest, major one. (until we release next, minor version). 
this version contains changes you need -> https://www.npmjs.com/package/@alfresco/adf-core/v/8.1.0-15139516875

PR #10858, opened 2025-05-14
[ACS-9619] Issues with trimming file's name in viewer

dominikiwanekhyland:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://hyland.atlassian.net/browse/ACS-9619

> - [ ] Yes
> - [x] No

**Other information**:


AleksanderSklorz:
@dominikiwanekhyland I see that solution makes some new issue. Now we have two ellipsis when name and extension are long but it should have only single ellipsis because it's single word (so when we have some name like fileName.png then that "fileName.png" is one word so should have only one ellipsis but now we have 2), see screenshot for following name: 

asdasdasdsadasdsadasdasdasdsadasdasdsadasdsadsadasdasdasdasdasdsaadasd.asdasdasdsadasdsadasdasdasdsadasdasdsadasdsadsadasdasdasdasdasdsaadasd

image


You see it has two separated ellipsis for name and for extension although it's one word. It should work like for example: 

veryLongNameveryLongNameveryLongNameveryLongNameveryLongName.veryLongExtensionveryLongExtension

should have ... in the middle like: 
veryLongName...veryLongExte

Then it makes more sense in my opinion because user can see part of name and part of extension and there is only one ellipsis as we should treat it as one thing

dominikiwanekhyland:
> @dominikiwanekhyland I see that solution makes some new issue. Now we have two ellipsis when name and extension are long but it should have only single ellipsis because it's single word (so when we have some name like fileName.png then that "fileName.png" is one word so should have only one ellipsis but now we have 2), see screenshot for following name:
> 
> asdasdasdsadasdsadasdasdasdsadasdasdsadasdsadsadasdasdasdasdasdsaadasd.asdasdasdsadasdsadasdasdasdsadasdasdsadasdsadsadasdasdasdasdasdsaadasd
> 
> image
> You see it has two separated ellipsis for name and for extension although it's one word. It should work like for example:
> 
> veryLongNameveryLongNameveryLongNameveryLongNameveryLongName.veryLongExtensionveryLongExtension
> 
> should have ... in the middle like: veryLongName...ongExtension
> 
> Then it makes more sense in my opinion because user can see part of name and part of extension and there is only one ellipsis as we should treat it as one thing. So basically in above case it should to cut end of file name and beginning of extension

Hi @AleksanderSklorz , I thought about it and I thought it would be better to see both beginning of name and beginning of file format, rather than some of the ending letters which I don't think would be informative at all. Beside it would look like it's end of file name rather than file format name, which I think could be also confusing. What do you think?

AleksanderSklorz:
> > @dominikiwanekhyland I see that solution makes some new issue. Now we have two ellipsis when name and extension are long but it should have only single ellipsis because it's single word (so when we have some name like fileName.png then that "fileName.png" is one word so should have only one ellipsis but now we have 2), see screenshot for following name:
> > asdasdasdsadasdsadasdasdasdsadasdasdsadasdsadsadasdasdasdasdasdsaadasd.asdasdasdsadasdsadasdasdasdsadasdasdsadasdsadsadasdasdasdasdasdsaadasd
> > image
> > You see it has two separated ellipsis for name and for extension although it's one word. It should work like for example:
> > veryLongNameveryLongNameveryLongNameveryLongNameveryLongName.veryLongExtensionveryLongExtension
> > should have ... in the middle like: veryLongName...ongExtension
> > Then it makes more sense in my opinion because user can see part of name and part of extension and there is only one ellipsis as we should treat it as one thing. So basically in above case it should to cut end of file name and beginning of extension
> 
> Hi @AleksanderSklorz , I thought about it and I thought it would be better to see both beginning of name and beginning of file format, rather than some of the ending letters which I don't think would be informative at all. Beside it would look like it's end of file name rather than file format name, which I think could be also confusing. What do you think?

Hi @dominikiwanekhyland. I have asked @MichalKinas about help to decide which approach will be better. He will comment here soon

lib/core/src/lib/viewer/components/viewer.component.spec.ts
```diff
@@ -143,36 +143,63 @@ describe('ViewerComponent', () => {
...
 
-        it('should set fileName providing fileName input', () => {
-            component.fileName = 'testFileName.jpg';
-            spyOn(viewUtilService, 'getFilenameFromUrl').and.returnValue('fakeFileName.jpeg');
-            const mockSimpleChanges: any = { urlFile: { currentValue: 'https://fakefile.url/fakeFileName.jpeg' } };
+        describe('fileName setter integration', () => {
+            it('should fileName be set by urlFile input if the fileName is not provided as Input', () => {
+                component.fileName = '';
+                spyOn(viewUtilService, 'getFilenameFromUrl').and.returnValue('fakeFileName.jpeg');
+                const mockSimpleChanges: any = { urlFile: { currentValue: 'https://fakefile.url/fakeFileName.jpeg' } };
```
AleksanderSklorz, line 185: can we mock proper value to avoid eslint issues for any?
image

lib/core/src/lib/viewer/components/viewer.component.spec.ts
```diff
@@ -143,36 +143,63 @@ describe('ViewerComponent', () => {
         });
     });
 
-    describe('File Name Test', () => {
-        const getFileNameWithoutExtension = (): string =>
-            testingUtils.getByCSS('.adf-viewer__display-name-without-extension').nativeElement.textContent;
-        const getExtension = (): string => testingUtils.getByCSS('.adf-viewer__display-name-extension').nativeElement.textContent;
+    describe('File Name Display Tests', () => {
+        const getFileName = (): string => testingUtils.getByCSS('#adf-viewer-display-name').nativeElement.textContent;
```
AleksanderSklorz, line 147: there is already implemented getFileName function in 54th line. Please reuse it

PR #10860, opened 2025-05-14
AAE-34675 Fix default selection required

eromano:

> - [X] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

AAE-34675

> - [ ] Yes
> - [ ] No

**Other information**:


lib/process-services-cloud/src/lib/form/components/widgets/dropdown/dropdown-cloud.widget.ts
```diff
@@ -195,7 +195,7 @@ export class DropdownCloudWidgetComponent extends WidgetComponent implements OnI
     }
 
     private setFormControlValue(): void {
-        this.dropdownControl.setValue(this.field?.value, { emitEvent: false });
+        this.dropdownControl.setValue({ id: this.field?.value, name: '' }, { emitEvent: false });
```
BSekula, line 198: Do not understand this, we should set value but we are updating a `id` and `name` 🤔 

If mapping is not working, maybe just calling `updateFormControlState` would be enough?  it will revalidate field again?

cc @tomgny  I think you have some more knowledge about it WDYT?

eromano: that form control should have id and name. we set it in this way when it come from the mapping correct

lib/process-services-cloud/src/lib/form/components/widgets/dropdown/validator.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+    it('should return a required error when no valid option is selected', () => {
+        const validator = defaultValueValidator(mockField);
+        const control = new FormControl(null);
+
+        const result = validator(control);
+
+        expect(result).toEqual({ required: true });
+    });
+
+    it('should return a required error when the default option is selected', () => {
```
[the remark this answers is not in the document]
eromano: default is not valid is an empty i will rename it to be more clear

eromano:
The value was there but it wasn't the right object

lib/process-services-cloud/src/lib/form/components/widgets/dropdown/dropdown-cloud.widget.ts
BSekula: Ok got it, dropdown value should point to dropdown option using `option id`.

I assume that previous code `this.dropdownControl.setValue(this.field?.value..),` works fine i other cases, therefore `this.field?.value` should had correct object `{id, name}`.

I was thinking if we could map value from mapping into correct object { id, name  }?  instead of updating it here? So we wont break other case (e.g linking dropdowns, async dropdowns, manual value update etc hope we have e2e for it 🤞 )

eromano:
when you select a value inside the value you have:
- {id: name:}
when we instead set if from the mapping value it only contains the id as plain text this is different calling only updateValueAndValidity() doesn't help

lib/process-services-cloud/src/lib/form/components/widgets/dropdown/dropdown-cloud.widget.ts
eromano: i can try

eromano: same that doesn't make our plain an object

BSekula:
> when you select a value inside the value you have:
> 
> * {id: name:}
>   when we instead set if from the mapping value it only contains the id as plain text this is different calling only updateValueAndValidity() doesn't help

as a quick solution we could check if value is an object or not, if object we can just setValue in other case case we can set `this.dropdownControl.setValue({ id: this.field?.value, name: '' })`

eromano:
I can check tomorrow if something is wrong in the mapping indeed

eromano:
Turns out that the mapping can be both an JSON obj or a string so it needs to be handled in the code this double possibility.

BSekula (approved):

PR #10867, opened 2025-05-16
AAE-33909 Ensure onProcessFinish event triggers when invoked from onFormLoaded event

BSekula:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

[AAE-33909](https://hyland.atlassian.net/browse/AAE-33909)

> - [ ] Yes
> - [ ] No

**Other information**:

[AAE-33909]: https://hyland.atlassian.net/browse/AAE-33909?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/process-services-cloud/src/lib/process/start-process/components/start-process-cloud.component.ts
+++ b/lib/process-services-cloud/src/lib/process/start-process/components/start-process-cloud.component.ts
@@ -35,7 +35,8 @@ import {
     FormModel,
     InplaceFormInputComponent,
     LocalizedDatePipe,
-    TranslationService
+    TranslationService,
+    isOutcomeButtonVisible
 } from '@alfresco/adf-core';
 import { AbstractControl, FormControl, FormGroup, ReactiveFormsModule, ValidatorFn, Validators } from '@angular/forms';
 import { MatAutocompleteModule, MatAutocompleteTrigger } from '@angular/material/autocomplete';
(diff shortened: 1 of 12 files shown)


DudaRobert (approved):

lib/process-services-cloud/src/lib/process/start-process/components/start-process-cloud.component.ts
```diff
@@ -95,8 +96,6 @@ export class StartProcessCloudComponent implements OnChanges, OnInit {
     @ViewChild(MatAutocompleteTrigger)
     inputAutocomplete: MatAutocompleteTrigger;
 
-    @ViewChild('startForm') startForm: FormCloudComponent;
```
BSekula, line 98: Currently if the `@Input data` in `FormCloudComponent` does not have any value we are not `refreshing` form, since we do not have nothing to be refreshed, even more if we have some already updated value refreshed we were overriding all current changes. Therefore I moved logic for checking if we have any outcomes into separate file `buttos-visibility/isOutcomeButtonVisible`. So it can be used on both places.

The problem that I had was that the `refreshForm` was emitting second time event onLoadEvent (even if we do not need to reload our form because `data` is undefined). Since now we are not emitting it second time [startForm](https://github.com/Alfresco/alfresco-ng2-components/pull/10867/files#diff-59c36d7292f6b0461ecb6e668d275aacb9357e411bedc530493d43e9bd3e16a0L287) was always undefined.

PR #10903, opened 2025-05-30
[MNT-25043] Skip adding json as content type for FormData body requests

MichalKinas:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://hyland.atlassian.net/browse/MNT-25043

When request with FormData as body is intercepted we should not set json as content type.

> - [ ] Yes
> - [x] No

**Other information**:


lib/core/auth/src/authentication-interceptor/authentication.interceptor.ts
AleksanderSklorz, line 1: could you please add some unit tests for that fix?

MichalKinas: I tried but in that case we have multiple nested observables and I was having a lot issues trying to add some working tests there

AleksanderSklorz: ok, so let's skip them in that case

AleksanderSklorz (approved):

nikita-web-ua (approved):

PR #10923, opened 2025-06-09
AAE-35649 Fix missing spinner when loading in ViewerRenderComponent

wojd0:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://hyland.atlassian.net/browse/AAE-35649

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/core/src/lib/viewer/components/pdf-viewer/pdf-viewer.component.ts
+++ b/lib/core/src/lib/viewer/components/pdf-viewer/pdf-viewer.component.ts
@@ -257,7 +257,10 @@ export class PdfViewerComponent implements OnChanges, OnDestroy {
                 .then(() => {
                     setTimeout(() => this.scalePage('init'));
                 })
-                .catch(() => this.error.emit());
+                .catch((e) => {
+                    console.error(e);
+                    this.error.emit();
+                });
         });
     }
 
--- a/lib/core/src/lib/viewer/components/viewer-render/viewer-render.component.ts
+++ b/lib/core/src/lib/viewer/components/viewer-render/viewer-render.component.ts
@@ -139,6 +139,7 @@ export class ViewerRenderComponent implements OnChanges, OnInit {
     internalFileName: string;
     viewerType: string = 'unknown';
     isLoading = false;
+    isFileLoading = false;
 
     /**
      * Returns a list of the active Viewer content extensions.
@@ -182,6 +183,8 @@ export class ViewerRenderComponent implements OnChanges, OnInit {
     }
 
     ngOnChanges() {
+        this.isFileLoading = !this.blobFile && !this.urlFile;
+
         if (this.blobFile) {
             this.setUpBlobData();
         } else if (this.urlFile) {
--- a/lib/core/src/lib/viewer/components/viewer-render/viewer-render.component.html
+++ b/lib/core/src/lib/viewer/components/viewer-render/viewer-render.component.html
@@ -1,15 +1,17 @@
-<div *ngIf="isLoading" class="adf-viewer-render-main-loader">
-    <div class="adf-viewer-render-layout-content adf-viewer__fullscreen-container">
-        <div class="adf-viewer-render-content-container">
-            <div class="adf-viewer-render__loading-screen">
-                <h2>{{ 'ADF_VIEWER.LOADING' | translate }}</h2>
-                <div>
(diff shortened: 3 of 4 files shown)


lib/core/src/lib/viewer/components/pdf-viewer/pdf-viewer.component.ts
```diff
@@ -257,7 +257,10 @@ export class PdfViewerComponent implements OnChanges, OnDestroy {
                 .then(() => {
                     setTimeout(() => this.scalePage('init'));
                 })
-                .catch(() => this.error.emit());
+                .catch((e) => {
+                    console.error(e);
```
VitoAlbano, line 261: we should have the log service

lib/core/src/lib/viewer/components/viewer-render/viewer-render.component.ts
```diff
@@ -182,6 +183,8 @@ export class ViewerRenderComponent implements OnChanges, OnInit {
     }
 
     ngOnChanges() {
+        this.isFileLoading = !this.blobFile && !this.urlFile;
```
VitoAlbano, line 186: Why having two values like isLoading and isFileLoading?

VitoAlbano (changes requested):

lib/core/src/lib/viewer/components/pdf-viewer/pdf-viewer.component.ts
wojd0: this one? https://github.com/Alfresco/alfresco-ng2-components/blob/AAE-35649-missing-loading-spinner-on-viewer-render-component/dist/libs/core/lib/common/services/log.service.d.ts
marked as deprecated

lib/core/src/lib/viewer/components/viewer-render/viewer-render.component.ts
wojd0: `isLoading` is for the loading state of specific viewers (image, pdf and media), it's true once those viewers are done loading
`isFileLoading` is for the loading state of the file itself, so we don't attempt to render the specific viewer until the file is loaded

VitoAlbano: Can't we have a single condition/method for isLoading so we wrap up inside all the condition in which the viewer shouldn't be ready to be viewed yet?

MichalKinas: I think we can, also this solution may not work well with custom viewers I think, as e.g. urlFile or blob can be ready quite fast but custom viewers may need some more time to actually be ready to display the content, please wait a moment I'm experimenting with the solution I mentioned yesterday which should work for regular and custom viewers

MichalKinas: @wojd0 @VitoAlbano I was thinking about something like this: https://github.com/Alfresco/alfresco-ng2-components/compare/develop...bugfix/dev-mkinas-AAE-35649 which will work also for renderers that are custom and can be registered dynamically, please let me know what you think, if this is fine I will add unit tests, docs etc

MichalKinas: I tested that in ACA with all kinds of viewers, even a custom one that I created and registered and it seems to work fine

wojd0: ok let me have a look on our side, we had problems with viewer component inside mat-tabs

wojd0: @MichalKinas looks like your solution will work better, please create a PR with it so I can close this one

MichalKinas: Sure I will just add some unit tests and docs coverage beforehand

MichalKinas: https://github.com/Alfresco/alfresco-ng2-components/pull/10924 created :D

wojd0:
https://github.com/Alfresco/alfresco-ng2-components/pull/10924 was merged, taking care of this issue

PR #10929, opened 2025-06-12
Fix/aae 34482 modelling inconsistencies in mat form fields label alignments ii

Isoumyajit:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

> - [ ] Yes
> - [x] No

**Other information**:
https://hyland.atlassian.net/browse/AAE-34482


lib/core/src/lib/card-view/components/card-view-selectitem/card-view-selectitem.component.html
```diff
@@ -1,36 +1,46 @@
...
-            <div
-                [attr.data-automation-id]="'card-select-label-' + property.key"
-                class="adf-property-label"
-                [ngClass]="{
-                    'adf-property-value-editable': isEditable,
-                    'adf-property-readonly-value': isReadonlyProperty
-                }"
+            <div *ngIf="!isEditable"
+            [attr.data-automation-id]="'card-select-label-' + property.key"
+            class="adf-property-label adf-property-label-not-editable"
```
VitoAlbano: Shouldn't this be driven by a condition as well? otherwise you will have adf-property-label-not-editable and adf-property-value-editable together which sounds confusing

lib/core/src/lib/card-view/components/card-view-textitem/card-view-textitem.component.html
```diff
@@ -5,7 +5,7 @@
             [ngClass]="{
                 'adf-property-read-only': !isEditable
             }"
-            [floatLabel]="'always'"
+            [floatLabel]="'auto'"
```
VitoAlbano: Why auto here? won't this clash with the placeholder?

VitoAlbano (changes requested):

lib/core/src/lib/card-view/components/card-view-textitem/card-view-textitem.component.html
Isoumyajit: Updated

lib/core/src/lib/card-view/components/card-view-selectitem/card-view-selectitem.component.html
Isoumyajit: I think `adf-property-label-not-editable` will only apply when isEditable is false and if it is false then the `adf-property-valye-editable` will not be applied.

VitoAlbano: ah wait it's worse than i saw it why should we have a condition for an editable style if this part of the HTML is meant to be used when it's not? What's happening here?

Isoumyajit: I updated the code and removed the unnecessary scss classes

lib/core/src/lib/card-view/components/card-view-selectitem/card-view-selectitem.component.html
```diff
@@ -1,36 +1,42 @@
...
                     class="adf-property-value adf-property-read-only"
                     [attr.data-automation-id]="'select-readonly-value-' + property.key"
-                    data-automation-class="read-only-value">{{ (property.displayValue | async) | translate }}
+                    data-automation-class="read-only-value"
+                >
+                    {{ property.displayValue | async | translate }}
                 </div>
                 <div *ngIf="isEditable">
-                    <mat-form-field class="adf-property-value" [ngClass]="{'adf-property-value-editable': isEditable}">
+                    <mat-form-field class="adf-property-value" [ngClass]="{ 'adf-property-value-editable': isEditable }">
```
VitoAlbano: This conditional for the style is redundant and we don't need it

lib/core/src/lib/card-view/components/card-view-selectitem/card-view-selectitem.component.html
```diff
@@ -1,36 +1,42 @@
...
+                        <mat-label
+                            [attr.data-automation-id]="'card-select-label-' + property.key"
+                            class="adf-property-label"
+                            [ngClass]="{
+                                'adf-property-value-editable': isEditable,
```
VitoAlbano: It's still not clear to me why we have this condition here when when have a condition for this section to be shown only when it's editable

lib/core/src/lib/card-view/components/card-view-selectitem/card-view-selectitem.component.spec.ts
```diff
@@ -173,10 +173,11 @@ describe('CardViewSelectItemComponent', () => {
             component.ngOnChanges({});
             component.editable = true;
             fixture.detectChanges();
-
+            const not_editable_label = fixture.nativeElement.querySelector('.adf-property-label-not-editable');
             const field = await testingUtils.getMatFormFieldByCSS('.adf-property-value');
 
-            expect(await field.hasLabel()).toBeFalse();
+            expect(await field.hasLabel()).toBeTrue();
+            expect(not_editable_label).toBeNull();
```
VitoAlbano, line 180: Why instead of using a query to try to fetch the class you don't check if the field has that class? which element you are targeting with that query?

lib/core/src/lib/card-view/components/card-view-textitem/card-view-textitem.component.html
```diff
@@ -111,7 +111,7 @@
         (keyup.enter)="clicked()"
         (click)="clicked()"
     >
-        <mat-form-field class="adf-property-field adf-card-textitem-field" [floatLabel]="'always'">
+        <mat-form-field class="adf-property-field adf-card-textitem-field" [floatLabel]="'auto'">
```
VitoAlbano: Why changing this to 'auto'?

lib/core/src/lib/card-view/components/card-view-textitem/card-view-textitem.component.html
```diff
@@ -82,6 +81,7 @@
             *ngIf="isEditable"
             class="adf-property-field adf-textitem-chip-list-input"
             [ngClass]="{ 'adf-property-read-only': !isEditable }"
+            [floatLabel]="'auto'"
```
VitoAlbano: why adding 'auto'?

lib/core/src/lib/form/components/widgets/number/number.widget.html
```diff
@@ -9,7 +9,7 @@
         </label>
     </div>
     <div>
-        <mat-form-field [hideRequiredMarker]="true" class="adf-form-field-input" [floatLabel]="field.placeholder ? 'always' : 'auto'">
+        <mat-form-field [hideRequiredMarker]="true" class="adf-form-field-input" [floatLabel]="getFloatLabel()">
```
VitoAlbano: please avoid having functions called into the template as this leads to performance issue as they are called an incredible amount of times

lib/process-services-cloud/src/lib/task/task-header/components/task-header-cloud.component.spec.ts
```diff
@@ -164,15 +164,15 @@ describe('TaskHeaderCloudComponent', () => {
             component.refreshData();
             fixture.detectChanges();
             await fixture.whenStable();
-            const valueEl = fixture.debugElement.query(By.css('[data-automation-id="header-dueDate"] .adf-property-value'));
+            const valueEl = fixture.debugElement.query(By.css('[data-automation-id="header-dueDate"] .adf-property-value .adf-datepicker-toggle'));
```
VitoAlbano: This looks wrong : we are trying to get the value of the Element but from the toggle button?

VitoAlbano (changes requested):

lib/core/src/lib/card-view/components/card-view-selectitem/card-view-selectitem.component.html
Isoumyajit: Removed.

lib/core/src/lib/form/components/widgets/number/number.widget.html
Isoumyajit: Changed.

lib/core/src/lib/card-view/components/card-view-selectitem/card-view-selectitem.component.html
Isoumyajit: Removed the unnecessary condition.

lib/core/src/lib/card-view/components/card-view-textitem/card-view-textitem.component.html
Isoumyajit: Removed auto by removing it'll behave as default.

lib/core/src/lib/card-view/components/card-view-textitem/card-view-textitem.component.html
Isoumyajit: Removed.

lib/process-services-cloud/src/lib/task/task-header/components/task-header-cloud.component.spec.ts
Isoumyajit: It's actually the span, which displays value but here if we click on the value then also it works as a toggle button its a feature so I guess that's why the class name is like that.

lib/core/src/lib/card-view/components/card-view-selectitem/card-view-selectitem.component.spec.ts
Isoumyajit: As here we are leveraging the MatFormFieldHarness and we are already having the functions to get the formFieldHarness using the css classes like `getMatFormFieldByCSS()` and we have plenty of more custom functions so I just leverage those functions which we have inside the project itself. and `hasLebel()` is a function which comes from the MatFormFieldharness class. But QuerySelector returns HtmlElement Object so it'll be difficult to follow the consistency how the spec file is actually written.

lib/process-services-cloud/src/lib/task/task-header/components/task-header-cloud.component.spec.ts
VitoAlbano: ok so please rename that variable in something more meaningful

lib/core/src/lib/card-view/components/card-view-selectitem/card-view-selectitem.component.spec.ts
VitoAlbano: Ok but our target is to move and use the Harness library so if there's the possibility to improve and use it please improve it don't just leave it as it is.

lib/process-services-cloud/src/lib/task/task-header/components/task-header-cloud.component.spec.ts
Isoumyajit: Changed the name.

lib/core/src/lib/card-view/components/card-view-selectitem/card-view-selectitem.component.spec.ts
Isoumyajit: I already used the matFormFieldharness Library here.

VitoAlbano (approved):

PR #10949, opened 2025-06-17
[ACS-9224] Accessibility fix for gridcell roles and ARIA labels

Shivangi917:
This PR adds proper ARIA roles (`role="gridcell"`) and `aria-label` attributes to datatable cells in order to improve accessibility compliance.

Related HX PR: https://github.com/Alfresco/hxp-frontend-apps/pull/12572

[ci:force]

### DIFF
--- a/lib/core/src/lib/datatable/components/datatable/datatable.component.html
+++ b/lib/core/src/lib/datatable/components/datatable/datatable.component.html
@@ -18,12 +18,13 @@
 
 
             <!-- Drag -->
-            <div *ngIf="enableDragRows" class="adf-datatable-cell-header adf-drag-column">
+            <div *ngIf="enableDragRows" class="adf-datatable-cell-header adf-drag-column" role="columnheader" tabindex="0" [attr.aria-label]="'ADF-DATATABLE.ACCESSIBILITY.DRAG' | translate">
                 <span class="adf-sr-only">{{ 'ADF-DATATABLE.ACCESSIBILITY.DRAG' | translate }}</span>
             </div>
 
             <!-- Actions (left) -->
-            <div *ngIf="actions && actionsPosition === 'left'" class="adf-actions-column adf-datatable-cell-header">
+            <div *ngIf="actions && actionsPosition === 'left'" class="adf-actions-column adf-datatable-cell-header" role="columnheader" tabindex="0" [attr.aria-label]="'ADF-DATATABLE.ACCESSIBILITY.ACTIONS' | translate"
+>
                 <span class="adf-sr-only">{{ 'ADF-DATATABLE.ACCESSIBILITY.ACTIONS' | translate }}</span>
             </div>
 
@@ -33,7 +34,7 @@
                               [checked]="isSelectAllChecked"
                               (change)="onSelectAllClick($event)"
                               class="adf-checkbox-sr-only"
-                              [aria-label]="'ADF-DATATABLE.ACCESSIBILITY.SELECT_ALL' | translate">
+                              [attr.aria-label]="'ADF-DATATABLE.ACCESSIBILITY.SELECT_ALL' | translate">
                     {{ 'ADF-DATATABLE.ACCESSIBILITY.SELECT_ALL' | translate }}
                 </mat-checkbox>
             </div>
@@ -106,7 +107,13 @@
                         {{ getSortLiveAnnouncement(col) | translate: { string: col.title | translate } }}
                     </span>
 
-                    <span *ngIf="!col.title && !col.sortable && !headerFilterTemplate" [attr.title]="'ADF-DATATABLE.ACCESSIBILITY.EMPTY_HEADER' | translate"></span>
+                    <span
+                        class="adf-sr-only"
+                        [attr.aria-label]="'ADF-DATATABLE.ACCESSIBILITY.EMPTY_HEADER' | translate"
+                        tabindex="0">
+                        {{ 'ADF-DATATABLE.ACCESSIBILITY.EMPTY_HEADER' | translate }}
+                    </span>
+
                 </ng-container>
 
                 <div *ngIf="col.header" class="adf-datatable-cell-value">
@@ -124,6 +131,8 @@
                     *ngIf="col.draggable"
                     cdkDragHandle
                     [ngClass]="{ 'adf-datatable-cell-header-drag-icon': !isResizing }"
+                    aria-hidden="true"
+                    role="presentation"
                 >
                     <adf-icon
                         *ngIf="hoveredHeaderColumnIndex === columnIndex && !isResizing"
@@ -134,14 +143,16 @@
             </div>
             <div
                 *ngIf="isResizingEnabled && col.resizable && !lastColumn"
-                [ngClass]="hoveredHeaderColumnIndex === columnIndex && !isResizing || resizingColumnIndex === columnIndex ? 'adf-datatable__resize-handle-visible' : 'adf-datatable__resize-handle-hidden'"
+                [ngClass]="hoveredHeaderColumnIndex === columnIndex && !isResizing || resizingColumnIndex === columnIndex ? 'adf-datatable_resize-handle-visible' : 'adf-datatable_resize-handle-hidden'"
                 adf-resize-handle
                 tabindex="0"
                 role="button"
                 (click)="$event.stopPropagation()"
                 (keyup.enter)="$event.stopPropagation()"
                 class="adf-datatable__resize-handle"
-                [resizableContainer]="resizableElement">
+                [resizableContainer]="resizableElement"
+                [attr.aria-label]="'ADF-DATATABLE.ACCESSIBILITY.RESIZE_COLUMN' | translate">
+                
                 <div class="adf-datatable__resize-handle--divider"></div>
             </div>
             <div class="adf-drop-header-cell-placeholder" *cdkDragPlaceholder></div>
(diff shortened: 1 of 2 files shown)


lib/core/src/lib/datatable/components/datatable/datatable.component.html
```diff
@@ -16,24 +16,30 @@
             class="adf-datatable-row"
             role="row">
 
-
             <!-- Drag -->
-            <div *ngIf="enableDragRows" class="adf-datatable-cell-header adf-drag-column">
+            <th *ngIf="enableDragRows"
```
MichalKinas: This does not look correct, the datatable row is actually above not on this element

lib/core/src/lib/datatable/components/datatable/datatable.component.html
```diff
@@ -16,24 +16,30 @@
...
+                class="adf-datatable-cell-header adf-drag-column"
+                tabindex="0"
```
MichalKinas: Why do we want to make this element focusable? This is for drag&drop purpose so I don't think it should be focusable with keyboard

lib/core/src/lib/datatable/components/datatable/datatable.component.html
```diff
@@ -16,24 +16,30 @@
...
+                [attr.aria-label]="'ADF-DATATABLE.ACCESSIBILITY.DRAG' | translate">
```
MichalKinas: Why does this element need aria label?

lib/core/src/lib/datatable/components/datatable/datatable.component.html
```diff
@@ -16,24 +16,30 @@
...
                 <span class="adf-sr-only">{{ 'ADF-DATATABLE.ACCESSIBILITY.DRAG' | translate }}</span>
-            </div>
+            </th>
 
             <!-- Actions (left) -->
-            <div *ngIf="actions && actionsPosition === 'left'" class="adf-actions-column adf-datatable-cell-header">
+            <th *ngIf="actions && actionsPosition === 'left'"
```
MichalKinas: You should not nest 2 th within one datatble row, it doesn't really make sense

lib/core/src/lib/datatable/components/datatable/datatable.component.html
```diff
@@ -106,7 +112,13 @@
                         {{ getSortLiveAnnouncement(col) | translate: { string: col.title | translate } }}
                     </span>
 
-                    <span *ngIf="!col.title && !col.sortable && !headerFilterTemplate" [attr.title]="'ADF-DATATABLE.ACCESSIBILITY.EMPTY_HEADER' | translate"></span>
+                    <span
```
MichalKinas: Why did you remobe ngIf from this element? This will cause a bug and seems like a breaking change

lib/core/src/lib/datatable/components/datatable/datatable.component.html
```diff
@@ -124,6 +136,7 @@
                     *ngIf="col.draggable"
                     cdkDragHandle
                     [ngClass]="{ 'adf-datatable-cell-header-drag-icon': !isResizing }"
+                    aria-hidden="true"
```
MichalKinas: This should not be true always, there can be cases when this element is not hidden

lib/core/src/lib/datatable/components/datatable/datatable.component.html
```diff
@@ -134,14 +147,16 @@
             </div>
             <div
                 *ngIf="isResizingEnabled && col.resizable && !lastColumn"
-                [ngClass]="hoveredHeaderColumnIndex === columnIndex && !isResizing || resizingColumnIndex === columnIndex ? 'adf-datatable__resize-handle-visible' : 'adf-datatable__resize-handle-hidden'"
+                [ngClass]="hoveredHeaderColumnIndex === columnIndex && !isResizing || resizingColumnIndex === columnIndex ? 'adf-datatable_resize-handle-visible' : 'adf-datatable_resize-handle-hidden'"
```
MichalKinas: Why did you change the class name without changes in scss file? This will make the element lose some styles

lib/core/src/lib/datatable/components/datatable/datatable.component.html
```diff
@@ -184,7 +199,7 @@
     <div
         *ngIf="!loading; else loadingRowTemplate"
         class="adf-datatable-body"
-        [ngClass]="{ 'adf-blur-datatable-body': blurOnResize && (isDraggingHeaderColumn || isResizing), 'adf-datatable-body__draggable': enableDragRows && !isDraggingRow, 'adf-datatable-body__dragging': isDraggingRow }"
+        [ngClass]="{ 'adf-blur-datatable-body': blurOnResize && (isDraggingHeaderColumn || isResizing), 'adf-datatable-body_draggable': enableDragRows && !isDraggingRow, 'adf-datatable-body_dragging': isDraggingRow }"
```
MichalKinas: Same question here

lib/core/src/lib/datatable/components/datatable/datatable.component.html
```diff
@@ -236,28 +251,29 @@
                         </button>
                     </mat-menu>
                 </div>
-
-                <label *ngIf="multiselect"
-                        (keydown.enter)="onEnterKeyPressed(row, $any($event))"
-                        (click)="onCheckboxLabelClick(row, $event)"
```
MichalKinas, line 242: Why did you remove on click part? This will cause a bug

lib/core/src/lib/datatable/components/datatable/datatable.component.html
```diff
@@ -236,28 +251,29 @@
...
-                        [for]="'select-file-' + idx"
-                        class="adf-datatable-cell adf-datatable-checkbox adf-datatable-checkbox-single"
-                        tabindex="0">
+                <td
```
MichalKinas: This is only the checkbox part not the whole cell so I don't think this is a valid change

lib/core/src/lib/datatable/components/datatable/datatable.component.html
```diff
@@ -236,28 +251,29 @@
...
+                    class="adf-datatable-cell adf-datatable-checkbox adf-datatable-checkbox-single"
+                    tabindex="0"
+                    [attr.aria-label]="'ADF-DATATABLE.ACCESSIBILITY.SELECT_FILE' | translate"
+                >
                     <mat-checkbox
+                        *ngIf="multiselect"
```
MichalKinas: This condition should be set one level up I believe

MichalKinas (changes requested):

lib/core/src/lib/datatable/components/datatable/datatable.component.html
```diff
@@ -14,26 +14,30 @@
...
             [disabled]="!isHeaderVisible()"
             class="adf-datatable-row"
-            role="row">
-
+            role="row"> 
 
             <!-- Drag -->
             <div *ngIf="enableDragRows" class="adf-datatable-cell-header adf-drag-column">
-                <span class="adf-sr-only">{{ 'ADF-DATATABLE.ACCESSIBILITY.DRAG' | translate }}</span>
+                <div cdkDragHandle
```
MichalKinas: This is sr only so it means that this element is added purely for a screen reader not as actual element so it should not have the `cdkDragHandle` as you can see there is already the other element that has it applied

lib/core/src/lib/datatable/components/datatable/datatable.component.html
```diff
@@ -14,26 +14,30 @@
...
-            role="row">
-
+            role="row"> 
 
             <!-- Drag -->
             <div *ngIf="enableDragRows" class="adf-datatable-cell-header adf-drag-column">
-                <span class="adf-sr-only">{{ 'ADF-DATATABLE.ACCESSIBILITY.DRAG' | translate }}</span>
+                <div cdkDragHandle 
+                    class="adf-drag-handle"
+                    aria-hidden="true">
```
MichalKinas: Why are we adding this always set to true?

lib/core/src/lib/datatable/components/datatable/datatable.component.html
```diff
@@ -14,26 +14,30 @@
...
+                </div>
             </div>
 
             <!-- Actions (left) -->
-            <div *ngIf="actions && actionsPosition === 'left'" class="adf-actions-column adf-datatable-cell-header">
-                <span class="adf-sr-only">{{ 'ADF-DATATABLE.ACCESSIBILITY.ACTIONS' | translate }}</span>
+            <div *ngIf="actions && actionsPosition === 'left'"
+                class="adf-actions-column adf-datatable-cell-header">
+                    <span class="adf-sr-only">{{ 'ADF-DATATABLE.ACCESSIBILITY.ACTIONS' | translate }}</span>
```
MichalKinas: Indentation seems wrong here

lib/core/src/lib/datatable/components/datatable/datatable.component.html
```diff
@@ -141,7 +148,9 @@
                 (click)="$event.stopPropagation()"
                 (keyup.enter)="$event.stopPropagation()"
                 class="adf-datatable__resize-handle"
-                [resizableContainer]="resizableElement">
+                [resizableContainer]="resizableElement"
+                [attr.aria-label]="'ADF-DATATABLE.ACCESSIBILITY.RESIZE_COLUMN' | translate">
```
MichalKinas: Is this a valid added translation? I can't see it in localization files

lib/core/src/lib/datatable/components/datatable/datatable.component.scss
```diff
@@ -288,37 +288,79 @@ $data-table-cell-min-width-file-size: $data-table-cell-min-width-1 !default;
     /* stylelint-disable-next-line no-duplicate-selectors */
     .adf-datatable-cell,
     .adf-datatable-cell-header {
-        flex: 1;
+        flex: 1 1 0%;
```
MichalKinas: Why are we chaning styling for datatable if the ticket should only fix accessibility changes?

PR #10955, opened 2025-06-18
AAE-35201 File viewer toolbar has unnecessary separator with close button disabled

deepusingh1304:
https://hyland.atlassian.net/browse/AAE-35201

### DIFF
--- a/lib/core/src/lib/viewer/components/viewer.component.ts
+++ b/lib/core/src/lib/viewer/components/viewer.component.ts
@@ -39,7 +39,7 @@ import { MatButtonModule } from '@angular/material/button';
 import { MatDialog } from '@angular/material/dialog';
 import { MatIconModule } from '@angular/material/icon';
 import { MatMenuModule } from '@angular/material/menu';
-import { TranslateModule } from '@ngx-translate/core';
+import { TranslatePipe } from '@ngx-translate/core';
 import { fromEvent } from 'rxjs';
 import { filter, first, skipWhile } from 'rxjs/operators';
 import { AppConfigService } from '../../app-config';
@@ -79,7 +79,7 @@ const DEFAULT_NON_PREVIEW_CONFIG = {
         ToolbarComponent,
         ToolbarTitleComponent,
         MatButtonModule,
-        TranslateModule,
+        TranslatePipe,
         MatIconModule,
         MatMenuModule,
         ToolbarDividerComponent,
--- a/lib/content-services/package.json
+++ b/lib/content-services/package.json
@@ -22,7 +22,7 @@
     "@angular/platform-browser-dynamic": ">=14.1.3",
     "@angular/router": ">=14.1.3",
     "@alfresco/js-api": ">=9.0.0",
-    "@ngx-translate/core": ">=14.0.0",
+    "@ngx-translate/core": ">=16.0.0",
     "@alfresco/adf-core": ">=8.0.0"
   },
   "keywords": [
--- a/lib/content-services/src/lib/aspect-list/aspect-list-dialog.component.ts
+++ b/lib/content-services/src/lib/aspect-list/aspect-list-dialog.component.ts
@@ -18,7 +18,7 @@
 import { Component, Inject, OnInit, ViewEncapsulation } from '@angular/core';
 import { MAT_DIALOG_DATA, MatDialogModule, MatDialogRef } from '@angular/material/dialog';
 import { AspectListDialogComponentData } from './aspect-list-dialog-data.interface';
-import { TranslateModule } from '@ngx-translate/core';
+import { TranslatePipe } from '@ngx-translate/core';
 import { MatButtonModule } from '@angular/material/button';
 import { AspectListComponent } from './aspect-list.component';
 import { AutoFocusDirective } from '../directives/auto-focus.directive';
@@ -27,7 +27,7 @@ import { CommonModule } from '@angular/common';
 @Component({
     selector: 'adf-aspect-list-dialog',
     standalone: true,
-    imports: [CommonModule, MatDialogModule, TranslateModule, MatButtonModule, AspectListComponent, AutoFocusDirective],
+    imports: [CommonModule, MatDialogModule, TranslatePipe, MatButtonModule, AspectListComponent, AutoFocusDirective],
     templateUrl: './aspect-list-dialog.component.html',
     styleUrls: ['./aspect-list-dialog.component.scss'],
     encapsulation: ViewEncapsulation.None
--- a/lib/content-services/src/lib/aspect-list/aspect-list.component.ts
+++ b/lib/content-services/src/lib/aspect-list/aspect-list.component.ts
@@ -25,14 +25,14 @@ import { AspectEntry } from '@alfresco/js-api';
 import { CommonModule } from '@angular/common';
 import { MatExpansionModule } from '@angular/material/expansion';
 import { MatTableModule } from '@angular/material/table';
-import { TranslateModule } from '@ngx-translate/core';
+import { TranslatePipe } from '@ngx-translate/core';
 import { MatProgressSpinnerModule } from '@angular/material/progress-spinner';
 import { takeUntilDestroyed } from '@angular/core/rxjs-interop';
 
 @Component({
     selector: 'adf-aspect-list',
     standalone: true,
-    imports: [CommonModule, MatExpansionModule, MatCheckboxModule, MatTableModule, TranslateModule, MatProgressSpinnerModule],
+    imports: [CommonModule, MatExpansionModule, MatCheckboxModule, MatTableModule, TranslatePipe, MatProgressSpinnerModule],
     templateUrl: './aspect-list.component.html',
     styleUrls: ['./aspect-list.component.scss'],
     encapsulation: ViewEncapsulation.None
--- a/lib/content-services/src/lib/breadcrumb/breadcrumb.component.ts
+++ b/lib/content-services/src/lib/breadcrumb/breadcrumb.component.ts
@@ -21,13 +21,13 @@ import { Node, PathElement } from '@alfresco/js-api';
 import { DocumentListComponent } from '../document-list/components/document-list.component';
 import { CommonModule } from '@angular/common';
 import { MatIconModule } from '@angular/material/icon';
-import { TranslateModule } from '@ngx-translate/core';
+import { TranslatePipe } from '@ngx-translate/core';
 import { takeUntilDestroyed } from '@angular/core/rxjs-interop';
 
 @Component({
     selector: 'adf-breadcrumb',
     standalone: true,
-    imports: [CommonModule, MatIconModule, TranslateModule, MatSelectModule],
+    imports: [CommonModule, MatIconModule, TranslatePipe, MatSelectModule],
     templateUrl: './breadcrumb.component.html',
     styleUrls: ['./breadcrumb.component.scss'],
     encapsulation: ViewEncapsulation.None,
--- a/lib/content-services/src/lib/breadcrumb/dropdown-breadcrumb.component.ts
+++ b/lib/content-services/src/lib/breadcrumb/dropdown-breadcrumb.component.ts
@@ -21,12 +21,12 @@ import { PathElement, Node } from '@alfresco/js-api';
 import { BreadcrumbComponent } from './breadcrumb.component';
 import { CommonModule } from '@angular/common';
 import { MatIconModule } from '@angular/material/icon';
-import { TranslateModule } from '@ngx-translate/core';
+import { TranslatePipe } from '@ngx-translate/core';
 
 @Component({
     selector: 'adf-dropdown-breadcrumb',
     standalone: true,
-    imports: [CommonModule, MatIconModule, MatSelectModule, TranslateModule],
+    imports: [CommonModule, MatIconModule, MatSelectModule, TranslatePipe],
     templateUrl: './dropdown-breadcrumb.component.html',
     styleUrls: ['./dropdown-breadcrumb.component.scss'],
     encapsulation: ViewEncapsulation.None,
--- a/lib/content-services/src/lib/category/categories-management/categories-management.component.ts
+++ b/lib/content-services/src/lib/category/categories-management/categories-management.component.ts
@@ -35,7 +35,7 @@ import { debounce, first, map, tap } from 'rxjs/operators';
 import { CategoriesManagementMode } from './categories-management-mode';
 import { CategoryService } from '../services/category.service';
 import { CommonModule } from '@angular/common';
-import { TranslateModule } from '@ngx-translate/core';
+import { TranslatePipe } from '@ngx-translate/core';
 import { AutoFocusDirective } from '../../directives';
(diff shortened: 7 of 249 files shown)


lib/core/src/lib/viewer/components/viewer.component.html
```diff
@@ -135,11 +119,21 @@
...
+                    <img class="adf-viewer__mimeicon"
+                         [alt]="'ADF_VIEWER.ARIA.MIME_TYPE_ICON' | translate"
+                         [src]="mimeTypeIconUrl"
+                         data-automation-id="adf-file-thumbnail">
+                    <div class="adf-viewer__display-name"
+                         id="adf-viewer-display-name"
+                         [title]="fileName">
+                        <span>{{ displayName }}</span>
+                    </div>
+                </div>
```
fkindgen: Please move the navigation buttons to this element, too because there belonging to the file name

lib/core/src/lib/viewer/components/viewer.component.html
```diff
@@ -78,7 +66,6 @@
                     </mat-menu>
                 </ng-container>
 
-                <adf-toolbar-divider />
```
fkindgen: Where is the divider? Seems like it is missing now

lib/core/src/lib/viewer/components/viewer.component.spec.ts
```diff
@@ -695,4 +696,67 @@ describe('ViewerComponent', () => {
...
+                declarations: [ViewerComponent]
+            }).compileComponents();
+        });
+
+        beforeEach(() => {
+            fixture = TestBed.createComponent(ViewerComponent);
+            component = fixture.componentInstance;
+            component.showViewer = true;
+            fixture.detectChanges();
+        });
```
fkindgen: Is this still needed? I think it is already done before/above. Just compare it with the describes above.

lib/core/src/lib/viewer/components/viewer.component.spec.ts
```diff
@@ -695,4 +696,67 @@ describe('ViewerComponent', () => {
...
+
+        const createToolbarTest = (
+            allowRightSidebar: boolean,
+            hideInfoButton: boolean,
+            allowFullScreen: boolean,
+            allowGoBack: boolean,
+            expectedSeparatorCount: number
+        ) => {
+            it(`should have correct UI state when [R:${allowRightSidebar}, I:${hideInfoButton}, F:${allowFullScreen}, B:${allowGoBack}]`, () => {
+                component.allowRightSidebar = allowRightSidebar;
```
fkindgen: I think the left sidebar has effects on the rendering of the divider
```suggestion
                component.allowLeftSidebar = allowLeftSidebar;
```

lib/core/src/lib/viewer/components/viewer.component.spec.ts
```diff
@@ -695,4 +696,67 @@ describe('ViewerComponent', () => {
...
+                component.hideInfoButton = hideInfoButton;
+                component.allowFullScreen = allowFullScreen;
+                component.allowGoBack = allowGoBack;
```
fkindgen: I think this button can be rendered before the divider so I think we should include it in the tests
```suggestion
                component.allowGoBack = allowGoBack; mnuOpenWith
                component.mnuOpenWith = ...
```

lib/core/src/lib/viewer/components/viewer.component.spec.ts
```diff
@@ -695,4 +696,67 @@ describe('ViewerComponent', () => {
...
+                fixture.detectChanges();
+
+                const separators = fixture.debugElement.queryAll(
+                    // .adf-toolbar-divider is the class used for separators
+                    // If your divider uses a different selector, update here
```
fkindgen: Usually comments should be avoided also if there are helpful (that is a general rule here)

lib/core/src/lib/viewer/components/viewer.component.spec.ts
```diff
@@ -695,4 +696,67 @@ describe('ViewerComponent', () => {
...
+                    By.css('adf-toolbar-divider')
+                );
+                expect(separators.length).toBe(expectedSeparatorCount);
```
fkindgen: I think the divider will be visible inside the html always, because `display: none` will not remove the divider from the html DOM. Maybe check the `display: none` status for the first divider

lib/core/src/lib/viewer/components/viewer.component.html
```diff
@@ -31,35 +31,37 @@
...
-                    </button>
-                    <img class="adf-viewer__mimeicon"
-                         [alt]="'ADF_VIEWER.ARIA.MIME_TYPE_ICON' | translate"
-                         [src]="mimeTypeIconUrl"
-                         data-automation-id="adf-file-thumbnail">
-                    <div class="adf-viewer__display-name"
-                         id="adf-viewer-display-name"
-                         [title]="fileName">
-                        <span>{{ displayName }}</span>
+                    <div class="adf-viewer__file-title">
```
fkindgen: Do you reverted something? this element must be placed at the end of all children of `<adf-toolbar>`

lib/core/src/lib/viewer/components/viewer.component.spec.ts
```diff
@@ -18,8 +18,9 @@
 import { Component, SimpleChanges } from '@angular/core';
 import { ComponentFixture, discardPeriodicTasks, fakeAsync, flush, TestBed, tick } from '@angular/core/testing';
 import { MatButtonModule } from '@angular/material/button';
-import { MatDialog } from '@angular/material/dialog';
+import { MatDialog, MatDialogModule } from '@angular/material/dialog';
```
fkindgen: There is a TS error
```suggestion
import { MatDialog } from '@angular/material/dialog';
```

fkindgen (approved):

lib/core/src/lib/viewer/components/viewer.component.html
```diff
@@ -9,7 +9,7 @@
         <ng-content select="adf-viewer-toolbar" />
         <ng-container *ngIf="showToolbar && !toolbar">
             <adf-toolbar id="adf-viewer-toolbar" class="adf-viewer-toolbar">
-                <adf-toolbar-title>
```
MichalKinas: Why did you remove that?

lib/core/src/lib/viewer/components/viewer.component.html
```diff
@@ -31,35 +31,7 @@
                             (click)="onClose()">
                         <mat-icon>close</mat-icon>
                     </button>
-                </adf-toolbar-title>
 
-                <div class="adf-viewer__file-title">
```
MichalKinas: This looks like a breaking change, why did you move the title as the last element in the toolbar?

lib/core/src/lib/viewer/components/viewer.component.scss
```diff
@@ -35,6 +35,10 @@
     &-toolbar {
         #{ms.$mat-toolbar} {
             background-color: var(--adf-theme-background-card-color-087);
+
+            adf-toolbar-divider:first-child {
```
MichalKinas: This doesn't work for all cases for example it also hides the divider between info button and close button

lib/core/src/lib/viewer/components/viewer.component.spec.ts
```diff
@@ -722,4 +723,27 @@ describe('ViewerComponent', () => {
...
+        [false, false, true, false, null, 0, 'info button and fullscreen'],
+        [false, true, true, true, null, 1, 'close button and fullscreen'],
+        [false, true, false, true, null, 1, 'only close button'],
+        [false, true, false, false, {}, 0, 'only open with'],
+        [false, true, true, false, {}, 0, 'open with and fullscreen'],
+        [true, false, true, false, {}, 1, 'left sidebar, info, open with, fullscreen']
+    ].forEach(([allowLeftSidebar, hideInfoButton, allowFullScreen, allowGoBack, mnuOpenWith, expectedSeparatorCount, description]) => {
+        it(`should show ${expectedSeparatorCount} separator(s) when ${description}`, () => {
+            component.allowLeftSidebar = !!allowLeftSidebar;
+            component.hideInfoButton = !!hideInfoButton;
```
MichalKinas: Why are you adding `!!` if you're passing booleans?

lib/core/src/lib/viewer/components/viewer.component.spec.ts
```diff
@@ -722,4 +723,27 @@ describe('ViewerComponent', () => {
...
+            component.allowFullScreen = !!allowFullScreen;
+            component.allowGoBack = !!allowGoBack;
+            component.mnuOpenWith = mnuOpenWith;
+            fixture.detectChanges();
+
+            const separators = fixture.debugElement.queryAll(By.css('.adf-toolbar-divider'));
```
MichalKinas: For element queries you can use `UnitTestingUtils` helper class

MichalKinas (changes requested):

MichalKinas:
Please also make sure to fix issues reported by SonarCloud

lib/core/src/lib/viewer/components/viewer.component.spec.ts
```diff
@@ -722,4 +723,27 @@ describe('ViewerComponent', () => {
             expect(component.downloadFile.emit).toHaveBeenCalled();
         }));
     });
+    [
```
MichalKinas: This looks bad to be honest, you're sacrificing code readability in favour of couple lines of code, this should not be a case for unit tests

lib/core/src/lib/viewer/components/viewer.component.html
deepusingh1304: Removing <adf-toolbar-title> has no impact because it was always empty, it was rendered always before the toolbar divider  so its removal allows CSS rule of first-child to work on toolbar .

lib/core/src/lib/viewer/components/viewer.component.html
deepusingh1304: The "adf-viewer__file-title"  was inserted inside the adf-toolbar to group the navigation buttons, file icon, and file name together with the rest of the toolbar actions and the The adf-toolbar-divider became the first child of toolbar.

lib/core/src/lib/viewer/components/viewer.component.spec.ts
deepusingh1304: will Update the test cases.

lib/core/src/lib/viewer/components/viewer.component.spec.ts
deepusingh1304: Understood, Thank you for informing.

lib/core/src/lib/viewer/components/viewer.component.html
MichalKinas: What do you mean by `it was always empty`? It wasn't, the items placed there were content-projected to toolbar title and this component adds `adf-toolbar-title` class that is used in several places. This piece of code is used when sidebar should be placed on the left. Also removing the item from html only for the CSS rule to work usually is not a good idea.

lib/core/src/lib/viewer/components/viewer.component.html
MichalKinas: So you're saying that you moved it so that CSS rule would work? Again this is not the proper solution in the first place and additionally this is a breaking change since the viewer order will be broken after those changes

deepusingh1304: Understood, will revert back to avoid the breaking change

lib/core/src/lib/viewer/components/viewer.component.html
deepusingh1304: Understood, made the changes back.

lib/core/src/lib/viewer/components/viewer.component.html
```diff
@@ -78,12 +78,14 @@
                     </mat-menu>
                 </ng-container>
 
-                @if (showToolbarDividers) {
```
MichalKinas, line 81: You're deleting existing feature that allows to hide each dividers, this will break the upstream in other repos

lib/core/src/lib/viewer/components/viewer.component.html
```diff
@@ -78,12 +78,14 @@
...
-                    <adf-toolbar-divider />
-                }
+@if (shouldShowDivider) {
+    <adf-toolbar-divider data-automation-id="adf-toolbar-divider-middle">
```
MichalKinas: Do you this data automation id anywhere?

lib/core/src/lib/viewer/components/viewer.component.html
```diff
@@ -125,9 +127,9 @@
                 </ng-container>
 
                 <ng-container *ngIf="allowGoBack && closeButtonPosition === CloseButtonPosition.Right">
-                    @if (showToolbarDividers) {
```
MichalKinas: Why this condition was removed? This also break existing feature

MichalKinas (changes requested):

lib/core/src/lib/viewer/components/viewer.component.html
```diff
@@ -78,12 +78,13 @@
...
                 </ng-container>
 
-                @if (showToolbarDividers) {
+             @if (shouldShowDivider) {
                     <adf-toolbar-divider />
                 }
-
                 <ng-content select="adf-viewer-toolbar-custom-actions" />
 
+
```
MichalKinas: This empty line can be removed

lib/core/src/lib/viewer/components/viewer.component.html
```diff
@@ -78,12 +78,13 @@
                     </mat-menu>
                 </ng-container>
 
-                @if (showToolbarDividers) {
+             @if (shouldShowDivider) {
```
MichalKinas: Indentation seems wrong

lib/core/src/lib/viewer/components/viewer.component.html
```diff
@@ -78,12 +78,13 @@
...
                     <adf-toolbar-divider />
                 }
-
                 <ng-content select="adf-viewer-toolbar-custom-actions" />
 
+
+
```
MichalKinas: This one too

lib/core/src/lib/viewer/components/viewer.component.ts
```diff
@@ -559,4 +559,16 @@ export class ViewerComponent<T> implements OnDestroy, OnInit, OnChanges {
         this.urlFile = '';
         this.blobFile = null;
     }
+
+    get shouldShowDivider(): boolean {
+        if (!this.showToolbarDividers) {
+            return false;
+        }
+        const hasLeftSidebar = this.allowLeftSidebar;
```
MichalKinas: Why do you check left sidebar here and left close button position? Those are not the case for this divider as it is displayed after the title anyway, the proper conditions to display it or not would be `menu open with` or `adf-viewer-toolbar-actions` if one of the is present this divider should be shown

fkindgen:
Closed because it is not an issue any more.

PR #10970, opened 2025-06-27
AAE-36260 global test setup

DenysVuika:
single place to setup NoopAnimations and NoopTranslate, no more imports into each test file


AleksanderSklorz:
@DenysVuika there are still couple places where we import NoopAnimationsModule and NoopTranslateModule instead of using global settings. Can you change it too?

DenysVuika:
> @DenysVuika there are still couple places where we import NoopAnimationsModule and NoopTranslateModule instead of using global settings. Can you change it too?

I've updated some Core lib cases, note that Process Cloud is not enabled for now (due to some issues with the flakiness in the Form widgets), can be done in the future

PR #10985, opened 2025-07-03
[ACS-9327] Add prohibited symbols validation for Categories dialog

nikita-web-ua:

> - [X] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://hyland.atlassian.net/browse/ACS-9327

Added prohibited symbols validation for Categories dialog

> - [ ] Yes
> - [X] No

**Other information**:

### DIFF
--- a/lib/content-services/src/lib/category/categories-management/categories-management.component.spec.ts
+++ b/lib/content-services/src/lib/category/categories-management/categories-management.component.spec.ts
@@ -531,7 +531,7 @@ describe('CategoriesManagementComponent', () => {
                 component.categoryNameControl.markAsTouched();
                 fixture.detectChanges();
 
-                expect(getFirstError()).toBe('CATEGORIES_MANAGEMENT.ERRORS.SPECIAL_CHARACTERS');
+                expect(getFirstError()).toBe('CATEGORIES_MANAGEMENT.ERRORS.ENDS_WITH_DOT');
             }));
 
             it('should not display validation error when dot used in positions other than the end', fakeAsync(() => {
--- a/lib/content-services/src/lib/category/categories-management/categories-management.component.spec.ts
+++ b/lib/content-services/src/lib/category/categories-management/categories-management.component.spec.ts
@@ -525,6 +525,22 @@ describe('CategoriesManagementComponent', () => {
 
                 expect(getFirstError()).toBe('CATEGORIES_MANAGEMENT.ERRORS.SPECIAL_CHARACTERS');
             }));
+
+            it('should display validation error when dot placed in the end', fakeAsync(() => {
+                typeCategory('category.');
+                component.categoryNameControl.markAsTouched();
+                fixture.detectChanges();
+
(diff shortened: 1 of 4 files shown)


lib/content-services/src/lib/category/categories-management/categories-management.component.spec.ts
```diff
@@ -303,12 +303,6 @@ describe('CategoriesManagementComponent', () => {
             expect(component.categoryNameControl.hasValidator(Validators.required)).toBeFalse();
         });
 
-        it('should display validation error when searching for empty category', fakeAsync(() => {
```
nikita-web-ua, line 306: fyi, duplicated test removed

MichalKinas (approved):

lib/content-services/src/lib/category/categories-management/categories-management.component.ts
```diff
@@ -358,6 +360,11 @@ export class CategoriesManagementComponent implements OnInit, OnDestroy {
         return categoryNameControl.value.length && !categoryNameControl.value.trim() ? { emptyCategory: true } : null;
     }
 
+    private validateSpecialCharacters(categoryNameControl: FormControl<string>): CategoryNameControlErrors | null {
+        const specialSymbolsRegex = /[:"\\|<>/?*]/;
```
AleksanderSklorz, line 372: I think dot . is also disallowed character. Can you add that to list?

nikita-web-ua: Looks like it is not allowed only in the end
Screenshot 2025-07-07 at 08 43 19

AleksanderSklorz: so what do you think - should we have some separate validator which checks if it's not allowed at the end?

nikita-web-ua: added additional validator for this case

lib/content-services/src/lib/category/categories-management/categories-management.component.ts
```diff
@@ -358,6 +366,15 @@ export class CategoriesManagementComponent implements OnInit, OnDestroy {
...
+        return categoryNameControl.value.length && specialSymbolsRegex.test(categoryNameControl.value) ? { specialCharacters: true } : null;
+    }
+
+    private validateEndsWithDot(categoryNameControl: FormControl<string>): CategoryNameControlErrors | null {
+        return categoryNameControl.value.length && categoryNameControl.value.trim().endsWith('.') ? { specialCharacters: true } : null;
```
AleksanderSklorz: I think first condition is redundant because if that will be empty string '' then endsWith for dot will give false so you will return null. No need to check length of value

nikita-web-ua: right, fixed ty

PR #10995, opened 2025-07-07
[ADF-5582] Fix avatar rendering logic in adf-node-comments

Shivangi917:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [x] Documentation

The new adf-node-comments component does not display user avatar
The object sent to the getContentUrl service is neither a string or a NodeEntry.
A Person object is actually sent to the service resulting in the nodeId variable being undefined.

The new adf-node-comments component should display user avatar

> - [ ] Yes
> - [x] No

**Other information**:
https://hyland.atlassian.net/browse/ADF-5582

### DIFF
--- a/lib/content-services/src/lib/node-comments/services/node-comments.service.ts
+++ b/lib/content-services/src/lib/node-comments/services/node-comments.service.ts
@@ -86,6 +86,12 @@ export class NodeCommentsService implements CommentsService {
         });
     }
 
+    /**
+     * Gets the avatar image URL for a given user ID.
+     *
+     * @param userId ID of the user
+     * @returns The URL of the user's avatar image
+     */
     getUserImage(userId: string): string {
         return this.peopleApi.getAvatarImageUrl(userId);
     }
--- a/lib/js-api/src/api/content-rest-api/docs/PeopleApi.md
+++ b/lib/js-api/src/api/content-rest-api/docs/PeopleApi.md
@@ -12,7 +12,7 @@ All URIs are relative to *https://localhost/alfresco/api/-default-/public/alfres
 | [requestPasswordReset](#requestPasswordReset) | **POST** /people/{personId}/request-password-reset | Request password reset  |
 | [resetPassword](#resetPassword)               | **POST** /people/{personId}/reset-password         | Reset password          |
 | [updateAvatarImage](#updateAvatarImage)       | **PUT** /people/{personId}/avatar                  | Update avatar image     |
-| [getAvatarImageUrl](#getAvatarImageUrl)       | **(local)**                                        | Returns avatar image url|
+| [getAvatarImageUrl](#getAvatarImageUrl)       | **GET** /people/{personId}/avatar?placeholder=true | Returns avatar image url|
 | [updatePerson](#updatePerson)                 | **PUT** /people/{personId}                         | Update person           |
 
 ## createPerson


docs/core/components/comment-list.component.md
```diff
@@ -48,6 +48,49 @@ export class SomeComponent implements OnInit {
...
+import { CommentModel } from '@alfresco/adf-core';
+
+export class SomeComponent implements OnInit {
+  comments: CommentModel[] = [
+    {
+      id: 'user1',
+      firstName: 'Alice',
+      lastName: 'Smith',
+      pictureId: 1234
+    }
```
AleksanderSklorz: In that example code here is issue - missing comma, please add comma

docs/core/components/comment-list.component.md
```diff
@@ -48,6 +48,49 @@ export class SomeComponent implements OnInit {
...
+
+    {
+      id: 'user2',
+      firstName: 'Bob',
+      lastName: 'Jones',
+      avatarId: 'avatar-xyz'
+    }
```
AleksanderSklorz: In that example code here is issue - missing comma, please add comma

docs/core/components/comment-list.component.md
Shivangi917: Okay

docs/core/components/comment-list.component.md
Shivangi917: Okay

lib/core/src/lib/comments/comment-list/comment-list.component.html
```diff
@@ -3,10 +3,12 @@
                   (click)="selectComment(comment)"
                   class="adf-comment-list-item">
         <div class="adf-comment-img-container">
-            <div *ngIf="!comment.hasAvatarPicture" class="adf-comment-user-icon">{{ comment.userInitials }}</div>
-            <img *ngIf="comment.hasAvatarPicture" class="adf-people-img"
-                 [alt]="'COMMENTS.PROFILE_IMAGE' | translate"
-                 [src]="getUserImage(comment.createdBy.id.toString())" />
+            <div *ngIf="!getUserImage(comment.createdBy)" class="adf-comment-user-icon">
```
AleksanderSklorz: We are in Angular 19 now so we can use new control flow and variables in html to avoid calling getUserImage multiple times. Something like that: 
```html
@if (getUserImage(comment.createdBy); as userImage) {
  <img class="adf-people-img"
           [alt]="'COMMENTS.PROFILE_ALT' | translate:{ name: comment.userDisplayName }"
           [src]="userImage" />
} @else {
  <div class="adf-comment-user-icon">
    {{ comment.userInitials }}
  </div>
}
```

Thanks to else we don't need to repeat ngIf in both places. I added also "as userImage" in if to avoid calculating the same when assigning to src. 

See documentation: 
https://angular.dev/guide/templates/control-flow

https://angular.dev/guide/templates/variables

lib/core/src/lib/comments/comment-list/comment-list.component.spec.ts
```diff
@@ -163,4 +163,86 @@ describe('CommentListComponent', () => {
...
+                lastName: 'User',
+                pictureId: 1001
+            }
+        });
+
+        commentList.comments = [comment];
+        fixture.detectChanges();
+        await fixture.whenStable();
+
+        const img = testingUtils.getByCSS('.adf-people-img');
```
AleksanderSklorz: We are looking for that selector multiple times in that file so could you please wrap ```testingUtils.getByCSS('.adf-people-img')``` in separate function to avoid repeating selector?

lib/core/src/lib/comments/comment-list/comment-list.component.spec.ts
```diff
@@ -29,6 +29,15 @@ describe('CommentListComponent', () => {
     let fixture: ComponentFixture<CommentListComponent>;
     let testingUtils: UnitTestingUtils;
 
+    /**
+     * Helper to retrieve the avatar image element from the DOM.
```
AleksanderSklorz: No need for comment, we use comments only for public functions which can be used by customer

lib/core/src/lib/comments/comment-list/comment-list.component.spec.ts
```diff
@@ -29,6 +29,15 @@ describe('CommentListComponent', () => {
...
+     *
+     * @returns The avatar image element.
+     */
+    function getAvatarImg() {
```
AleksanderSklorz: please set return type

lib/core/src/lib/comments/comment-list/comment-list.component.spec.ts
```diff
@@ -29,6 +29,15 @@ describe('CommentListComponent', () => {
...
+        return testingUtils.getByCSS('.adf-people-img');
```
AleksanderSklorz: I see there is one more usage of that string in 162nd line. To avoid that you should put that string .adf-people-img into some const and then use that const here and also in 162nd line.

lib/core/src/lib/comments/comment-list/comment-list.component.spec.ts
Shivangi917: I am getting this error  Missing JSDoc comment

AleksanderSklorz: @Shivangi917 try to write it in arrow function format, then error should be gone: 
```ts
const getAvatarImg = () => testingUtils.getByCSS('.adf-people-img');

Comment (Shivangi917) [reply lib/core/src/lib/comments/comment-list/comment-list.component.spec.ts]: Okay Thanks

File: lib/core/src/lib/comments/comment-list/comment-list.component.spec.ts [hunk 8]
```diff
@@ -23,12 +23,17 @@ import { CommentListServiceMock } from './mocks/comment-list.service.mock';
...
 import { NoopTranslateModule } from '../../testing/noop-translate.module';
 import { UnitTestingUtils } from '../../testing/unit-testing-utils';
+import { DebugElement } from '@angular/core';
 
 describe('CommentListComponent', () => {
     let commentList: CommentListComponent;
     let fixture: ComponentFixture<CommentListComponent>;
     let testingUtils: UnitTestingUtils;
 
+    const AVATAR_IMG_SELECTOR = '.adf-people-img';
```
AleksanderSklorz: for const we use camelCase so should be named: 'avatarImgSelector'. We use uppercase only for readonly static fields.

AleksanderSklorz (approved):

lib/core/src/lib/comments/comment-list/comment-list.component.html
```diff
@@ -3,10 +3,16 @@
                   (click)="selectComment(comment)"
                   class="adf-comment-list-item">
         <div class="adf-comment-img-container">
-            <div *ngIf="!comment.hasAvatarPicture" class="adf-comment-user-icon">{{ comment.userInitials }}</div>
-            <img *ngIf="comment.hasAvatarPicture" class="adf-people-img"
-                 [alt]="'COMMENTS.PROFILE_IMAGE' | translate"
```
MichalKinas: Is the old translation key still used somewhere?

lib/core/src/lib/comments/comment-list/comment-list.component.html
```diff
@@ -3,10 +3,16 @@
...
-                 [src]="getUserImage(comment.createdBy.id.toString())" />
+            <ng-container *ngIf="getUserImage(comment.createdBy) as userImage; else noAvatar">
+                <img class="adf-people-img"
+                    [alt]="'COMMENTS.PROFILE_ALT' | translate:{ name: comment.userDisplayName }"
```
MichalKinas: I can't see this translation key added, why did you change it in the first place?

lib/core/src/lib/comments/comment-list/comment-list.component.html
```diff
@@ -3,10 +3,16 @@
...
+                    [src]="userImage" />
+            </ng-container>
+            <ng-template #noAvatar>
+                <div class="adf-people-img adf-people-img-default">
```
MichalKinas: Why did you change the css classes here? I can't see any styling changes

lib/core/src/lib/comments/comment-list/comment-list.component.ts
```diff
@@ -47,7 +48,14 @@ export class CommentListComponent {
         this.clickRow.emit(comment);
     }
 
-    getUserImage(userId: string): string {
-        return this.commentsService.getUserImage(userId);
+    getUserImage(user: User): string | null {
+        if (!user) {
```
MichalKinas: This is a breaking change since `commentsService` has different implementations based on the context such as `NodeCommentsService`, `CommentProcessService` etc, if you need to modify the behaviour it should be done in corresponding `getUserImage` method

lib/core/src/lib/comments/comment-list/comment-list.component.html
Shivangi917: No it is not used. Thank you for pointing it. I will remove it from i18n file

lib/core/src/lib/comments/comment-list/comment-list.component.html
Shivangi917: I had added COMMENTS.PROFILE_ALT but forgot to save the file before pushing. I updated the key to fix an accessibility issue flagged by SonarCloud.

lib/core/src/lib/comments/comment-list/comment-list.component.html
Shivangi917: I didn’t intend to change the CSS behavior. I added adf-people-img-default while refactoring but didn’t realize it’s not a defined style. I will fix it.

lib/core/src/lib/comments/comment-list/comment-list.component.ts
Shivangi917: I will fix this thing. Thank you

lib/content-services/src/lib/node-comments/services/node-comments.service.ts
AleksanderSklorz, line 1: missing unit tests for changes in that file

AleksanderSklorz:
Please fix sonar issue: 
image

lib/content-services/src/lib/node-comments/services/node-comments.service.spec.ts
```diff
@@ -81,5 +81,43 @@ describe('NodeCommentsService', () => {
...
+                createdBy: {
+                    id: 'test-user',
+                    firstName: 'Test',
+                    lastName: 'User',
+                    email: 'test@user.com',
+                    avatarId: 'avatar123'
+                }
+            };
+
+            const contentService = (service as any).contentService;
```
AleksanderSklorz: I know contentServices is private but we should not refer to private services like that. To refer to private service please use: 
TestBed.inject(ContentService)

This will give you reference to that service in your test

lib/content-services/src/lib/node-comments/services/node-comments.service.spec.ts
```diff
@@ -81,5 +81,43 @@ describe('NodeCommentsService', () => {
...
+            spyOn(contentService, 'getContentUrl').and.returnValue('fake-url');
+
+            const result = (service as any).newCommentModel(mockComment);
```
AleksanderSklorz: Similarly here - we should not use any to refer to private function. To execute private function you need to check in which public place that function is used and execute that thing instead. In that case I see that private newCommentModel function is used in public add function so use add function instead and thanks that it will call private function too and you don't need use of any then

lib/content-services/src/lib/node-comments/services/node-comments.service.spec.ts
```diff
@@ -81,5 +81,43 @@ describe('NodeCommentsService', () => {
...
+
+            expect(contentService.getContentUrl).toHaveBeenCalledWith('avatar123');
+            expect((service as any).avatarCache.get('test-user')).toBe('fake-url');
```
AleksanderSklorz: similarly like in previous comments - we shouldn't refer to private fields by using any. Instead you can notice that avatarCache is used in public getUserImage function too so you can check if it stores correct data by checking if result of getUserImage function is correct

lib/content-services/src/lib/node-comments/services/node-comments.service.spec.ts
```diff
@@ -81,5 +81,43 @@ describe('NodeCommentsService', () => {
                 responseText: JSON.stringify(fakeContentComments)
             });
         });
+
+        it('should cache the avatar URL when new comment is processed', () => {
+            const mockComment = {
```
AleksanderSklorz: please set correct type

lib/content-services/src/lib/node-comments/services/node-comments.service.spec.ts
```diff
@@ -81,5 +81,43 @@ describe('NodeCommentsService', () => {
...
+
+            const result = (service as any).newCommentModel(mockComment);
+
+            expect(contentService.getContentUrl).toHaveBeenCalledWith('avatar123');
+            expect((service as any).avatarCache.get('test-user')).toBe('fake-url');
+            expect(result.createdBy.id).toBe('test-user');
+        });
+
+        it('should return avatar URL from cache via getUserImage()', () => {
+            (service as any).avatarCache.set('cached-user', 'cached-url');
```
AleksanderSklorz: see above comments regarding to using any to read private things

lib/content-services/src/lib/node-comments/services/node-comments.service.spec.ts
Shivangi917: Done. Thank you

lib/content-services/src/lib/node-comments/services/node-comments.service.spec.ts
```diff
@@ -81,5 +84,44 @@ describe('NodeCommentsService', () => {
                 responseText: JSON.stringify(fakeContentComments)
             });
         });
+
+        it('should cache the avatar URL when new comment is processed', (done) => {
+            spyOn(contentService, 'getContentUrl').and.returnValue('fake-url');
+
+            service.add('999', fakeContentComment.entry.content).subscribe((result: CommentModel) => {
```
AleksanderSklorz: no need to set type for result here - it will be detected automatically based on left side of subscribe: 
image

Please remove that type

lib/content-services/src/lib/node-comments/services/node-comments.service.spec.ts
```diff
@@ -81,5 +84,44 @@ describe('NodeCommentsService', () => {
...
+                expect(contentService.getContentUrl).toHaveBeenCalledWith('123-123-123');
+                expect(service.getUserImage('fake-email@dom.com')).toBe('fake-url');
+                expect(result.createdBy.id).toBe('fake-email@dom.com');
+                done();
+            });
+
+            jasmine.Ajax.requests.mostRecent().respondWith({
```
AleksanderSklorz: If you want to mock out this.commentsApi.createComment then you should not use jasmine.Ajax for that. We should not touch anything outside file which we test so you should instead spyOn createComment function on commentsApi and then you can mock out result of that function

lib/content-services/src/lib/node-comments/services/node-comments.service.spec.ts
```diff
@@ -81,5 +84,44 @@ describe('NodeCommentsService', () => {
...
+        it('should return avatar URL from cache via getUserImage()', (done) => {
+            spyOn(contentService, 'getContentUrl').and.returnValue('cached-url');
+
+            service.add('999', fakeContentComment.entry.content).subscribe(() => {
+                const result = service.getUserImage('fake-email@dom.com');
+                expect(result).toBe('cached-url');
+                done();
+            });
+
+            jasmine.Ajax.requests.mostRecent().respondWith({
```
AleksanderSklorz: same comment like above for jasmine.Ajax

lib/content-services/src/lib/node-comments/services/node-comments.service.spec.ts
```diff
@@ -81,5 +84,44 @@ describe('NodeCommentsService', () => {
...
+                status: 200,
+                contentType: 'application/json',
+                responseText: JSON.stringify(fakeContentComment)
+            });
+        });
+
+        it('should return avatar URL from cache via getUserImage()', (done) => {
```
AleksanderSklorz: I have feeling that this test tests the same scenario like previous test. Is that right? Looking at titles both are testing scenario when avatar is cached. So if both test the same scenario can we combine both tests into single test to avoid repeating?

lib/content-services/src/lib/node-comments/services/node-comments.service.ts
```diff
@@ -73,15 +75,25 @@ export class NodeCommentsService implements CommentsService {
     }
 
     private newCommentModel(comment: Comment): CommentModel {
+        const user = new User(comment.createdBy);
+        const userId = user.id?.toString();
```
MichalKinas: Why are you adding `toString`? As far as I can see Person's id is a string already

lib/content-services/src/lib/node-comments/services/node-comments.service.ts
```diff
@@ -73,15 +75,25 @@ export class NodeCommentsService implements CommentsService {
...
+
+        const imageId = user.pictureId?.toString() || user.avatarId;
+
+        if (userId && imageId && !this.avatarCache.has(userId)) {
+            const avatarUrl = this.contentService.getContentUrl(imageId);
```
MichalKinas: Are you sure this endpoint will work with both `pictureId` and `avatarId`? I tested that for `hruser` and the request failed leaving the broken image

lib/content-services/src/lib/node-comments/services/node-comments.service.ts
```diff
@@ -73,15 +75,25 @@ export class NodeCommentsService implements CommentsService {
     }
 
     private newCommentModel(comment: Comment): CommentModel {
+        const user = new User(comment.createdBy);
+        const userId = user.id?.toString();
+
+        const imageId = user.pictureId?.toString() || user.avatarId;
```
MichalKinas: Why are we checking `pictureId` here? As far as I can see this is the activiti user property

MichalKinas (changes requested):

AleksanderSklorz:
@Shivangi917 I see sonarqube was not able to check your PR 
image

Could you please push empty commit 
git commit --allow-empty -m "Trigger sonar"

or try to close and reopen PR 

to restart sonarqube?

lib/content-services/src/lib/node-comments/services/node-comments.service.ts
```diff
@@ -66,22 +68,39 @@ export class NodeCommentsService implements CommentsService {
...
             message: comment.content,
             created: comment.createdAt,
-            createdBy: new User(comment.createdBy)
+            createdBy: user
         });
     }
 
-    getUserImage(avatarId: string): string {
-        return this.contentService.getContentUrl(avatarId);
+    getUserImage(userId: string): string {
```
AleksanderSklorz, line 95: Missing comment above that public function (it's required in ADF)

lib/content-services/src/lib/node-comments/services/node-comments.service.ts
```diff
@@ -28,6 +28,8 @@ import { AlfrescoApiService } from '../../services/alfresco-api.service';
 })
 export class NodeCommentsService implements CommentsService {
     private _commentsApi: CommentsApi;
+    private readonly avatarCache = new Map<string, string>();
```
AleksanderSklorz: please add empty line between private and private readonly fields (we should to group fields like public fields, private fields, private readonly fields etc.)

lib/content-services/src/lib/node-comments/services/node-comments.service.ts
```diff
@@ -66,22 +68,39 @@ export class NodeCommentsService implements CommentsService {
         return from(this.commentsApi.createComment(id, { content: message })).pipe(map((response) => this.newCommentModel(response.entry)));
     }
 
+    /**
+     * Gets the avatar cache containing userId–avatarUrl mappings.
+     *
+     * @returns A map of userId to avatar URL
+     */
+    getAvatarCache(): Map<string, string> {
```
AleksanderSklorz: Please remove that function completely. I see that function is used only in spec file (tests). We can't create something public only because of tests. Instead you need to follow whole flow: 
Function which you test uses avatarCache which is set in newCommentModel so to set caches in test you need to use something already public what uses newCommentModel function like for example add function

lib/content-services/src/lib/node-comments/services/node-comments.service.ts
MichalKinas: Still unfixed

lib/js-api/src/api/content-rest-api/docs/PeopleApi.md
```diff
@@ -2,17 +2,18 @@
...
+| [createPerson](#createPerson)                 | **POST** /people                                   | Create person           |
+| [deleteAvatarImage](#deleteAvatarImage)       | **DELETE** /people/{personId}/avatar               | Delete avatar image     |
+| [getAvatarImage](#getAvatarImage)             | **GET** /people/{personId}/avatar                  | Get avatar image        |
+| [getPerson](#getPerson)                       | **GET** /people/{personId}                         | Get a person            |
+| [listPeople](#listPeople)                     | **GET** /people                                    | List people             |
+| [requestPasswordReset](#requestPasswordReset) | **POST** /people/{personId}/request-password-reset | Request password reset  |
+| [resetPassword](#resetPassword)               | **POST** /people/{personId}/reset-password         | Reset password          |
+| [updateAvatarImage](#updateAvatarImage)       | **PUT** /people/{personId}/avatar                  | Update avatar image     |
+| [getAvatarImageUrl](#getAvatarImageUrl)       | **(local)**                                        | Returns avatar image url|
```
MichalKinas: request address should be fixed

MichalKinas:
Please also rebase the branch to get latest changes

MichalKinas (approved):

PR #11000, opened 2025-07-09
AAE-36368 Multiple styles issues for metadata sidebar

Isoumyajit:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

> - [ ] Yes
> - [x] No

**Other information**:
https://hyland.atlassian.net/browse/AAE-36368

### DIFF
--- a/lib/core/src/lib/card-view/components/card-view-selectitem/card-view-selectitem.component.html
+++ b/lib/core/src/lib/card-view/components/card-view-selectitem/card-view-selectitem.component.html
@@ -17,7 +17,7 @@
                     {{ property.displayValue | async | translate }}
                 </div>
                 <div *ngIf="isEditable">
-                    <mat-form-field class="adf-property-value">
+                    <mat-form-field class="adf-property-value adf-form-field-input">
                         <mat-label
                             [attr.data-automation-id]="'card-select-label-' + property.key"
                             class="adf-property-label adf-property-value-editable"
@@ -46,7 +46,7 @@
         </div>
         <div *ngSwitchCase="'autocompleteBased'">
             <mat-form-field
-                class="adf-property-field adf-card-selectitem-autocomplete"
+                class="adf-property-field adf-card-selectitem-autocomplete adf-form-field-input"
                 [ngClass]="{ 'adf-property-read-only': !isEditable }"
                 [floatLabel]="property.default ? 'always' : null"
             >
--- a/lib/core/src/lib/card-view/components/card-view-selectitem/card-view-selectitem.component.html
+++ b/lib/core/src/lib/card-view/components/card-view-selectitem/card-view-selectitem.component.html
@@ -17,7 +17,7 @@
                     {{ property.displayValue | async | translate }}
                 </div>
                 <div *ngIf="isEditable">
-                    <mat-form-field class="adf-property-value adf-form-field-input">
+                    <mat-form-field class="adf-property-value">
                         <mat-label
                             [attr.data-automation-id]="'card-select-label-' + property.key"
                             class="adf-property-label adf-property-value-editable"
@@ -46,7 +46,7 @@
         </div>
         <div *ngSwitchCase="'autocompleteBased'">
             <mat-form-field
-                class="adf-property-field adf-card-selectitem-autocomplete adf-form-field-input"
+                class="adf-property-field adf-card-selectitem-autocomplete   "
                 [ngClass]="{ 'adf-property-read-only': !isEditable }"
                 [floatLabel]="property.default ? 'always' : null"
             >
--- a/lib/core/src/lib/card-view/components/card-view-selectitem/card-view-selectitem.component.scss
+++ b/lib/core/src/lib/card-view/components/card-view-selectitem/card-view-selectitem.component.scss
@@ -71,9 +71,4 @@
             font-size: 10px;
         }
     }
-
-    #{ms.$mat-form-field-infix} {
-        padding-left: 3%;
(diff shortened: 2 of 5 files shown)


lib/core/src/lib/card-view/components/card-view-selectitem/card-view-selectitem.component.html
```diff
@@ -17,7 +17,7 @@
                     {{ property.displayValue | async | translate }}
                 </div>
                 <div *ngIf="isEditable">
-                    <mat-form-field class="adf-property-value">
+                    <mat-form-field class="adf-property-value adf-form-field-input">
```
VitoAlbano: Why start appending styles ? maybe it would be better have a specific styles that we want to apply on this element to avoid side effects

lib/core/src/lib/card-view/components/card-view-selectitem/card-view-selectitem.component.scss
```diff
@@ -14,7 +14,7 @@
         z-index: 1;
     }
 
-    .adf-property-value-editable.adf-property-value-editable {
+    .adf-property-value-editable {
```
VitoAlbano, line 17: was this redundant?

lib/core/src/lib/card-view/components/card-view-selectitem/card-view-selectitem.component.scss
```diff
@@ -72,4 +71,9 @@
             font-size: 10px;
         }
     }
+
+    #{ms.$mat-form-field-infix} {
```
VitoAlbano: isn't this targeting too many elements possibly? plus in percentage this padding might start to take too much space

VitoAlbano (changes requested):

VitoAlbano:
Even though no checks have been triggered and it's mergable please avoid until the comments have been solved.

lib/core/src/lib/card-view/components/card-view-selectitem/card-view-selectitem.component.html
Isoumyajit: Every form fields are styled from a single place for studio-hxp so the styles are written in a file that is present inside shared in monorepo so when it compiles every adf form fields which are having the classes and all of the form-fields should have the styles will be applied to them which is applied to other hxp form-fields which are not part of ADF. @VitoAlbano .  

You can check this file `libs/shared/core/src/styles/_hxp-inputs.scss`

lib/core/src/lib/card-view/components/card-view-selectitem/card-view-selectitem.component.scss
Isoumyajit: Yes

lib/core/src/lib/card-view/components/card-view-selectitem/card-view-selectitem.component.scss
Isoumyajit: No actually if you see the class is present inside `.adf-card-view-selectitem` so yes the styles will be applied to all the card-view-select items and only on the mat-form-fields inside the card-view select type components. which is actually expected and it breaks initially because we earlier didn't use mat-form-field appearance we placed labels outside form-field which is not a correct way as per Material.

lib/core/src/lib/card-view/components/card-view-selectitem/card-view-selectitem.component.html
VitoAlbano: This doesn't answer my question, and the issue here i'm trying to avoid is : Piling up multiple styles that can fight each other

lib/core/src/lib/card-view/components/card-view-selectitem/card-view-selectitem.component.scss
VitoAlbano: Ok but why we are changing material styles now if we are using the standard label? we should avoid if it's possible to add more tampering with material as those styles are private and not meant to be changed

Isoumyajit: If we place the mat-labels outside the mat-form-fields it will be fine for apa but in hxp we can't have that because the UI design proposed by the UX team - `https://www.figma.com/board/5qmiAirnoMXo6yUqP1feH6/Automate-Workspace?t=3vugbR89wEw8qRGS-0` must have the mat-labels inside the mat-form-field to achieve the 
looks as they mentioned on the document there is a need to override the styles using material classes otherwise apa will look like this as the labels now part of the mat-form-field
image

and the look we want in hxp is 
image

now the label-float behaviour is auto handled by the materials and there will not be any issues if we do not have a background color for apa and content-ee-apa and here we do not want any background color.

so to handle this kind of issues to maintain two different styles in two different applications using same shared component I added some minor tweaks.

lib/core/src/lib/card-view/components/card-view-selectitem/card-view-selectitem.component.html
Isoumyajit: I added that class `adf-form-field-input` because these components are being used by hxp as well and hxp demands different styles than apa so if we do not add this class there will be no way by which we can override the styles for hxp. and if we add the styles directly inside ADF then it will affect all the other applications like ACA , APA , Content-ee-apa and others. so I added this but if you have any other idea please do let me know I can try that approach.

VitoAlbano (approved):

PR #11005, opened 2025-07-10
[ADF-5583] Update style to make version visible outside dialog

Shivangi917:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

When the adf-version-list component is used outside of a dialog, the version list is not visible.

The adf-version-list component should automatically display and scroll version entries, whether used inside a dialog or as a standalone component, without requiring consumers to manually apply height styles. The component should manage its layout internally using proper flex and scroll properties.

> - [ ] Yes
> - [x] No

**Other information**:
https://hyland.atlassian.net/browse/ADF-5583

### DIFF
--- a/lib/content-services/src/lib/version-manager/version-list.component.scss
+++ b/lib/content-services/src/lib/version-manager/version-list.component.scss
@@ -1,7 +1,6 @@
 .adf-version-list {
     &-viewport {
         display: flex;
-        flex-direction: column;
         min-height: 200px;
         max-height: 60vh;
         height: 100%;
--- a/lib/content-services/src/lib/version-manager/version-list.component.scss
+++ b/lib/content-services/src/lib/version-manager/version-list.component.scss
@@ -2,11 +2,9 @@
     &-viewport {
         display: flex;
         flex-direction: column;
-        flex: 1 1 auto;
         min-height: 200px;
         max-height: 60vh;
         height: 100%;
-        overflow: hidden auto;
 
         :first-child {
             max-width: 100%;
--- a/lib/content-services/src/lib/version-manager/version-list.component.scss
+++ b/lib/content-services/src/lib/version-manager/version-list.component.scss
@@ -1,8 +1,12 @@
 .adf-version-list {
     &-viewport {
-        height: 100%;
-        overflow-x: hidden;
         display: flex;
+        flex-direction: column;
+        flex: 1 1 auto;
+        min-height: 200px;
+        max-height: 60vh;
+        height: 100%;
+        overflow: hidden auto;
 
         :first-child {
             max-width: 100%;


lib/content-services/src/lib/version-manager/version-list.component.scss
```diff
@@ -1,8 +1,12 @@
 .adf-version-list {
     &-viewport {
-        height: 100%;
-        overflow-x: hidden;
         display: flex;
+        flex-direction: column;
```
AleksanderSklorz: Do we need that property? I can't see difference when I remove it so seems redundant

lib/content-services/src/lib/version-manager/version-list.component.scss
```diff
@@ -1,8 +1,12 @@
...
+        flex: 1 1 auto;
```
AleksanderSklorz: Do we need that property? I can't see difference when I remove it so seems redundant

lib/content-services/src/lib/version-manager/version-list.component.scss
```diff
@@ -1,8 +1,12 @@
...
+        min-height: 200px;
+        max-height: 60vh;
```
AleksanderSklorz, line 5: can we use px instead of vh?

lib/content-services/src/lib/version-manager/version-list.component.scss
```diff
@@ -1,8 +1,12 @@
...
+        height: 100%;
+        overflow: hidden auto;
```
AleksanderSklorz: Do we need that property? I can't see difference when I remove it so seems redundant

lib/content-services/src/lib/version-manager/version-list.component.scss
Shivangi917: I added flex-direction: column to ensure version items stack vertically, and flex: 1 1 auto so the container can grow and shrink properly inside flexible layouts. It helps maintain consistent scroll behavior, especially outside dialogs. If you want should i remove it?

lib/content-services/src/lib/version-manager/version-list.component.scss
Shivangi917: I used vh as it scales with screen size (60vh means 60% of the screen height).
px will be less flexible as a 300px list may look fine on desktop but too tall or too short on mobile.
So should i change it or keep it

lib/content-services/src/lib/version-manager/version-list.component.scss
Shivangi917: the difference might not be visible in the dialog, but outside it, min-height and max-height ensure the list doesn't collapse or overflow. overflow: hidden auto enables scrolling only when needed.

lib/content-services/src/lib/version-manager/version-list.component.scss
AleksanderSklorz: ok you can leave it

lib/content-services/src/lib/version-manager/version-list.component.scss
AleksanderSklorz: ok you can leave it for now if that works

lib/content-services/src/lib/version-manager/version-list.component.scss
AleksanderSklorz: I see it's unused value: 
image

you can see in browser that this property is crossed out and if you look a bit higher you can see that it's overridden by .cdk-virtual-scrollable class which also has that style. So I believe we can remove that property as it always does nothing

Shivangi917: Sure. I will check this thing and will remove it.

lib/content-services/src/lib/version-manager/version-list.component.scss
Shivangi917: Yes, I have removed it.

PR #11051, opened 2025-07-23
AAE-0000 It was not possible to install directly due to broken dep

VitoAlbano:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

> - [ ] Yes
> - [ ] No

**Other information**:

### DIFF
--- a/package.json
+++ b/package.json
@@ -173,6 +173,9 @@
   "engines": {
     "node": ">=18.0.0"
   },
+    "overrides": {
+    "stylus": "github:stylus/stylus#0.64.0"
+  },
   "module": "./index.js",
   "typings": "./index.d.ts"
 }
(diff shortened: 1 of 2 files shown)


wojd0 (approved):

package.json
```diff
@@ -173,6 +173,9 @@
   "engines": {
     "node": ">=18.0.0"
   },
+  "overrides": {
```
AleksanderSklorz: This solution is fine for me but I have just one question - instead of overriding that - did you check if upgrading dependencies' versions which use stylus won't resolve that issue too? It would be better. But if you check and it wasn't working then it's fine for me

VitoAlbano: @AleksanderSklorz  i did, unfortunately @nx/webpack 21 is still using stylus 0.64 so until they put the package back up on npm or unless @nx doesn't change the deps we need to handle like this as this is the easiest approach

PR #11078, opened 2025-07-29
AAE-36869 Handles error messages with wrong JSON format

APuschkin:
Handles cases where the error message from the backend is not a valid JSON.

If the message is not a valid JSON, it displays the error as a string.
This prevents the application from crashing when receiving error messages with wrong JSON format.

Fixes AAE-36869

### DIFF
--- a/lib/process-services-cloud/src/lib/process/start-process/components/start-process-cloud.component.html
+++ b/lib/process-services-cloud/src/lib/process/start-process/components/start-process-cloud.component.html
@@ -73,7 +73,10 @@
                                 }
                                 @if (processInstanceName.hasError('maxlength')) {
                                     <span id="adf-start-process-maxlength-error">
-                                        {{ 'ADF_CLOUD_PROCESS_LIST.ADF_CLOUD_START_PROCESS.ERROR.MAXIMUM_LENGTH' | translate : { characters: maxNameLength } }}
+                                        {{
+                                            'ADF_CLOUD_PROCESS_LIST.ADF_CLOUD_START_PROCESS.ERROR.MAXIMUM_LENGTH'
+                                                | translate: { characters: maxNameLength }
+                                        }}
                                     </span>
                                 }
                                 @if (processInstanceName.hasError('pattern')) {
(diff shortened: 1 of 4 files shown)


lib/process-services-cloud/src/lib/process/start-process/components/start-process-cloud.component.html
```diff
@@ -76,7 +70,7 @@
                 </adf-inplace-form-input>
             </form>
 
-            <ng-container *ngIf="hasForm else taskFormCloudButtons">
+            <ng-container *ngIf="hasForm; else taskFormCloudButtons">
```
DenysVuika: I suggest using https://angular.dev/guide/templates/control-flow as we have access to that already

```
@if (hasForm) {
 // form content
} @else {
 // html here, no need for extra tempalte
}
```

lib/process-services-cloud/src/lib/process/start-process/components/start-process-cloud.component.ts
```diff
@@ -456,8 +456,8 @@ export class StartProcessCloudComponent implements OnChanges, OnInit {
                 this.isProcessStarting = false;
             },
             error: (err) => {
-                this.errorMessageId = 'ADF_CLOUD_PROCESS_LIST.ADF_CLOUD_START_PROCESS.ERROR.START';
-                this.unifyErrorResponse(err);
+                this.errorMessageId = err?.response?.body?.message || 'ADF_CLOUD_PROCESS_LIST.ADF_CLOUD_START_PROCESS.ERROR.START_PROCESS';
+                this.unifyErrorResponse(err?.response?.body);
```
JWollweber: I'm a bit wondering.. what does the error actually look like? You are accessing `.response.body` here already and inside `unifyErrorResponse` you are accessing `.response.body.entry` / `.response.body.message` again... is the overall structure really this nested (`err.response.body.response.body.entry` / `err.response.body.response.body.message`)? Or was your intention to access `err.response.body` before calling the method and removing this nesting from the method but you forgot to do so?

APuschkin: Thats funny. I checked it again and saw that this whole implementation of "unifyErrorResponse" is not needed, because the messages and the error definitions are already in the errors implemented. So the entry which we created there is redundant

APuschkin: After change, thats the error json which came from ng2 to the front-end-apps
image

PR #11080, opened 2025-07-30
AAE-36536 fix for root section and mobile form input layout - ng18-backup

dthornton-hyl:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/core/src/lib/form/components/form-section/form-section.component.ts
+++ b/lib/core/src/lib/form/components/form-section/form-section.component.ts
@@ -44,6 +44,10 @@ export class FormSectionComponent implements OnInit {
         const defaultColspan = 1;
         const fieldColspan = columnFields[firstColumnFieldIndex]?.colspan ?? defaultColspan;
 
-        return (100 / numberOfColumns) * fieldColspan + '';
+        if (typeof numberOfColumns !== 'number' || !numberOfColumns || numberOfColumns <= 0) {
+            return Math.min(100, 100 * fieldColspan) + '';
+        }
+
+        return Math.min(100, (100 / numberOfColumns) * fieldColspan) + '';
     }
 }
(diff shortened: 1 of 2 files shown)


amolodyh-hyland:
@dthornton-hyl Missed the SCSS change. Do we not need the `flex-direction: column` in this PR?

amolodyh-hyland:
@dthornton-hyl Ignore my above comment. I see that change is already in the SCSS file on this branch. All good.

lib/core/src/lib/form/components/form-section/form-section.component.ts
```diff
@@ -44,6 +44,10 @@ export class FormSectionComponent implements OnInit {
         const defaultColspan = 1;
         const fieldColspan = columnFields[firstColumnFieldIndex]?.colspan ?? defaultColspan;
 
-        return (100 / numberOfColumns) * fieldColspan + '';
+        if (typeof numberOfColumns !== 'number' || !numberOfColumns || numberOfColumns <= 0) {
+            return Math.min(100, 100 * fieldColspan) + '';
```
sdonnell-hyland: Wouldn't this always return '100' assuming fieldColspan will always be an integer of 1 or greater than 1?  Or is this protection in case somehow it gets set to < 1?

dthornton-hyl: @sdonnell-hyland @amolodyh-hyland 

Good point, maybe change to below? That way keep the protection. It is coming in as undefined. I will take a look and see why it is coming in undefined.

if (typeof numberOfColumns !== 'number' || !numberOfColumns || numberOfColumns <= 0) {
    numberOfColumns = 1;
}

dthornton-hyl: I went ahead and made that change. It was coming undefined because for root section it will be wrapped in container object, so numbercolumns wouldn't be set so this check will address that. Throught testing everything is looking as expected.

PR #11171, opened 2025-09-05
[MNT-24388] fix location search facet

g-jaskowski:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://hyland.atlassian.net/browse/MNT-24388

Existing sites and libraries are included in Location search facet

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/content-services/src/lib/search/components/search-filter-autocomplete-chips/search-filter-autocomplete-chips.component.ts
+++ b/lib/content-services/src/lib/search/components/search-filter-autocomplete-chips/search-filter-autocomplete-chips.component.ts
@@ -122,11 +122,12 @@ export class SearchFilterAutocompleteChipsComponent implements SearchWidget, OnI
     }
 
     onInputChange(value: string) {
-        if (this.settings.field === AutocompleteField.CATEGORIES && value) {
-            this.searchForExistingCategories(value);
-        }
-        if (this.settings.field === AutocompleteField.LOCATION && value) {
-            this.populateSitesOptions();
+        if (value) {
+            if (this.settings.field === AutocompleteField.CATEGORIES) {
+                this.searchForExistingCategories(value);
+            } else if (this.settings.field === AutocompleteField.LOCATION) {
+                this.populateSitesOptions();
+            }
         }
     }
 
@@ -197,7 +198,7 @@ export class SearchFilterAutocompleteChipsComponent implements SearchWidget, OnI
             .pipe(
                 map((sites) => {
                     const predefinedOptions = this.settings?.autocompleteOptions || [];
-                    const sitesOptions = sites.list.entries.map((siteEntry) => ({
+                    const sitesOptions = sites.list.entries.map<AutocompleteOption>((siteEntry) => ({
                         id: siteEntry.entry.id,
                         value: siteEntry.entry.title
                     }));
--- a/lib/content-services/src/lib/search/components/search-filter-autocomplete-chips/search-filter-autocomplete-chips.component.ts
+++ b/lib/content-services/src/lib/search/components/search-filter-autocomplete-chips/search-filter-autocomplete-chips.component.ts
@@ -56,9 +56,9 @@ export class SearchFilterAutocompleteChipsComponent implements SearchWidget, OnI
     private readonly destroyRef = inject(DestroyRef);
 
     constructor(
-        private tagService: TagService,
-        private categoryService: CategoryService,
-        private sitesService: SitesService
+        private readonly tagService: TagService,
+        private readonly categoryService: CategoryService,
+        private readonly sitesService: SitesService
     ) {
         this.options = new SearchFilterList<AutocompleteOption[]>();
     }
--- a/lib/content-services/src/lib/search/components/search-filter-autocomplete-chips/search-filter-autocomplete-chips.component.ts
+++ b/lib/content-services/src/lib/search/components/search-filter-autocomplete-chips/search-filter-autocomplete-chips.component.ts
@@ -30,6 +30,7 @@ import { SearchChipAutocompleteInputComponent } from '../search-chip-autocomplet
 import { TranslatePipe } from '@ngx-translate/core';
 import { MatButtonModule } from '@angular/material/button';
 import { takeUntilDestroyed } from '@angular/core/rxjs-interop';
+import { SitesService } from '../../../common/services/sites.service';
 
 @Component({
     selector: 'adf-search-filter-autocomplete-chips',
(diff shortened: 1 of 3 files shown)


AleksanderSklorz:
@g-jaskowski I found issue - when you have public library but you are not member of that (for example another user created that library) then you can't search files inside of that. But if you navigate to that library manually you can find some file there: 

https://github.com/user-attachments/assets/a285a54b-2b00-4ad5-b1ce-104e20d189bc

AleksanderSklorz:
@g-jaskowski I found interesting case. Can you ask somebody to decide what is correct behavior? See following scenario: 
If you are not member of library because for example library was created by different user then you can see that library in ADW in All Libraries section (if library is visible for your user). Then you can see that library also in Location in search. 

But what if user will log in ACA? In ACA there is no All Libraries section so user is not even aware about libraries from other users but that library can be still visible in Location in search: 

https://github.com/user-attachments/assets/b47b0eda-b8aa-4ec1-a144-83c8fd855b78

Is that expected or not?

lib/content-services/src/lib/search/components/search-filter-autocomplete-chips/search-filter-autocomplete-chips.component.ts
```diff
@@ -120,6 +125,9 @@ export class SearchFilterAutocompleteChipsComponent implements SearchWidget, OnI
         if (this.settings.field === AutocompleteField.CATEGORIES && value) {
             this.searchForExistingCategories(value);
         }
+        if (this.settings.field === AutocompleteField.LOCATION && value) {
```
AleksanderSklorz: value condition is repeated in above if. Can we split those conditions to don't repeat them? Something like: 
```ts
if (value) {
  if (this.settings.field === AutocompleteField.CATEGORIES) {
     ...
  } else if (this.settings.field === AutocompleteField.LOCATION) {
     ...
  } 
}
```

I have added outer if to avoid repeating. I also added 'else' because field can't be LOCATION and CATEGORIES in the same time. It can be only one of them so else will be better to avoid checking both

lib/content-services/src/lib/search/components/search-filter-autocomplete-chips/search-filter-autocomplete-chips.component.ts
```diff
@@ -173,4 +190,20 @@ export class SearchFilterAutocompleteChipsComponent implements SearchWidget, OnI
...
         });
     }
+
+    private populateSitesOptions(): void {
+        this.sitesService
+            .getSites()
+            .pipe(
+                map((sites) => {
+                    const predefinedOptions = this.settings?.autocompleteOptions || [];
+                    const sitesOptions = sites.list.entries.map((siteEntry) => ({
```
AleksanderSklorz: that map breaks typings: 
image
It sets type to 
```
{id: string; value: string}[]
```

Please set type manually: 
```ts
const sitesOptions = sites.list.entries.map<AutocompleteOption>((siteEntry) => ({
  id: siteEntry.entry.id,
  value: siteEntry.entry.title
}));

g-jaskowski:
@AleksanderSklorz thank you for the review, I've applied your comments.

> @g-jaskowski I found issue - when you have public library but you are not member of that (for example another user created that library) then you can't search files inside of that. But if you navigate to that library manually you can find some file there:
> 
>  Screen.Recording.2025-09-05.at.14.05.09.mov

I'm unable to replicate this issue, could you please check again?



> @g-jaskowski I found interesting case. Can you ask somebody to decide what is correct behavior? See following scenario: If you are not member of library because for example library was created by different user then you can see that library in ADW in All Libraries section (if library is visible for your user). Then you can see that library also in Location in search.
> 
> But what if user will log in ACA? In ACA there is no All Libraries section so user is not even aware about libraries from other users but that library can be still visible in Location in search:
> 
>  Screen.Recording.2025-09-05.at.14.21.37.mov 
> Is that expected or not?

I've contacted Arabella and discussed this finding. I'll be creating a separate ticket to move All Libraries page to ACA to prevent the inconsistencies in search.

AleksanderSklorz:
> @AleksanderSklorz thank you for the review, I've applied your comments.
> 
> > @g-jaskowski I found issue - when you have public library but you are not member of that (for example another user created that library) then you can't search files inside of that. But if you navigate to that library manually you can find some file there:
> > Screen.Recording.2025-09-05.at.14.05.09.mov
> 
> I'm unable to replicate this issue, could you please check again?
> 
> > @g-jaskowski I found interesting case. Can you ask somebody to decide what is correct behavior? See following scenario: If you are not member of library because for example library was created by different user then you can see that library in ADW in All Libraries section (if library is visible for your user). Then you can see that library also in Location in search.
> > But what if user will log in ACA? In ACA there is no All Libraries section so user is not even aware about libraries from other users but that library can be still visible in Location in search:
> > Screen.Recording.2025-09-05.at.14.21.37.mov
> > Is that expected or not?
> 
> I've contacted Arabella and discussed this finding. I'll be creating a separate ticket to move All Libraries page to ACA to prevent the inconsistencies in search.

@g-jaskowski regarding to first issue - I can't reproduce it as well now, so seems like fixed. 

Regarding to the second - ok, can you share link to that task which you are creating here?

g-jaskowski:
> > @AleksanderSklorz thank you for the review, I've applied your comments.
> > > @g-jaskowski I found issue - when you have public library but you are not member of that (for example another user created that library) then you can't search files inside of that. But if you navigate to that library manually you can find some file there:
> > > Screen.Recording.2025-09-05.at.14.05.09.mov
> > 
> > 
> > I'm unable to replicate this issue, could you please check again?
> > > @g-jaskowski I found interesting case. Can you ask somebody to decide what is correct behavior? See following scenario: If you are not member of library because for example library was created by different user then you can see that library in ADW in All Libraries section (if library is visible for your user). Then you can see that library also in Location in search.
> > > But what if user will log in ACA? In ACA there is no All Libraries section so user is not even aware about libraries from other users but that library can be still visible in Location in search:
> > > Screen.Recording.2025-09-05.at.14.21.37.mov
> > > Is that expected or not?
> > 
> > 
> > I've contacted Arabella and discussed this finding. I'll be creating a separate ticket to move All Libraries page to ACA to prevent the inconsistencies in search.
> 
> @g-jaskowski regarding to first issue - I can't reproduce it as well now, so seems like fixed.
> 
> Regarding to the second - ok, can you share link to that task which you are creating here?

Sure, the task I've created is [ACS-10165](https://hyland.atlassian.net/browse/ACS-10165) @AleksanderSklorz 

[ACS-10165]: https://hyland.atlassian.net/browse/ACS-10165?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

AleksanderSklorz (approved):

PR #11187, opened 2025-09-10
AAE-30878 Migrating from event-emitter to eventemitter3 which is … (#…

VitoAlbano:
…11116)

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

> - [ ] Yes
> - [ ] No

**Other information**:


lib/core/src/lib/auth/interfaces/authentication-service.interface.ts
```diff
@@ -16,18 +16,20 @@
  */
 
 import { HttpHeaders } from '@angular/common/http';
-import ee from 'event-emitter';
+import { EventEmitter } from 'eventemitter3';
 import { Observable } from 'rxjs';
 
+type EventEmitterInstance = InstanceType<typeof EventEmitter>;
```
ehsan-2019, line 22: I see this type in multiple places, maybe we can define it in one place and re-use it.

lib/js-api/src/authentication/contentAuth.ts
```diff
@@ -133,8 +132,7 @@ export class ContentAuth extends AlfrescoApiClient {
             );
         });
 
-        ee(promise); // jshint ignore:line
-        return promise;
+        return this.addPromiseListeners(promise, new EventEmitter());
```
ehsan-2019, line 135: hmmm, I'm not sure about passing a **new instance of EventEmitter** to `addPromiseListeners` every time validateTicket gets called. I think it will break the chain-ability of the returned promise. or maybe I'm wrong. just a heads-up to properly test this part.

lib/js-api/src/alfrescoApi.ts
```diff
@@ -26,11 +26,12 @@ import { AlfrescoApiConfig } from './alfrescoApiConfig';
 import { Authentication } from './authentication/authentication';
 import { AlfrescoApiType } from './to-deprecate/alfresco-api-type';
 import { HttpClient } from './api-clients/http-client.interface';
+import { AlfrescoApiClient, AlfrescoApiClientPromise } from './alfrescoApiClient';
 
-export class AlfrescoApi implements Emitter, AlfrescoApiType {
+type EventEmitterInstance = InstanceType<typeof EventEmitter>;
+
+export class AlfrescoApi extends AlfrescoApiClient implements AlfrescoApiType {
```
ehsan-2019, line 32: not sure about extending `AlfrescoApiClient` as it could make it more complicated to deprecate `LegacyHttpClient` in the future. I suggest to avoid unless it's strictly necessary.

lib/js-api/src/authentication/contentAuth.ts
VitoAlbano: This is exactly what was happening before, 'cause event-emitter was basically initialized via the ee(), i haven't changed the tests until is to move them to a more stable async/await of to provide a better mock for the missing calls, so we should be fine on testing side

lib/js-api/src/alfrescoApi.ts
VitoAlbano: Here as the other case i had to add the extends 'cause the Emitter interface doesn't exists but the main reason it was used here was to have the on/once etc methods, so we had to extend the base class instead of sharing the interface

ehsan-2019 (approved):

PR #11224, opened 2025-09-24
[ACS-10189] Remove old api docs generators from ADF

nikita-web-ua:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation
> - [X] Other... Please describe:

https://hyland.atlassian.net/browse/ACS-10189

- remove the tooling
- cleanup the root package.json

> - [ ] Yes
> - [X] No

**Other information**:


package.json
```diff
@@ -5,8 +5,6 @@
   "author": "Hyland Software, Inc. and its affiliates",
   "scripts": {
     "prepare": "husky install",
-    "build-doc-tools": "tsc -p ./tools/doc/tsconfig.json",
-    "docbuild": "(cd ./tools/doc; npm i) && node tools/doc/node_modules/typedoc/bin/typedoc --tsconfig lib/tsconfig.doc.json && node tools/doc/buildYamlSourceInfo.js docs/docs.json && node ./tools/doc/docProcessor.js",
```
AleksanderSklorz, line 9: do we need lib/tsconfig.doc.json anymore as you removed script which used that?

tools/doc/docProcessor.js
```diff
@@ -1,217 +0,0 @@
-const fs = require('fs');
-const path = require('path');
-const program = require('commander');
-const lodash = require('lodash');
-const jsyaml = require('js-yaml');
-const remark = require('remark');
-const frontMatter = require('remark-frontmatter');
-const mdCompact = require('mdast-util-compact');
-const minimatch = require('minimatch');
```
AleksanderSklorz, line 9: can we remove that minimatch dependency from package.json (I can't see more usages)?

nikita-web-ua: seems to be used here https://github.com/Alfresco/alfresco-ng2-components/blob/develop/lib/content-services/src/lib/common/services/upload.service.ts#L19

AleksanderSklorz: right, sorry, I didn't notice that

package.json
nikita-web-ua: I think it is not needed anymore. Removed, ty

PR #11238, opened 2025-10-01
[MNT-25336] ADW can't start a Microsoft 365 session when using Basic …

dominikiwanekhyland:
…Auth

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://hyland.atlassian.net/browse/MNT-25336

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/core/src/lib/auth/basic-auth/basic-alfresco-auth.service.ts
+++ b/lib/core/src/lib/auth/basic-auth/basic-alfresco-auth.service.ts
@@ -350,19 +350,23 @@ export class BasicAlfrescoAuthService extends BaseAuthenticationService {
      * @param requestUrl the request url
      * @returns The ticket or `null` if none was found
      */
-    getTicketEcmBase64(requestUrl: string): string | null {
+    getTicketEcmBase64(requestUrl: string): string {
+        let ticket = null;
         const bpmRoot = `/${this.appConfig.get<string>(AppConfigValues.CONTEXTROOTBPM) || 'activiti-app'}/`;
         const ecmRoot = `/${this.appConfig.get<string>(AppConfigValues.CONTEXTROOTECM) || 'alfresco'}/`;
+        const ooiServiceRootUrl = `/${this.appConfig.get<string>(AppConfigValues.OOI_CONNECTOR_URL) || 'ooi-service'}/`;
 
         if (requestUrl?.includes(ecmRoot) && !requestUrl.includes(bpmRoot)) {
-            return this.getContentServicesTicket();
+            ticket = this.getContentServicesTicket();
         } else if (requestUrl?.includes(bpmRoot) && !requestUrl.includes(ecmRoot)) {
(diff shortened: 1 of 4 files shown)


lib/core/src/lib/auth/basic-auth/basic-alfresco-auth.service.ts
```diff
@@ -351,19 +351,18 @@ export class BasicAlfrescoAuthService extends BaseAuthenticationService {
      * @returns The ticket or `null` if none was found
      */
     getTicketEcmBase64(requestUrl: string): string | null {
```
AleksanderSklorz: I think null type as return is redundant now and it's enough to leave string type here. Earlier there was 
let ticket = null; 
and 
return ticket; 

so null type had sense but now there can't be null I think

lib/core/src/lib/auth/basic-auth/basic-alfresco-auth.service.ts
```diff
@@ -351,19 +351,18 @@ export class BasicAlfrescoAuthService extends BaseAuthenticationService {
...
-            ticket = this.getContentServicesTicket();
+            return this.getContentServicesTicket();
         } else if (requestUrl?.includes(bpmRoot) && !requestUrl.includes(ecmRoot)) {
-            ticket = this.getProcessServicesTicket();
+            return this.getProcessServicesTicket();
         } else if (requestUrl?.includes(ecmRoot) && requestUrl.includes(bpmRoot)) {
-            ticket = requestUrl.indexOf(ecmRoot) < requestUrl.indexOf(bpmRoot) ? this.getContentServicesTicket() : this.getProcessServicesTicket();
+            return requestUrl.indexOf(ecmRoot) < requestUrl.indexOf(bpmRoot) ? this.getContentServicesTicket() : this.getProcessServicesTicket();
+        } else {
+            return this.getContentServicesTicket();
```
MichalKinas: I'm just thinking if that's the right solution, in that case every request outside ecmRoot or bpmRoot will receive content services ticket and I don't really think we want to do this, I wonder if this wouldn't be safer to rather add additional case for OOI connector rather then add it to every other request, @AleksanderSklorz what do you think?

AleksanderSklorz: Hmm, good point Michał. I agree with you that it will be safer, especially that in previous version we returned null for that case so we don't know what consequences this will cause if now we will be always that in this case

dominikiwanekhyland: @AleksanderSklorz @MichalKinas I thought it will be cleaner not to implement any login required for OOI Connector into ADF, but sure, I have updated my PR after your comments. THanks :)

MichalKinas (approved):

PR #11270, opened 2025-10-16
[ACS-10518]: Search input keyboard navigation accessibility fixes

rmnvch:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation
accessibility issues

https://hyland.atlassian.net/browse/ACS-10518

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/content-services/src/lib/content-node-selector/content-node-selector-panel/content-node-selector-panel.component-search.spec.ts
+++ b/lib/content-services/src/lib/content-node-selector/content-node-selector-panel/content-node-selector-panel.component-search.spec.ts
@@ -15,7 +15,7 @@
  * limitations under the License.
  */
 
-import { CUSTOM_ELEMENTS_SCHEMA } from '@angular/core';
+import { CUSTOM_ELEMENTS_SCHEMA, DebugElement } from '@angular/core';
 import { ComponentFixture, fakeAsync, flush, TestBed, tick } from '@angular/core/testing';
 import { By } from '@angular/platform-browser';
 import { Node, NodeEntry, NodePaging, RequestScope, ResultSetPaging, SiteEntry, SitePaging, SitePagingList } from '@alfresco/js-api';
@@ -63,16 +63,16 @@ describe('ContentNodeSelectorPanelComponent', () => {
     let searchQueryBuilderService: SearchQueryBuilderService;
     let testingUtils: UnitTestingUtils;
 
-    const typeToSearchBox = (searchTerm = 'string-to-search') => {
+    const typeToSearchBox = (searchTerm = 'string-to-search'): void => {
         const searchInput = testingUtils.getByCSS('[data-automation-id="content-node-selector-search-input"]');
         searchInput.nativeElement.value = searchTerm;
         component.searchInput.setValue(searchTerm);
         fixture.detectChanges();
     };
 
-    const getSearchIcon = (type: string) => testingUtils.getByCSS(`[data-automation-id="content-node-selector-search-${type}"]`);
+    const getSearchIcon = (type: string): DebugElement => testingUtils.getByCSS(`[data-automation-id="content-node-selector-search-${type}"]`);
 
-    const triggerSearchResults = (searchResults: ResultSetPaging) => {
+    const triggerSearchResults = (searchResults: ResultSetPaging): void => {
         const service = fixture.debugElement.injector.get(SearchQueryBuilderService);
         service.executed.next(searchResults);
     };


AleksanderSklorz:
@rmnvch please add A/N BDU label to that PR 
image

AleksanderSklorz:
@rmnvch if I open in one tab aca environment and in other tab local then I can see that when switching between browser tabs those X icons are changing position. This means that in local (on branch) that icon has a bit different styles related with position. Can you have a look and fix position to have the same like previously? Please ensure for each place which you changed 

https://github.com/user-attachments/assets/35fd4e1e-372f-4fc5-914b-3b5138064afe

lib/content-services/src/lib/content-node-selector/content-node-selector-panel/content-node-selector-panel.component.html
```diff
@@ -13,18 +13,26 @@
...
-            class="adf-content-node-selector-content-input-icon"
-            data-automation-id="content-node-selector-search-clear">clear
-        </mat-icon>
+        <button
+            matSuffix
+            mat-icon-button
+            *ngIf="searchTerm.length > 0"
+            data-automation-id="content-node-selector-search-clear"
+            class="adf-content-node-selector-search-clear-button"
+            (click)="clear()"
```
AleksanderSklorz, line 22: Because you moved click action from icon to button - can you correct existing test (or write new if doesn't exist) which checks if clicking on that button works?

AleksanderSklorz:
@rmnvch I see it works correctly for X icon but search icon (magnifier icon) can't be focused still. I believe it makes sense to don't focus search icon because search icon has not action and you can trigger search when focusing input and hit enter. But I can see that on jira they suggested this search icon should be focusable too. Did you check with them if we can skip focusing that? It's fine for me to skip that, we just may need to their confirmation to pass their checks next time: 
image

lib/content-services/src/lib/content-node-selector/content-node-selector-panel/content-node-selector-panel.component.html
rmnvch: @AleksanderSklorz , added

rmnvch:
> @rmnvch I see it works correctly for X icon but search icon (magnifier icon) can't be focused still. I believe it makes sense to don't focus search icon because search icon has not action and you can trigger search when focusing input and hit enter. But I can see that on jira they suggested this search icon should be focusable too. Did you check with them if we can skip focusing that? It's fine for me to skip that, we just may need to their confirmation to pass their checks next time: image

@AleksanderSklorz , ticket's updated https://hyland.atlassian.net/browse/ACS-10518

AleksanderSklorz:
> @rmnvch if I open in one tab aca environment and in other tab local then I can see that when switching between browser tabs those X icons are changing position. This means that in local (on branch) that icon has a bit different styles related with position. Can you have a look and fix position to have the same like previously? Please ensure for each place which you changed
> 
>  Screen.Recording.2025-10-17.at.13.22.10.mov

@rmnvch I see that X icon's position doesn't change but now search icon changes position: 

https://github.com/user-attachments/assets/0a5aa73d-1e88-464f-9a11-f27ef528b273

lib/content-services/src/lib/content-node-selector/content-node-selector-panel/content-node-selector-panel.component-search.spec.ts
```diff
@@ -378,6 +378,19 @@ describe('ContentNodeSelectorPanelComponent', () => {
...
 
+            it('should call clear method on the X (clear) button click', fakeAsync(() => {
+                fixture.detectChanges();
+                typeToSearchBox('123');
+                tick(debounceSearch);
+
+                fixture.detectChanges();
+
+                spyOn(component, 'clear');
+                const clearButton = fixture.debugElement.query(By.css('[data-automation-id="content-node-selector-search-clear"]'));
```
AleksanderSklorz: Similar comment which I added to other PR - please use UnitTestingUtils for getting element instead of calling query directly

AleksanderSklorz: I see that [data-automation-id="content-node-selector-search-clear"] selector is used multiple times in that file. After you use UniTestingUtils - could you please wrap getting that element in some function and reuse it for each place where that selector is used in that file? We prefer not repeating selectors as we should to avoid hardcoding strings (because we can always do typo etc.) so better to move to function and reuse

AleksanderSklorz: @rmnvch btw you have even clickByDataAutomationId function in UnitTestingUtils

lib/content-services/src/lib/content-node-selector/content-node-selector-panel/content-node-selector-panel.component-search.spec.ts
```diff
@@ -378,6 +378,19 @@ describe('ContentNodeSelectorPanelComponent', () => {
...
+                clearButton.nativeElement.click();
+                expect(component.clear).toHaveBeenCalled();
```
AleksanderSklorz, line 397: we don't check just if function is called because calling function doesn't mean that functionality in that function works as expected. Instead we test what that function does after clicking on that button

rmnvch: @AleksanderSklorz , 
as was suggested here https://github.com/Alfresco/alfresco-ng2-components/pull/11270#discussion_r2439473600
My idea was to add a test which checks if the button click is tied to the method call (clear()). This test is focused only on the DOM-to-Component binding, as the functionality inside clear() is verified by other existing tests. I believe it does its job.

lib/content-services/src/lib/content-node-selector/content-node-selector-panel/content-node-selector-panel.component-search.spec.ts
```diff
@@ -60,14 +61,17 @@ describe('ContentNodeSelectorPanelComponent', () => {
...
 
     const typeToSearchBox = (searchTerm = 'string-to-search') => {
-        const searchInput = fixture.debugElement.query(By.css('[data-automation-id="content-node-selector-search-input"]'));
+        const searchInput = testingUtils.getByCSS('[data-automation-id="content-node-selector-search-input"]');
         searchInput.nativeElement.value = searchTerm;
         component.searchInput.setValue(searchTerm);
         fixture.detectChanges();
     };
 
+    const getSearchIcon = (type: string) => testingUtils.getByCSS(`[data-automation-id="content-node-selector-search-${type}"]`);
```
AleksanderSklorz: missing return type for that function, should be 
```ts
 const getSearchIcon = (type: string): DebugElement => testingUtils.getByCSS(`[data-automation-id="content-node-selector-search-${type}"]`);

lib/content-services/src/lib/content-node-selector/content-node-selector-panel/content-node-selector-panel.component-search.spec.ts
AleksanderSklorz: @rmnvch yes by checking if click works I meant checking whole flow (test what that button does). Basically we try to test whole flow in our unit tests, for example: 
Imagine you have input and clear button. Then unit tests should to test if clicking on clear button will clear out input's value. So flow in test should be - first get input and button, fill input, click on button, check if input is empty. The only exception where we check if function is called is in case when the flow requires code from different dependency (for example if click event calls some function from some service then we don't test that function from service because it should have own tests, we just test if function from service is called). 

I will accept your solution in that PR because as you said it has already some tests (x years ago other people were working on that project so I believe they had own approaches) but just leaving that comment for future tests

PR #11275, opened 2025-10-17
[ACS-10325] a11y SR: Metadata edit button announced twice

nikita-web-ua:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation
> - [X] Other... Please describe: a11y

https://hyland.atlassian.net/browse/ACS-10325

Panel header & Edit button announced correctly by screen reader 

> - [ ] Yes
> - [X] No

**Other information**:

### DIFF
--- a/lib/content-services/src/lib/content-metadata/components/content-metadata/content-metadata.component.html
+++ b/lib/content-services/src/lib/content-metadata/components/content-metadata/content-metadata.component.html
@@ -9,7 +9,7 @@
         hideToggle>
         <mat-expansion-panel-header
             class="adf-metadata-properties-header"
-            [attr.aria-label]="('CORE.METADATA.BASIC.HEADER' | translate) + ' ' + ('CORE.METADATA.ACCESSIBILITY.SECTION' | translate)"
+            [attr.aria-label]="'CORE.METADATA.ACCESSIBILITY.SECTION' | translate: { sectionName: ('CORE.METADATA.BASIC.HEADER' | translate) }"
             [class.adf-metadata-properties-header-expanded]="currentPanel.panelTitle === DefaultPanels.PROPERTIES && currentPanel.expanded">
             <adf-content-metadata-header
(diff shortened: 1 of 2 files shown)


lib/content-services/src/lib/content-metadata/components/content-metadata/content-metadata.component.html
```diff
@@ -135,7 +138,7 @@
                         mat-icon-button
                         (click)="toggleGroupEditing(DefaultPanels.CATEGORIES, $event)"
                         [attr.title]="'CORE.METADATA.ACTIONS.EDIT' | translate"
-                        [attr.aria-label]="'CORE.METADATA.ACCESSIBILITY.EDIT' | translate"
+                        [attr.aria-label]="('CATEGORIES_MANAGEMENT.CATEGORIES_TITLE' | translate) + ' ' + ('CORE.METADATA.ACTIONS.EDIT' | translate)"
```
AleksanderSklorz: I would not concatenate translated texts. Reason is that in english it may sounds good, for example "Categories section" but for example in polish better order will be to have categories at second place, because in polish it would better sound to have "Sekcja kategorii". When using concatenation translation team can't change order of words but if we have it as single translation then they can use any order of words

AleksanderSklorz: For this you have 2 choices: 
1. You can have it as single translation key OR 
2. If you want to reuse "section" word then you can have translation key with parameter and pass as parameter translated word like categories or tags. It would be something like: 

'CORE.METADATA.ACTIONS.EDIT' | translate: { sectionName:  'CATEGORIES_MANAGEMENT.CATEGORIES_TITLE' | translate}

Both approaches should to allow translation team to change order of words

nikita-web-ua: Good point, applied. Thanks

PR #11285, opened 2025-10-20
AAE-37713 Amount locale display

tomaszhanaj:

> - [ ] Bugfix
> - [x] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://hyland.atlassian.net/browse/AAE-37713

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/core/src/lib/form/components/widgets/amount/amount.widget.html
+++ b/lib/core/src/lib/form/components/widgets/amount/amount.widget.html
@@ -28,7 +28,7 @@
                 (ngModelChange)="onFieldChangedAmountWidget()"
                 [disabled]="field.readOnly"
                 (focus)="amountWidgetOnFocus()"
-                (blur)="markAsTouched();amountWidgetOnBlur()"
+                (blur)="amountWidgetOnBlur()"
                 />
         </mat-form-field>
         <div class="adf-error-messages-container">
--- a/lib/core/src/lib/form/components/widgets/amount/amount.widget.spec.ts
+++ b/lib/core/src/lib/form/components/widgets/amount/amount.widget.spec.ts
@@ -79,7 +79,7 @@ describe('AmountWidgetComponent', () => {
 
     it('it should return locale based on browser', () => {
         const returnedLanguages: string[] = ['en-GB', 'en-US', 'en', 'de-DE', 'pl'];
-        const mockLanguages = spyOnProperty(window, 'navigator', 'get').and.returnValue({
+        const mockLanguages = spyOnProperty(window, 'navigator').and.returnValue({
             languages: returnedLanguages
         } as any);
         const locale = widget.getLocale();
@@ -90,7 +90,7 @@ describe('AmountWidgetComponent', () => {
 
     it('it should return default locale if browser does not return valid value', () => {
         const defaultLocale = 'en-US';
-        spyOnProperty(window, 'navigator', 'get').and.returnValue({
+        spyOnProperty(window, 'navigator').and.returnValue({
             languages: undefined
         } as any);
         const locale = widget.getLocale();
@@ -100,7 +100,7 @@ describe('AmountWidgetComponent', () => {
 
     it('should set initial values when enableDisplayBasedOnLocale is enabled', () => {
         const returnedLanguages: string[] = ['en-GB'];
-        spyOnProperty(window, 'navigator', 'get').and.returnValue({
+        spyOnProperty(window, 'navigator').and.returnValue({
             languages: returnedLanguages
         } as any);
         widget.field = new FormFieldModel(null, { id: 1, name: 'test', value: 25, currency: 'USD' });
@@ -116,7 +116,7 @@ describe('AmountWidgetComponent', () => {
 
     it('should set initial values with correct currency', () => {
         const returnedLanguages: string[] = ['en-GB'];
-        spyOnProperty(window, 'navigator', 'get').and.returnValue({
+        spyOnProperty(window, 'navigator').and.returnValue({
             languages: returnedLanguages
         } as any);
         widget.field = new FormFieldModel(null, { id: 2, name: 'test', value: 25, currency: 'GBP' });
@@ -130,7 +130,7 @@ describe('AmountWidgetComponent', () => {
 
     it('should set initial values with correct currency icon', () => {
         const returnedLanguages: string[] = ['en-GB'];
-        spyOnProperty(window, 'navigator', 'get').and.returnValue({
+        spyOnProperty(window, 'navigator').and.returnValue({
             languages: returnedLanguages
         } as any);
         widget.field = new FormFieldModel(null, { id: 2, name: 'test', value: 25, currency: '¥' });
@@ -144,7 +144,7 @@ describe('AmountWidgetComponent', () => {
 
     it('should set initial values without currency', () => {
         const returnedLanguages: string[] = ['en-GB'];
-        spyOnProperty(window, 'navigator', 'get').and.returnValue({
+        spyOnProperty(window, 'navigator').and.returnValue({
             languages: returnedLanguages
         } as any);
         widget.field = new FormFieldModel(null, { id: 3, name: 'test', value: 25, currency: '' });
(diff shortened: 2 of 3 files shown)


lib/core/src/lib/form/components/widgets/amount/amount.widget.spec.ts
```diff
@@ -76,6 +77,139 @@ describe('AmountWidgetComponent', () => {
         expect(widget.placeholder).toBe('1234');
     });
 
+    it('it should return locale based on browser', () => {
+        const returnedLanguages: string[] = ['en-GB', 'en-US', 'en', 'de-DE', 'pl'];
+        const mockLanguages = spyOnProperty(window, 'navigator', 'get').and.returnValue({
```
BSekula: is it working? I do not see `get` on navigator 🤔 

https://developer.mozilla.org/en-US/docs/Web/API/Navigator

We can e.g. use
https://developer.mozilla.org/en-US/docs/Web/API/Navigator/languages

[getLocale](https://github.com/Alfresco/alfresco-ng2-components/pull/11285/files#diff-42b17c99fffe2f2da81521730968631282f1ec97c460a86af7f532edb8b2231bR134) does not use get() but languages 🤔

lib/core/src/lib/form/components/widgets/amount/amount.widget.ts
```diff
@@ -49,36 +52,115 @@ export const ADF_AMOUNT_SETTINGS = new InjectionToken<AmountWidgetSettings>('adf
...
+            this.amountWidgetValue = this.valueAsNumber || this.valueAsNumber === 0 ? this.valueAsNumber.toString() : null;
+        }
+    }
+
+    getLocale() {
+        const defaultLocale = 'en-US';
+        if (typeof window?.navigator === 'undefined') {
+            return defaultLocale;
+        }
+        const wn = window.navigator as any;
```
BSekula: There is a type for it i.e. `Navigator`

lib/core/src/lib/form/components/widgets/amount/amount.widget.ts
```diff
@@ -49,36 +52,115 @@ export const ADF_AMOUNT_SETTINGS = new InjectionToken<AmountWidgetSettings>('adf
...
 
             if (this.field.readOnly) {
-                this.showPlaceholder = this.settings?.showReadonlyPlaceholder;
+                this.showPlaceholder = this.showReadonlyPlaceholder;
+            }
+            this.setInitialValues();
+        }
+    }
+
+    amountWidgetOnBlur() {
```
BSekula: Can we add return type to all methods

lib/core/src/lib/form/components/widgets/amount/amount.widget.ts
```diff
@@ -49,36 +52,115 @@ export const ADF_AMOUNT_SETTINGS = new InjectionToken<AmountWidgetSettings>('adf
...
+                this.valueAsNumber = null;
+                this.valueAsString = null;
+                this.amountWidgetValue = null;
             }
         }
     }
+
+    amountWidgetOnFocus() {
+        if (this.enableDisplayBasedOnLocale) {
+            this.amountWidgetValue = this.valueAsNumber || this.valueAsNumber === 0 ? this.valueAsNumber.toString() : null;
```
BSekula:
```suggestion
            const hasValue = this.valueAsNumber === 0 || !!this.valueAsNumber;
            this.amountWidgetValue = hasValue ? this.valueAsNumber.toString() : null;
```

Wdyt?

lib/core/src/lib/form/components/widgets/amount/amount.widget.ts
```diff
@@ -49,36 +52,115 @@ export const ADF_AMOUNT_SETTINGS = new InjectionToken<AmountWidgetSettings>('adf
...
+    }
+
+    onFieldChangedAmountWidget() {
+        this.field.value = this.amountWidgetValue;
+        super.onFieldChanged(this.field);
+    }
+
+    setInitialValues(): void {
+        if (this.enableDisplayBasedOnLocale) {
+            this.decimalProperty = this.field.enableFractions ? '1.2-2' : '1.0-0';
```
BSekula: can we somehow make this values `1.2-2'`, `1.0-0` more understandable? 

You can check magic number pattern e.g. https://refactoring.guru/replace-magic-number-with-symbolic-constant

> A magic number is a numeric value that’s encountered in the source but has no obvious meaning. This “anti-pattern” makes it harder to understand the program and refactor the code.

lib/core/src/lib/form/components/widgets/amount/amount.widget.html
```diff
@@ -21,12 +23,13 @@
...
-                [(ngModel)]="field.value"
-                (ngModelChange)="onFieldChanged(field)"
+                [value]="amountWidgetValue"
+                [(ngModel)]="amountWidgetValue"
+                (ngModelChange)="onFieldChangedAmountWidget()"
                 [disabled]="field.readOnly"
-                (blur)="markAsTouched()"
-            />
+                (focus)="amountWidgetOnFocus()"
+                (blur)="markAsTouched();amountWidgetOnBlur()"
```
BSekula: maybe we can add `markAsTouched` into `amountWidgetOnBlur`?

lib/core/src/lib/form/components/widgets/amount/amount.widget.ts
```diff
@@ -49,36 +52,115 @@ export const ADF_AMOUNT_SETTINGS = new InjectionToken<AmountWidgetSettings>('adf
...
 
+    amountWidgetValue: string;
     currency: string = AmountWidgetComponent.DEFAULT_CURRENCY;
+    currencyDisplay: string | boolean = 'symbol';
+    decimalProperty: string;
+    enableDisplayBasedOnLocale: boolean;
+    locale: string;
+    showReadonlyPlaceholder: boolean;
+    valueAsNumber: number;
+    valueAsString: string;
```
BSekula: Do we need both 🤔  `valueAsNumber` and `valueAsString`

BSekula: We are only using it here, I do not think we need this additional prop 🤔 
`this.amountWidgetValue = this.valueAsString;`

If the `valueAsNumber` is only set here `this.valueAsNumber = this.field.value;`

This is also not needed IMO

lib/core/src/lib/form/components/widgets/amount/amount.widget.spec.ts
tomaszhanaj: I think I took it from here 
lib/core/src/lib/common/services/download.service.spec.ts
and navigator is read-only so I thought it is valid. But it works also without it. Fixed.

lib/core/src/lib/form/components/widgets/amount/amount.widget.ts
tomaszhanaj: updated

lib/core/src/lib/form/components/widgets/amount/amount.widget.ts
tomaszhanaj: updated

lib/core/src/lib/form/components/widgets/amount/amount.widget.ts
tomaszhanaj: ok, updated

lib/core/src/lib/form/components/widgets/amount/amount.widget.ts
tomaszhanaj: updated

lib/core/src/lib/form/components/widgets/amount/amount.widget.html
tomaszhanaj: fixed

lib/core/src/lib/form/components/widgets/amount/amount.widget.ts
tomaszhanaj: valueAsNumber is needed. it is used in amountWidgetOnFocus() when user enters the field its value is assigned to field. It is set on amountWidgetOnBlur() when user leaves the field it is used to store numeric value and when setting initial values (so it is set in two places). Its intention is to store value from field as number to not convert it from string. But valueAsString can be removed. updated.

PR #11291, opened 2025-10-21
HXIDP-5126 Fix adf-cloud-task-screen re-initialization on completing

lkastsiuchenka:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://hyland.atlassian.net/browse/HXIDP-5126

adf-cloud-task-screen does not initialize again after task completion

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/content-services/src/lib/content-metadata/components/content-metadata/content-metadata.component.html
+++ b/lib/content-services/src/lib/content-metadata/components/content-metadata/content-metadata.component.html
@@ -9,6 +9,7 @@
         hideToggle>
         <mat-expansion-panel-header
             class="adf-metadata-properties-header"
+            [attr.aria-label]="'CORE.METADATA.ACCESSIBILITY.SECTION' | translate: { sectionName: ('CORE.METADATA.BASIC.HEADER' | translate) }"
             [class.adf-metadata-properties-header-expanded]="currentPanel.panelTitle === DefaultPanels.PROPERTIES && currentPanel.expanded">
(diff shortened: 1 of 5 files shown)


lib/process-services-cloud/src/lib/task/task-form/components/user-task-cloud/user-task-cloud.component.ts
```diff
@@ -247,8 +247,10 @@ export class UserTaskCloudComponent implements OnInit, OnChanges {
         this.taskClaimed.emit(this.taskId);
     }
 
-    onCompleteTask(openNextTask: boolean = false): void {
-        this.loadTask();
+    onCompleteTask(taskType: UserTaskType, openNextTask: boolean = false): void {
```
tomaszhanaj: Maybe it would be better to change the order of arguments. To not provide breaking change, old code will be passing only the first argument and it will be handled. And second parameter should be optional. Like:
`onCompleteTask(openNextTask: boolean = false, taskType?: UserTaskType): void {`
 Then in condition:
` if (!taskType || taskType !== UserTaskContentType.Screen) {  `
so we call loadTask when taskType is not passed(so old code works as previously) or when taskType is passed and it is different than Screen type. What do you think?

lkastsiuchenka: fixed

PR #11293, opened 2025-10-21
AAE-38748 Change clipboard directive to be a tooltip

jcain82:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://hyland.atlassian.net/browse/AAE-38748

Instead of inserting an element in the container with the parent element we now show a matTooltip
image

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,9 @@
     "build:libs": "nx run-many -t build --prod --skip-nx-cache",
     "build:schematics": "nx run-many -t build-schematics",
     "publish": "nx run-many -t npm-publish",
-    "clean": "rimraf dist node_modules dist/libs"
+    "clean": "rimraf dist node_modules dist/libs",
+    "nx:run-target": "nx run",
+    "nx:run-many": "nx run-many"
   },
   "repository": {
     "type": "git",


jcain82:
@eromano, I made the switch to a floating tooltip but there are some accessibility issues with doing it this way and because we set title elements as you can see from the image there can be overlap. The way around this would be to instead update the title element to append the "Click to Copy" text or to abandon using unnecessary title elements since screen readers already know how to read tabular data from a grid so setting the title to the cell value doesn't make any sense.

lib/core/src/lib/clipboard/clipboard.theme.scss
```diff
@@ -1,19 +0,0 @@
-.adf-copy-tooltip {
```
DavidOlson-Hyland, line 1: I'm assuming this file was removed intentionally?

DavidOlson-Hyland (approved):

lib/core/src/lib/clipboard/clipboard.theme.scss
jcain82: Yes since there is no longer a component to style it is not necessary.

amolodyh-hyland (approved):

sdonnell-hyland (approved):

PR #11311, opened 2025-10-28
[ACS-10196] other a11y all libraries error message not provided for name and library id fields

g-jaskowski:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://hyland.atlassian.net/browse/ACS-10196

Added error messages for required validator, improved unit test coverage, refactored usage of deprecated api

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/content-services/src/lib/dialogs/library/library.dialog.spec.ts
+++ b/lib/content-services/src/lib/dialogs/library/library.dialog.spec.ts
@@ -350,11 +350,27 @@ describe('LibraryDialogComponent', () => {
         expect(component.form.controls.id.errors).toEqual({ message: 'LIBRARY.ERRORS.ILLEGAL_CHARACTERS' });
     });
 
+    it('should not show required error on opening dialog', async () => {
+        findSitesSpy.and.returnValue(Promise.resolve(findSitesResponse));
+        fixture.detectChanges();
+        const titleFormField = await unitTestingUtils.getMatFormField();
+        const errors = await titleFormField.getTextErrors();
+        expect(errors.length).toBe(0);
+    });
+
+    it('should mark title as touched on value change', fakeAsync(() => {
+        findSitesSpy.and.returnValue(Promise.resolve(findSitesResponse));
+        spyOn(sitesService, 'getSite').and.returnValue(of(null));
+        fixture.detectChanges();
+        expect(component.form.controls.title.touched).toBeFalse();
+        setFormFieldValue('title', 'library title');
+        expect(component.form.controls.title.touched).toBeTrue();
+    }));
+
     it('should show correct error message when there is no library title', async () => {
         findSitesSpy.and.returnValue(Promise.resolve(findSitesResponse));
         fixture.detectChanges();
         component.form.controls.title.setValue('');
-        component.form.controls.title.markAsTouched();
         fixture.detectChanges();
 
         const titleFormField = await unitTestingUtils.getMatFormField();
--- a/lib/content-services/src/lib/dialogs/library/library.dialog.ts
+++ b/lib/content-services/src/lib/dialogs/library/library.dialog.ts
@@ -30,7 +30,7 @@ import {
 import { MatDialogModule, MatDialogRef } from '@angular/material/dialog';
 import { QueriesApi, SiteBodyCreate, SiteEntry, SitePaging } from '@alfresco/js-api';
 import { NotificationService } from '@alfresco/adf-core';
-import { debounceTime, finalize, map, mergeMap } from 'rxjs/operators';
+import { debounceTime, finalize, map, mergeMap, take } from 'rxjs/operators';
 import { SitesService } from '../../common/services/sites.service';
 import { CommonModule } from '@angular/common';
 import { TranslatePipe } from '@ngx-translate/core';
@@ -125,6 +125,10 @@ export class LibraryDialogComponent implements OnInit {
         });
 
         this.visibilityOption = this.visibilityOptions[0].value;
+
+        this.form.controls['title'].valueChanges
+            .pipe(take(1), takeUntilDestroyed(this.destroyRef))
+            .subscribe(() => this.form.controls['title'].markAsTouched());
 
         this.form.controls['title'].valueChanges
             .pipe(
@@ -248,13 +252,13 @@ export class LibraryDialogComponent implements OnInit {
         }
     }
 
-    private forbidSpecialCharacters({ value }: UntypedFormControl) {
+    private forbidSpecialCharacters({ value }: UntypedFormControl): ValidationErrors | null {
         if (value === null || value.length === 0) {
             return null;
         }
 
         const validCharacters: RegExp = /[^A-Za-z0-9-]/;
-        const isValid: boolean = !validCharacters.test(value);
+        const isValid = !validCharacters.test(value);
 
         return isValid
             ? null
@@ -263,12 +267,12 @@ export class LibraryDialogComponent implements OnInit {
               };
     }
 
-    private forbidOnlySpaces({ value }: UntypedFormControl) {
+    private forbidOnlySpaces({ value }: UntypedFormControl): ValidationErrors | null {
         if (value === null || value.length === 0) {
             return null;
         }
 
-        const isValid: boolean = !!value.trim();
+        const isValid = !!value.trim();
 
         return isValid
             ? null
@@ -277,12 +281,12 @@ export class LibraryDialogComponent implements OnInit {
               };
     }
 
-    private minLengthTrimmed({ value }: UntypedFormControl) {
+    private minLengthTrimmed({ value }: UntypedFormControl): ValidationErrors | null {
         if (value === null || value.length === 0) {
             return null;
         }
 
-        const isValid: boolean = value.trim().length !== 1;
+        const isValid = value.trim().length !== 1;
 
         return isValid
             ? null


AleksanderSklorz:
@g-jaskowski name field is getting red and displays error immediately when I just open dialog but for id field we need to unfocus field. Can we make also the same for name because I think there is no sense to display error and make field red until we unfocus name field because why user should to see error when user didn't do anything yet

https://github.com/user-attachments/assets/90ebe174-32c3-40bf-ad65-d02b7f8fd70c

AleksanderSklorz:
@g-jaskowski when I have 1 character in name then it says name requires at least 2 characters but when I enter space then this error is gone. I think it should not work like that (white characters should not count I believe): 

https://github.com/user-attachments/assets/db893b7a-7d08-4dce-b11f-9bb892ebf080

lib/content-services/src/lib/dialogs/library/library.dialog.html
```diff
@@ -11,21 +11,21 @@ <h2 mat-dialog-title>{{ createTitle | translate }}</h2>
...
-        {{ 'LIBRARY.ERRORS.TITLE_TOO_SHORT' | translate }}
-      </mat-error>
-
-      <mat-error *ngIf="form.controls['title'].errors?.message">
-        {{ form.controls['title'].errors?.message | translate }}
-      </mat-error>
+      @if(libraryTitleExists){
+        <mat-hint>{{ 'LIBRARY.HINTS.SITE_TITLE_EXISTS' | translate }}</mat-hint>
+      }
+      @if (form.controls['title'].hasError('maxlength')) {
```
AleksanderSklorz: If possible we should avoid to use functions inside templates. Can we convert that to something like: 
```
@if (form.controls['title'].errors.maxlength) {

}
```
Then we can avoid calling hasError function. Please also apply for rest places in that file. You can see example in 23rd line

lib/content-services/src/lib/dialogs/library/library.dialog.html
```diff
@@ -11,21 +11,21 @@ <h2 mat-dialog-title>{{ createTitle | translate }}</h2>
...
-      </mat-error>
-
-      <mat-error *ngIf="form.controls['title'].hasError('minlength')">
-        {{ 'LIBRARY.ERRORS.TITLE_TOO_SHORT' | translate }}
-      </mat-error>
-
-      <mat-error *ngIf="form.controls['title'].errors?.message">
-        {{ form.controls['title'].errors?.message | translate }}
-      </mat-error>
+      @if(libraryTitleExists){
```
AleksanderSklorz: formatting issue - can you add space after if and before { ?

lib/content-services/src/lib/dialogs/library/library.dialog.spec.ts
```diff
@@ -22,17 +22,21 @@ import { MatDialogRef } from '@angular/material/dialog';
...
+import { TestbedHarnessEnvironment } from '@angular/cdk/testing/testbed';
+import { MatRadioGroupHarness } from '@angular/material/radio/testing';
+import { MatFormFieldHarness } from '@angular/material/form-field/testing';
 
 describe('LibraryDialogComponent', () => {
     let fixture: ComponentFixture<LibraryDialogComponent>;
     let component: LibraryDialogComponent;
     let sitesService: SitesService;
-    let findSitesSpy;
+    let findSitesSpy: jasmine.Spy<(term: string, opts?: any) => Promise<SitePaging>>;
```
AleksanderSklorz: Looking at real declaration of findSites function I believe opts should have FindQuery type instead of any

lib/content-services/src/lib/dialogs/library/library.dialog.spec.ts
```diff
@@ -198,11 +217,31 @@ describe('LibraryDialogComponent', () => {
         });
     }));
 
+    it('should show generic error notification on error code other than 409', fakeAsync(() => {
```
AleksanderSklorz: This test looks similar to next test which is below. Can you create some reusable function to avoid repeating so many lines?

lib/content-services/src/lib/dialogs/library/library.dialog.spec.ts
```diff
@@ -270,4 +309,103 @@ describe('LibraryDialogComponent', () => {
...
+        findSitesSpy.and.returnValue(Promise.resolve(findSitesResponse));
+        fixture.detectChanges();
+        component.form.controls.title.setValue('library');
+        fixture.detectChanges();
+
+        expect(component.libraryTitleExists).toBe(false);
+    });
+
+    it('should clear timeout if validator is called multiple times', () => {
+        spyOn(component['sitesService'], 'getSite').and.returnValue(of(null));
```
AleksanderSklorz: we should not refer to private things using []. Please use TestBed.inject instead to get reference to SitesService

lib/content-services/src/lib/dialogs/library/library.dialog.spec.ts
```diff
@@ -270,4 +309,103 @@ describe('LibraryDialogComponent', () => {
...
+
+    it('should show correct error message when there is no library title', async () => {
+        findSitesSpy.and.returnValue(Promise.resolve(findSitesResponse));
+        fixture.detectChanges();
+        component.form.controls.title.setValue('');
+        component.form.controls.title.markAsTouched();
+        fixture.detectChanges();
+
+        const loader = TestbedHarnessEnvironment.loader(fixture);
+        const titleFormField = await loader.getHarness(MatFormFieldHarness);
```
AleksanderSklorz: UnitTestingUtils class has function to get harness for form field - please reuse instead of getting this manually

lib/content-services/src/lib/dialogs/library/library.dialog.spec.ts
```diff
@@ -270,4 +309,103 @@ describe('LibraryDialogComponent', () => {
...
+
+    it('should show correct error message when there is no library id', async () => {
+        findSitesSpy.and.returnValue(Promise.resolve(findSitesResponse));
+        fixture.detectChanges();
+        component.form.controls.id.setValue('');
+        component.form.controls.id.markAsTouched();
+        fixture.detectChanges();
+
+        const loader = TestbedHarnessEnvironment.loader(fixture);
+        const formFields = await loader.getAllHarnesses(MatFormFieldHarness);
```
AleksanderSklorz: same like https://github.com/Alfresco/alfresco-ng2-components/pull/11311/files#r2472463387

g-jaskowski:
> @g-jaskowski name field is getting red and displays error immediately when I just open dialog but for id field we need to unfocus field. Can we make also the same for name because I think there is no sense to display error and make field red until we unfocus name field because why user should to see error when user didn't do anything yet
> 
>  Screen.Recording.2025-10-29.at.11.10.11.mov

I've created another PR in ACA solving this issue: https://github.com/Alfresco/alfresco-content-app/pull/4883



> @g-jaskowski when I have 1 character in name then it says name requires at least 2 characters but when I enter space then this error is gone. I think it should not work like that (white characters should not count I believe):
> 
>  Screen.Recording.2025-10-29.at.11.14.08.mov

Replaced the minlength validator with custom solution

lib/content-services/src/lib/dialogs/library/library.dialog.html
g-jaskowski: Replaced all hasError calls

lib/content-services/src/lib/dialogs/library/library.dialog.html
g-jaskowski: I've corrected formatting, thanks for pointing it out

lib/content-services/src/lib/dialogs/library/library.dialog.spec.ts
g-jaskowski: Corrected opts type, thanks

lib/content-services/src/lib/dialogs/library/library.dialog.spec.ts
g-jaskowski: I've created a function for this case as well as setting form fields' values, thanks

lib/content-services/src/lib/dialogs/library/library.dialog.spec.ts
g-jaskowski: Fixed to use real service, thanks for catching that

lib/content-services/src/lib/dialogs/library/library.dialog.spec.ts
g-jaskowski: Done, thanks!

AleksanderSklorz:
> > @g-jaskowski name field is getting red and displays error immediately when I just open dialog but for id field we need to unfocus field. Can we make also the same for name because I think there is no sense to display error and make field red until we unfocus name field because why user should to see error when user didn't do anything yet
> > Screen.Recording.2025-10-29.at.11.10.11.mov
> 
> I've created another PR in ACA solving this issue: [Alfresco/alfresco-content-app#4883](https://github.com/Alfresco/alfresco-content-app/pull/4883)
> 
> > @g-jaskowski when I have 1 character in name then it says name requires at least 2 characters but when I enter space then this error is gone. I think it should not work like that (white characters should not count I believe):
> > Screen.Recording.2025-10-29.at.11.14.08.mov
> 
> Replaced the minlength validator with custom solution

@g-jaskowski first fix breaks now different case. Error for required field is displayed when I unfocus field what is correct (because I should do any action before displaying error) but now others errors are not displayed after making any action. For example - if I type only spaces then I should see error because I did action (I typed spaces) but I need to unfocus. Unfocus should be needed only for displaying required error

https://github.com/user-attachments/assets/ec3c91b7-42f6-4ed6-8eda-fe252b74caa0

lib/content-services/src/lib/dialogs/library/library.dialog.ts
```diff
@@ -277,6 +277,20 @@ export class LibraryDialogComponent implements OnInit {
               };
     }
 
+    private minLengthTrimmed({ value }: UntypedFormControl) {
```
AleksanderSklorz: missing return type, should be ValidationErrors

lib/content-services/src/lib/dialogs/library/library.dialog.ts
```diff
@@ -275,24 +277,39 @@ export class LibraryDialogComponent implements OnInit {
               };
     }
 
-    private createSiteIdValidator() {
+    private minLengthTrimmed({ value }: UntypedFormControl) {
+        if (value === null || value.length === 0) {
+            return null;
+        }
+
+        const isValid: boolean = value.trim().length !== 1;
```
AleksanderSklorz: no need to set type manually - when you assign condition to const/variable then it detects automatically that condition returns boolean

g-jaskowski:
> > > @g-jaskowski name field is getting red and displays error immediately when I just open dialog but for id field we need to unfocus field. Can we make also the same for name because I think there is no sense to display error and make field red until we unfocus name field because why user should to see error when user didn't do anything yet
> > > Screen.Recording.2025-10-29.at.11.10.11.mov
> > 
> > 
> > I've created another PR in ACA solving this issue: [Alfresco/alfresco-content-app#4883](https://github.com/Alfresco/alfresco-content-app/pull/4883)
> > > @g-jaskowski when I have 1 character in name then it says name requires at least 2 characters but when I enter space then this error is gone. I think it should not work like that (white characters should not count I believe):
> > > Screen.Recording.2025-10-29.at.11.14.08.mov
> > 
> > 
> > Replaced the minlength validator with custom solution
> 
> @g-jaskowski first fix breaks now different case. Error for required field is displayed when I unfocus field what is correct (because I should do any action before displaying error) but now others errors are not displayed after making any action. For example - if I type only spaces then I should see error because I did action (I typed spaces) but I need to unfocus. Unfocus should be needed only for displaying required error
> 
>  Screen.Recording.2025-11-03.at.13.09.04.mov

Validation is now manually triggered after user interacts with the title input

lib/content-services/src/lib/dialogs/library/library.dialog.ts
g-jaskowski: added to all custom validators, thanks

lib/content-services/src/lib/dialogs/library/library.dialog.ts
g-jaskowski: removed in all custom validators

nikita-web-ua (approved):

PR #11317, opened 2025-10-30
[MNT-25412] Proper date filter clean up after reset event

MichalKinas:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://hyland.atlassian.net/browse/MNT-25412

Date filters are properly cleaned up after receiving the reset event.

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/content-services/src/lib/search/components/search-date-range-tabbed/search-date-range-tabbed.component.ts
+++ b/lib/content-services/src/lib/search/components/search-date-range-tabbed/search-date-range-tabbed.component.ts
@@ -43,9 +43,9 @@ const DEFAULT_DATE_DISPLAY_FORMAT = 'dd-MMM-yy';
 })
 export class SearchDateRangeTabbedComponent implements SearchWidget, OnInit {
     private value: { [key: string]: Partial<SearchDateRange> } = {};
-    private queryMapByField = new Map<string, string>();
-    private displayValueMapByField = new Map<string, string>();
 
+    private readonly queryMapByField = new Map<string, string>();
+    private readonly displayValueMapByField = new Map<string, string>();
     private readonly destroyRef = inject(DestroyRef);
     private readonly resetSubject$ = new Subject<void>();
 


docs/content-services/components/search-date-range.component.md
```diff
@@ -25,12 +25,13 @@ before that release, please ensure that your component configuration is updated
 
 ### Properties
 
-| Name         | Type            | Description                                                                                                                                                                                                                                                |
+| Name         | Type                                             | Description                                                                                                                                                                                                                                                |
```
AleksanderSklorz: can we format that table properly?
image

lib/content-services/src/lib/search/components/search-date-range-tabbed/search-date-range-tabbed.component.ts
```diff
@@ -42,6 +42,13 @@ const DEFAULT_DATE_DISPLAY_FORMAT = 'dd-MMM-yy';
     encapsulation: ViewEncapsulation.None
 })
 export class SearchDateRangeTabbedComponent implements SearchWidget, OnInit {
+    private value: { [key: string]: Partial<SearchDateRange> } = {};
+    private queryMapByField: Map<string, string> = new Map<string, string>();
```
AleksanderSklorz: no need to set type manually when we initialize field inline. The same next line

lib/content-services/src/lib/search/components/search-date-range-tabbed/search-date-range/search-date-range.component.ts
```diff
@@ -189,4 +194,14 @@ export class SearchDateRangeComponent implements OnInit {
                 return true;
         }
     }
+
+    reset(): void {
+        this.form.reset({
```
AleksanderSklorz, line 199: I think it can be simplified to just: 
this.form.reset(); 

Value is optional and is needed only when we want to reset form to different value than initial but I can see that the same object was set as initial value so we can have reset without argument

lib/content-services/src/lib/search/components/search-date-range-tabbed/search-date-range/search-date-range.component.ts
```diff
@@ -189,4 +194,14 @@ export class SearchDateRangeComponent implements OnInit {
                 return true;
         }
     }
+
+    reset(): void {
```
AleksanderSklorz, line 198: when you address this comment: https://github.com/Alfresco/alfresco-ng2-components/pull/11317/files#r2481260536

then do we still need that reset wrapper function? I guess then it will be redundant because that reset function will wrap just other empty reset function so we can call this.form.reset directly instead of wrapping in extra reset function

lib/content-services/src/lib/search/components/search-date-range-tabbed/search-date-range/search-date-range.component.ts
MichalKinas: I was testing this and for some reason form is not getting the initial value when reset is called with no params, all values in the controls are just null

docs/content-services/components/search-date-range.component.md
MichalKinas: Fixed

lib/content-services/src/lib/search/components/search-date-range-tabbed/search-date-range-tabbed.component.ts
MichalKinas: Done

PR #11318, opened 2025-10-31
[MNT-25413] ADW - First login after a clear cache does not show the s…

dominikiwanekhyland:
…aved searches

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://hyland.atlassian.net/browse/MNT-25413

> - [ ] Yes
> - [ ] No

**Other information**:

### DIFF
--- a/lib/content-services/src/lib/common/services/saved-searches.service.spec.ts
+++ b/lib/content-services/src/lib/common/services/saved-searches.service.spec.ts
@@ -57,7 +57,6 @@ describe('SavedSearchesService', () => {
         });
         service = TestBed.inject(SavedSearchesService);
         authService = TestBed.inject(AuthenticationService);
-        spyOn(service.nodesApi, 'getNode').and.callFake(() => Promise.resolve({ entry: { id: testNodeId } } as NodeEntry));
         spyOn(service.nodesApi, 'getNodeContent').and.callFake(() => createBlob());
         spyOn(service.nodesApi, 'deleteNode').and.callFake(() => Promise.resolve());
         spyOn(service.preferencesApi, 'getPreference').and.callFake(() =>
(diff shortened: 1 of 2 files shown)


lib/content-services/src/lib/common/services/saved-searches.service.ts
```diff
@@ -70,22 +73,19 @@ export class SavedSearchesService {
...
                         return this.migrateSavedSearches();
                     } else {
-                        return from(this.preferencesApi.getPreference('-me-', 'saved-searches')).pipe(
-                            map((preference) => JSON.parse(preference.entry.value)),
-                            catchError(() => of([]))
-                        );
+                        return this.getSavedSearchesFromPreferenceApi();
                     }
+                }),
+                catchError(() => {
```
AleksanderSklorz: This line gives eslint warning, please fix  
image

lib/content-services/src/lib/common/services/saved-searches.service.spec.ts
```diff
@@ -177,6 +177,34 @@ describe('SavedSearchesService', () => {
         });
     });
 
+    it('should fallback to preferences API if getting saved searches node ID fails', (done) => {
+        spyOn(authService, 'getUsername').and.returnValue(testUserName);
+        spyOn(localStorage, 'getItem').and.returnValue('');
+        const error = new Error(JSON.stringify({ error: { statusCode: 500 } }));
+        (service.nodesApi.getNode as jasmine.Spy).and.returnValue(Promise.reject(error));
```
AleksanderSklorz: jasmine.Spy is generic type so we should specify type of function. If we don't specify that then code is not type safe because then for example returnValue function doesn't check type property: 
Screenshot 2025-10-31 at 13 42 02

You see on screenshot - it thinks that getNode returns any. 

It should be specified by something like that: 
```ts
service.nodesApi.getNode as jasmine.Spy<(nodeId: string, opts?: { relativePath?: string } & NodesIncludeQuery) => Promise<NodeEntry>>
```
Because that is a bit long line then you can store result of spyOn(service.nodesApi, 'getNode') which is called in 61st line in some variable (and set suggested generic type for that variable). Thanks that you will have setting this type only once without repeating that

lib/content-services/src/lib/common/services/saved-searches.service.spec.ts
```diff
@@ -72,136 +71,142 @@ describe('SavedSearchesService', () => {
...
-        service.savedSearches$.pipe().subscribe((searches) => {
-            expect(searches.length).toBe(2);
-            expect(searches[0].name).toBe('Search 1');
-            expect(service.preferencesApi.getPreference).toHaveBeenCalledWith('-me-', 'saved-searches');
-            done();
-        });
+        it('should save a new search', (done) => {
+            spyOn(authService, 'getUsername').and.callFake(() => testUserName);
+            spyOn(localStorage, 'getItem').and.callFake(() => 'true');
+            const newSearch = { name: 'Search 3', description: 'Description 3', encodedUrl: 'url3' };
```
AleksanderSklorz, line 111: please set one of those types to this const: 
```ts
SavedSearch
```
or
```ts
Pick<SavedSearch, 'name' | 'description' | 'encodedUrl'>

File: lib/content-services/src/lib/common/services/saved-searches.service.spec.ts [hunk 4]
```diff
@@ -72,136 +71,142 @@ describe('SavedSearchesService', () => {
...
                 done();
             });
+
+            service.getSavedSearches().subscribe();
+        });
+
+        it('should emit updated saved searches after saving a new search', (done) => {
+            spyOn(authService, 'getUsername').and.callFake(() => testUserName);
+            spyOn(localStorage, 'getItem').and.callFake(() => 'true');
+            const newSearch = { name: 'Search 3', description: 'Description 3', encodedUrl: 'url3' };
```
AleksanderSklorz, line 139: please set one of those types to this const: 
```ts
SavedSearch
```
or
```ts
Pick<SavedSearch, 'name' | 'description' | 'encodedUrl'>

File: lib/content-services/src/lib/common/services/saved-searches.service.spec.ts [hunk 5]
```diff
@@ -72,136 +71,142 @@ describe('SavedSearchesService', () => {
...
+                });
+            });
         });
-    });
 
-    it('should delete a search', (done) => {
-        const searchToDelete = { name: 'Search 1', description: 'Description 1', encodedUrl: 'url1', order: 0 };
-        prepareDefaultMock();
+        it('should edit a search', (done) => {
+            const updatedSearch = { name: 'Search 3', description: 'Description 3', encodedUrl: 'url3', order: 0 };
```
AleksanderSklorz, line 153: please set SavedSearch to that const

lib/content-services/src/lib/common/services/saved-searches.service.spec.ts
dominikiwanekhyland: @AleksanderSklorz why? The type is already enforced by the saveSearch method signature (Pick<SavedSearch, 'name' | 'description' | 'encodedUrl'>), and TS infers type to that const automatically.
Adding an explicit type here would be redundant and doesn’t add additional type safety — if the method signature changes, the compiler will already catch mismatches in the test.

AleksanderSklorz: I was thinking rather for a case if in future this test would extend and be more complex (as logic may always change) and then that const may be used in more places but your point is valid too so I will close that comment and approve PR

PR #11319, opened 2025-10-31
AAE-37713 Amount locale display v.2

tomaszhanaj:

> - [ ] Bugfix
> - [x] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://hyland.atlassian.net/browse/AAE-37713

The idea behind this change is to allow the user to see the amount formatted according to local settings. For example, instead of seeing the number 12345.67, the user will see something like $12,345.67 (this may vary depending on the locale). The current solution for the amount widget is based on using the field.value property to display the field's value. With this PR, this will be changed. The amountWidgetValue property will be used to display the value in the input field. 

The exact value of this property will be based on the state of the input field. If the field is focused, it will show a numeric value, e.g., 12345.67. If the field is blurred, it will be converted into a currency string, e.g., $12,345.67. The widget also subscribes to form changes to respond accordingly to changes in the field.value. Changes can come both from form rules and from other external sources.

> - [x] Yes
> - [ ] No

Parameters of method parseForm from class FormCloudService has been updated and parameter for formService has been added. Currently this method creates FormModel with passing undefined instead of instance of FormService. But instance of FormService is needed to track the changes of field of forms like changes from form rules. 
**Other information**:
Call of FormService method parseForm in FormEditorComponent needs to be updated.

### DIFF
--- a/lib/core/src/lib/form/components/widgets/amount/amount.widget.spec.ts
+++ b/lib/core/src/lib/form/components/widgets/amount/amount.widget.spec.ts
@@ -27,7 +27,6 @@ import { of } from 'rxjs';
 import { FormService } from '../../../services/form.service';
 import { FormFieldEvent } from '../../../events/form-field.event';
 import { TranslationService } from '../../../../translation/translation.service';
-import { AppConfigService } from '../../../../app-config/app-config.service';
 
 describe('AmountWidgetComponent', () => {
     let loader: HarnessLoader;
@@ -38,10 +37,7 @@ describe('AmountWidgetComponent', () => {
     beforeEach(() => {
         TestBed.configureTestingModule({
             imports: [AmountWidgetComponent],
-            providers: [
-                { provide: TranslationService, useValue: { userLang: 'de' } },
-                { provide: AppConfigService, useValue: { getLocale: () => 'en-US' } }
-            ]
+            providers: [{ provide: TranslationService, useValue: { getLocale: () => 'en-US' } }]
         });
 
         fixture = TestBed.createComponent(AmountWidgetComponent);
@@ -126,10 +122,6 @@ describe('AmountWidgetComponent', () => {
     });
 
     it('should set initial values with correct currency', () => {
-        const returnedLanguages: string[] = ['en-GB'];
-        spyOnProperty(window, 'navigator').and.returnValue({
-            languages: returnedLanguages
-        } as any);
         widget.field = new FormFieldModel(null, { id: 2, name: 'test', value: 25, currency: 'GBP' });
         widget.enableDisplayBasedOnLocale = true;
         widget.currency = 'GBP';
@@ -140,10 +132,6 @@ describe('AmountWidgetComponent', () => {
     });
 
     it('should set initial values with correct currency icon', () => {
-        const returnedLanguages: string[] = ['en-GB'];
-        spyOnProperty(window, 'navigator').and.returnValue({
-            languages: returnedLanguages
-        } as any);
         widget.field = new FormFieldModel(null, { id: 2, name: 'test', value: 25, currency: '¥' });
         widget.enableDisplayBasedOnLocale = true;
         widget.currency = '¥';
@@ -154,10 +142,6 @@ describe('AmountWidgetComponent', () => {
     });
 
     it('should set initial values without currency', () => {
-        const returnedLanguages: string[] = ['en-GB'];
-        spyOnProperty(window, 'navigator').and.returnValue({
-            languages: returnedLanguages
-        } as any);
         widget.field = new FormFieldModel(null, { id: 3, name: 'test', value: 25, currency: '' });
         widget.enableDisplayBasedOnLocale = true;
         widget.currency = '';
@@ -490,14 +474,13 @@ describe('AmountWidgetComponent - rendering', () => {
             beforeEach(async () => {
                 TestBed.configureTestingModule({
                     imports: [AmountWidgetComponent],
-                    providers: [{ provide: ADF_AMOUNT_SETTINGS, useValue: of({ enableDisplayBasedOnLocale: true }) }]
+                    providers: [
+                        { provide: ADF_AMOUNT_SETTINGS, useValue: of({ enableDisplayBasedOnLocale: true }) },
+                        { provide: TranslationService, useValue: { getLocale: () => 'en-US' } }
+                    ]
                 });
                 fixture = TestBed.createComponent(AmountWidgetComponent);
                 widget = fixture.componentInstance;
-                const returnedLanguages: string[] = ['en-GB', 'en-US', 'en', 'de-DE', 'pl'];
-                spyOnProperty(window, 'navigator').and.returnValue({
-                    languages: returnedLanguages
-                } as any);
                 fixture.componentRef.setInput('field', mockField);
                 loader = TestbedHarnessEnvironment.loader(fixture);
                 testingUtils = new UnitTestingUtils(fixture.debugElement, loader);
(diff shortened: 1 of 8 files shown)


lib/core/src/lib/form/components/widgets/amount/amount.widget.spec.ts
```diff
@@ -23,6 +23,8 @@ import { FormModel } from '../core/form.model';
 import { HarnessLoader } from '@angular/cdk/testing';
 import { TestbedHarnessEnvironment } from '@angular/cdk/testing/testbed';
 import { UnitTestingUtils } from '../../../../testing/unit-testing-utils';
+import { of } from 'rxjs';
+import { FormFieldEvent, FormService } from '@alfresco/adf-core';
```
DenysVuika: circular import, should be relative path as it's the same lib

lib/core/src/lib/form/components/widgets/amount/amount.widget.spec.ts
```diff
@@ -327,6 +462,315 @@ describe('AmountWidgetComponent - rendering', () => {
             expect(asterisk.textContent).toEqual('*');
         });
     });
+
+    describe('Test widget with ADF_AMOUNT_SETTINGS as observable', () => {
+        beforeEach(() => {
+            TestBed.resetTestingModule();
```
DenysVuika, line 462: should not be required as entire test suite is configured to do that, see https://github.com/Alfresco/alfresco-ng2-components/blob/b16e326ac34b72d8b55f46ded412881529856298/lib/core/test.ts#L26

lib/core/src/lib/form/components/widgets/amount/amount.widget.spec.ts
```diff
@@ -327,6 +462,315 @@ describe('AmountWidgetComponent - rendering', () => {
...
+            mockField.value = '300';
+            formService.formFieldValueChanged.next(new FormFieldEvent(mockField.form, mockField));
+
+            expect(widget.amountWidgetValue).toBe('300');
+        });
+    });
+
+    describe('Test widget with different setting for enableDisplayBasedOnLocale', () => {
+        beforeEach(() => {
+            TestBed.resetTestingModule();
```
DenysVuika, line 630: should not be needed, see https://github.com/Alfresco/alfresco-ng2-components/blob/b16e326ac34b72d8b55f46ded412881529856298/lib/core/test.ts#L26

lib/core/src/lib/form/components/widgets/amount/amount.widget.ts
```diff
@@ -17,21 +17,25 @@
...
+import { takeUntilDestroyed } from '@angular/core/rxjs-interop';
+import { FormFieldEvent } from '../../../events/form-field.event';
 
 export interface AmountWidgetSettings {
     showReadonlyPlaceholder: boolean;
+    enableDisplayBasedOnLocale: boolean;
 }
 
-export const ADF_AMOUNT_SETTINGS = new InjectionToken<AmountWidgetSettings>('adf-amount-settings');
+export const ADF_AMOUNT_SETTINGS = new InjectionToken<Observable<AmountWidgetSettings> | AmountWidgetSettings>('adf-amount-settings');
```
DenysVuika, line 39: I suggest following the pattern of other setting classes and making it a static object, not Observable. In real-life scenarios, is will be just "useValue: obj"

lib/core/src/lib/form/components/widgets/amount/amount.widget.ts
```diff
@@ -49,36 +53,139 @@ export const ADF_AMOUNT_SETTINGS = new InjectionToken<AmountWidgetSettings>('adf
...
     get placeholder(): string {
         return this.showPlaceholder ? this.field.placeholder : '';
     }
 
     constructor(
         public formService: FormService,
-        @Inject(ADF_AMOUNT_SETTINGS)
-        @Optional()
-        private settings: AmountWidgetSettings
+        @Optional() @Inject(ADF_AMOUNT_SETTINGS) settings: Observable<AmountWidgetSettings> | AmountWidgetSettings,
```
DenysVuika, line 83: too complicated imho, should not be observable for plain settings coming from the app level

lib/core/src/lib/form/components/widgets/amount/amount.widget.ts
```diff
@@ -49,36 +53,139 @@ export const ADF_AMOUNT_SETTINGS = new InjectionToken<AmountWidgetSettings>('adf
...
+        private currencyPipe: CurrencyPipe
     ) {
         super(formService);
+        if (isObservable(settings)) {
```
DenysVuika, line 88: will go away if you remove Observable type

lib/core/src/lib/form/components/widgets/amount/amount.widget.ts
```diff
@@ -49,36 +53,139 @@ export const ADF_AMOUNT_SETTINGS = new InjectionToken<AmountWidgetSettings>('adf
...
+    amountWidgetOnFocus(): void {
+        this.isInputInFocus = true;
+        if (this.enableDisplayBasedOnLocale) {
+            const hasValue = this.valueAsNumber === 0 || this.valueAsNumber;
+            this.amountWidgetValue = hasValue ? this.valueAsNumber.toString() : null;
+        }
+    }
+
+    getLocale(): string {
+        const defaultLocale = 'en-US';
```
DenysVuika: Hardcoded value. There are scenarios when customers set different default language (for example, German). So this won't work properly. I suggest getting the user lang from `TranslationService` (https://github.com/Alfresco/alfresco-ng2-components/blob/5d64c7f0ed3d99a4fa6d2b493016768701edf3a7/lib/core/src/lib/translation/translation.service.ts#L54), or checking if we have anything in the `AppConfigService`

lib/process-services-cloud/src/lib/form/services/form-cloud.service.ts
```diff
@@ -211,9 +211,10 @@ export class FormCloudService extends BaseCloudService implements FormCloudServi
      * @param json JSON data to create the form
      * @param data Values for the form's fields
      * @param readOnly Toggles whether the form should be read-only
+     * @param formService Implements Process Services form methods
      * @returns Form created from the JSON specification
      */
-    parseForm(json: any, data?: TaskVariableCloud[], readOnly: boolean = false): FormModel {
+    parseForm(json: any, data?: TaskVariableCloud[], readOnly: boolean = false, formService?: FormService): FormModel {
```
DenysVuika: This can be a private field here instead of the public (potential breaking change).
```
private formService = inject(FormService)
```

lib/core/src/lib/form/components/widgets/amount/amount.widget.spec.ts
tomaszhanaj: fixed

lib/core/src/lib/form/components/widgets/amount/amount.widget.spec.ts
tomaszhanaj: I need to do it because below I am defining testing module for ADF_AMOUNT_SETTINGS as observable.

lib/core/src/lib/form/components/widgets/amount/amount.widget.spec.ts
tomaszhanaj: as above

lib/core/src/lib/form/components/widgets/amount/amount.widget.ts
tomaszhanaj: updated and based on userLang

lib/process-services-cloud/src/lib/form/services/form-cloud.service.ts
tomaszhanaj: updated, I removed parameter for FormService

PR #11324, opened 2025-11-03
AAE-39155 Implement button widget on ADF

wiktord2000:

> - [ ] Bugfix
> - [x] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://hyland.atlassian.net/browse/AAE-39155

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/core/src/lib/form/components/widgets/button/button.widget.scss
+++ b/lib/core/src/lib/form/components/widgets/button/button.widget.scss
@@ -3,8 +3,11 @@
     justify-content: center;
     align-items: center;
     height: 100%;
+    width: 100%;
+    min-height: 96px;
 
     &__button {
+        margin-bottom: 32px;
         word-break: break-word;
         overflow: hidden;
     }


lib/core/src/lib/form/components/widgets/button/button.widget.html
```diff
@@ -0,0 +1,11 @@
+<button
+    mat-flat-button
+    class="adf-button-widget__button"
+    [color]="'primary'"
```
tomgny, line 4: pls check if its needed (for md3 for sure its not, but we are still on md2)

lib/core/src/lib/form/components/widgets/button/button.widget.ts
```diff
@@ -0,0 +1,52 @@
...
+import { TranslatePipe } from '@ngx-translate/core';
+import { MatButtonModule } from '@angular/material/button';
+import { MatTooltipModule } from '@angular/material/tooltip';
+
+@Component({
+    selector: 'button-widget',
+    templateUrl: './button.widget.html',
+    styleUrl: './button.widget.scss',
+    host: {
+        '[class]': 'hostClasses'
```
tomgny, line 32: what about click? maybe we can use @ HostBinding instead of setting it as component metadata?

tomgny (commented):
please check the comments

lib/core/src/lib/form/components/widgets/button/button.widget.html
wiktord2000: Color won't apply without it so far, leaving it

lib/core/src/lib/form/components/widgets/button/button.widget.ts
wiktord2000: Click is attached directly to the button instead of the entire container. I believe other widgets should follow the same but so far we decided to leave for them the same behavior and only reduce the click event calls. @HostBinding will be deprecated

image

tomgny (approved):

acichon-h (approved):
tootlip ✅ 
click event only on the button ✅ 
text doesn't go beyond the button area ✅ 
long names handled ✅  it's not perfect in terms or readability, but that can be fixed with tooltip. the most important aspect - not breaking form layout, is handled well and there are no issues
button without name visible ✅ 
button disabled when readonly ✅ 

QA approved ✅

PR #11330, opened 2025-11-06
[ACS-10300] a11y Fix: visible and accessible Group label is missing 

nikita-web-ua:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation
> - [X] Other... Please describe: a11y fix

https://hyland.atlassian.net/browse/ACS-10300

The radio buttons have a visible group label, and also, the group label announced when the radio buttons are focused for the screen reader user.

> - [ ] Yes
> - [X] No

**Other information**:

### DIFF
--- a/lib/content-services/src/lib/version-manager/version-upload.component.scss
+++ b/lib/content-services/src/lib/version-manager/version-upload.component.scss
@@ -1,30 +1,5 @@
 @use '../mat-selectors.scss' as ms;
 
-.adf-version-type-label {
-    display: block;
-    margin-bottom: 8px;
-}
-
-.adf-new-version-radio-group {
-    display: inline-flex;
-    flex-direction: column;
-}
-
-.adf-new-version-radio-button {
-    margin: 5px;
-    margin-top: 2px;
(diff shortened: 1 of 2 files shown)


lib/content-services/src/lib/version-manager/version-upload.component.scss
```diff
@@ -1,5 +1,10 @@
 @use '../mat-selectors.scss' as ms;
 
+.adf-version-type-label {
```
AleksanderSklorz: can we wrap those all styles in component's selector (because component uses ViewEncapsulation.NONE so to avoid issues with styles in future)?

AleksanderSklorz:
@nikita-web-ua here is some sonar issue. I think that's valid issue because Intellij gives similar error: 
image

lib/content-services/src/lib/version-manager/version-upload.component.scss
nikita-web-ua: done

PR #11386, opened 2025-11-26
[MNT-25408]: updates changing search config logic; exposes stream for active filters

rmnvch:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://hyland.atlassian.net/browse/MNT-25408

Exposes 2 streams to watch filter change; adds Proxy to detect direct property assignment for BaseQueryBuilderService's queryFragments

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/content-services/src/lib/search/services/base-query-builder.service.ts
+++ b/lib/content-services/src/lib/search/services/base-query-builder.service.ts
@@ -80,16 +80,17 @@ export abstract class BaseQueryBuilderService {
     userFacetBucketsUpdate = new BehaviorSubject<{ [key: string]: FacetFieldBucket[] }>({});
 
     categories: SearchCategory[] = [];
-    private _queryFragments: { [id: string]: string } = {};
     filterQueries: FilterQuery[] = [];
     filterRawParams: { [key: string]: any } = {};
     paging: { maxItems?: number; skipCount?: number } = null;
     sorting: SearchSortingDefinition[] = [];
     sortingOptions: SearchSortingDefinition[] = [];
+
     private encodedQuery: string;
     private scope: RequestScope;
     private selectedConfiguration: number;
     private _userQuery = '';
+    private _queryFragments: { [id: string]: string } = {};
 
     private readonly queryFragmentsHandler: ProxyHandler<{ [key: string]: any }> = {
         set: (target: { [key: string]: any }, property: string, value: any) => {


lib/content-services/src/lib/search/services/base-query-builder.service.ts
```diff
@@ -73,8 +73,14 @@ export abstract class BaseQueryBuilderService {
...
 
+    /*  Stream that emits every time queryFragments change */
+    queryFragmentsUpdate = new BehaviorSubject<{ [key: string]: any }>({});
+
+    /*  Stream that emits every time userFacetBuckets change */
+    userFacetBucketsUpdate = new BehaviorSubject<{ [key: string]: FacetFieldBucket[] }>({});
+
     categories: SearchCategory[] = [];
-    queryFragments: { [id: string]: string } = {};
+    _queryFragments: { [id: string]: string } = {};
```
AleksanderSklorz: should be private - when we have getters and setters then original field should be private with _ prefix. Please also separate private from public fields then

lib/content-services/src/lib/search/services/base-query-builder.service.ts
```diff
@@ -615,4 +633,16 @@ export abstract class BaseQueryBuilderService {
             queryParamsHandling: 'merge'
         });
     }
+
+    private readonly queryFragmentsHandler: ProxyHandler<{ [key: string]: any }> = {
```
AleksanderSklorz: Similar comment for usages for any in that function: 
https://github.com/Alfresco/alfresco-ng2-components/pull/11386/files#r2567686029

lib/content-services/src/lib/search/services/base-query-builder.service.ts
```diff
@@ -73,8 +73,14 @@ export abstract class BaseQueryBuilderService {
     /*  Stream that emits the initial value for some or all search filters */
     populateFilters = new BehaviorSubject<{ [key: string]: any }>({});
 
+    /*  Stream that emits every time queryFragments change */
+    queryFragmentsUpdate = new BehaviorSubject<{ [key: string]: any }>({});
```
AleksanderSklorz: similar comment for any: 
https://github.com/Alfresco/alfresco-ng2-components/pull/11386/files#r2567686029

lib/content-services/src/lib/search/services/base-query-builder.service.ts
AleksanderSklorz: queryFragmentsHandler is private readonly field. It's not a function. So please move it above class to place where all fields are placed

lib/content-services/src/lib/search/services/search-query-builder.service.spec.ts
```diff
@@ -15,6 +15,8 @@
  * limitations under the License.
  */
 
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
AleksanderSklorz: This will hide eslint issues for any in whole file so we won't see any problems here. Instead adding that comment, could you do following:
Create separate task in jira to replace any with correct types in old code of that file

Then it will be better to fix those places rather than hide problems. 

If you prefer you can fix old places in that PR but I'm fine to create separate task for that as well.

lib/content-services/src/lib/search/services/base-query-builder.service.ts
```diff
@@ -87,6 +93,15 @@ export abstract class BaseQueryBuilderService {
 
     protected userFacetBuckets: { [key: string]: FacetFieldBucket[] } = {};
 
+    get queryFragments(): { [key: string]: any } {
+        return this._queryFragments;
+    }
+
+    set queryFragments(value: { [key: string]: any }) {
```
AleksanderSklorz, line 109: Two things: 
1. Can we set string instead of any because I see that queryFragments field has string as value type (right side)? If it works then it would be better option
2. If no then see following
Any here and in setter below give eslint issue. I know you don't know what the type will be as they can have any type but can we use different type than any like "uknown" or "never"? They are a bit safer alternatives 
https://rahuulmiishra.medium.com/typescripts-any-vs-unknown-vs-never-when-and-how-to-use-them-35d81ea57c01
image

Would it have sense to use one of those types?

lib/content-services/src/lib/search/services/base-query-builder.service.ts
```diff
@@ -87,6 +93,15 @@ export abstract class BaseQueryBuilderService {
 
     protected userFacetBuckets: { [key: string]: FacetFieldBucket[] } = {};
 
+    get queryFragments(): { [key: string]: any } {
```
AleksanderSklorz, line 105: I believe here we can have just string instead of any because queryFragments field has string too there and we just return that field

lib/content-services/src/lib/search/services/base-query-builder.service.ts
AleksanderSklorz, line 77: here we should have just string instead of any because it emits queryFragments in next function and queryFragments has string there instead of any

lib/content-services/src/lib/search/services/base-query-builder.service.ts
AleksanderSklorz: See comment, please check both points: 
https://github.com/Alfresco/alfresco-ng2-components/pull/11386/files#r2567745878

lib/content-services/src/lib/search/services/base-query-builder.service.ts
```diff
@@ -615,4 +633,16 @@ export abstract class BaseQueryBuilderService {
...
+
+    private readonly queryFragmentsHandler: ProxyHandler<{ [key: string]: any }> = {
+        set: (target: { [key: string]: any }, property: string, value: any) => {
+            target[property as keyof typeof target] = value;
+            this.queryFragmentsUpdate.next(this._queryFragments);
+            return true;
+        }
+    };
+
+    private createQueryFragmentsProxy(target: { [key: string]: any }): { [key: string]: any } {
```
AleksanderSklorz, line 646: See comment, please check both points: 
https://github.com/Alfresco/alfresco-ng2-components/pull/11386/files#r2567745878

lib/content-services/src/lib/search/services/search-query-builder.service.spec.ts
rmnvch: @AleksanderSklorz, agree. I didn't mean to leave it there just forgot to revert it back, thanks for pointing that out.

lib/content-services/src/lib/search/services/base-query-builder.service.ts
rmnvch: @AleksanderSklorz , minding that it's a library and type _might be_ other than string, "unknown" would be the right choice. I've checked and all over the code inside of ADF we just once assign undefined there instead of string and that wouldn't be a problem. The problem I see is that we cannot just change it to 'unknown' because it would be a breaking change - some consumers would need type narrowing now, so I believe it should be supported by a sort of migration guide. Probably it should be a dedicated ticket for that

lib/content-services/src/lib/search/services/search-query-builder.service.spec.ts
rmnvch: ticket's created 
https://hyland.atlassian.net/browse/ACS-10797

lib/content-services/src/lib/search/services/base-query-builder.service.ts
```diff
@@ -73,8 +73,14 @@ export abstract class BaseQueryBuilderService {
...
 
+    /*  Stream that emits every time queryFragments change */
+    queryFragmentsUpdate = new BehaviorSubject<{ [key: string]: any }>({});
+
+    /*  Stream that emits every time userFacetBuckets change */
+    userFacetBucketsUpdate = new BehaviorSubject<{ [key: string]: FacetFieldBucket[] }>({});
+
     categories: SearchCategory[] = [];
-    queryFragments: { [id: string]: string } = {};
+    private _queryFragments: { [id: string]: string } = {};
```
AleksanderSklorz: can you group all private fields together?

lib/content-services/src/lib/search/services/base-query-builder.service.ts
AleksanderSklorz: Let me ask Michał what steps we should do in that case. @MichalKinas can you help us make decision? You can check that comment and open comments below which are related

MichalKinas: Unknow may throw some TS issues when compared with any other type so I would avoid it in that case, since it can take some other types except of string I think any usage would be justified here, we can think of suppresing the eslint warning there too

MichalKinas (approved):

PR #11781, opened 2026-03-30
AAE-28918 Fix for start process button is enabled when required widgets are read only without value

dthornton-hyl:

> - [ x] The commit message follows our [guidelines](https://github.com/Alfresco/alfresco-ng2-components/wiki/Commit-format)
> - [ x] Tests for the changes have been added (for bug fixes / features)

> - [ x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://hyland.atlassian.net/browse/AAE-28918
Start button is enable if all fields are read-only and required. Clicking Start button will return validation error from BE.

Start button is disable if all fields are read only and required

> - [ ] Yes
> - [ x] No

**Other information**:

### DIFF
--- a/lib/core/src/lib/form/components/widgets/core/form-field.model.ts
+++ b/lib/core/src/lib/form/components/widgets/core/form-field.model.ts
@@ -170,24 +170,16 @@ export class FormFieldModel extends FormWidgetModel {
     validate(): boolean {
         this.validationSummary = new ErrorMessageModel();
 
-        if (this.isFieldValidatable()) {
-            const validators = this.form.fieldValidators || [];
-            for (const validator of validators) {
-                if (!validator.validate(this)) {
-                    this._isValid = false;
-                    return this._isValid;
-                }
+        const validators = this.form?.fieldValidators || [];
+        for (const validator of validators) {
+            if (!validator.validate(this)) {
+                this._isValid = false;
+                return this._isValid;
             }
-        } else if (this.readOnly && this.required && !this.form?.isFieldOrParentHidden(this) && this.isValueEmpty()) {
-            this._isValid = false;
-            return this._isValid;
         }
+
         this._isValid = true;
         return this._isValid;
-    }
-
-    private isFieldValidatable(): boolean {
-        return !this.readOnly || FormFieldTypes.isValidatableType(this.type);
     }
 
     isValueEmpty(): boolean {
--- a/lib/core/src/lib/form/components/widgets/core/form-field.model.ts
+++ b/lib/core/src/lib/form/components/widgets/core/form-field.model.ts
@@ -178,7 +178,7 @@ export class FormFieldModel extends FormWidgetModel {
                     return this._isValid;
                 }
             }
-        } else if (this.readOnly && this._required && !this.form?.isFieldOrParentHidden(this) && this.isValueEmpty()) {
+        } else if (this.readOnly && this.required && !this.form?.isFieldOrParentHidden(this) && this.isValueEmpty()) {
             this._isValid = false;
             return this._isValid;
         }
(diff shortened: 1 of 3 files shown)


lib/core/src/lib/form/components/widgets/core/form-field.model.ts
```diff
@@ -187,6 +190,22 @@ export class FormFieldModel extends FormWidgetModel {
         return !this.readOnly || FormFieldTypes.isValidatableType(this.type);
```
sdonnell-hyland: I'm surprised we don't have a `isFieldOrParentHidden` check here - if it is correct/valid to check for field visibility as part of isFieldValidatable I would probably refactor things a bit, but my next question is should we only be doing the required check validation for readonly fields, or should all validation be ran?

sdonnell-hyland: And personal opinion (I know you didn't write this code), but I would rename this something like `shouldValidate` instead of `isFieldValidatable`

lib/core/src/lib/form/components/widgets/core/form-field.model.ts
```diff
@@ -187,6 +190,22 @@ export class FormFieldModel extends FormWidgetModel {
...
+        if (typeof this.value === 'string' && this.value.trim().length === 0) {
+            return true;
+        }
+        if (this.type === FormFieldTypes.BOOLEAN) {
+            return !this.value;
+        }
+        if (Array.isArray(this.value)) {
+            return this.value.length === 0;
+        }
+        return false;
```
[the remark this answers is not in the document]
dthornton-hyl: fixed

lib/core/src/lib/form/components/widgets/core/form-field.model.ts
dthornton-hyl: I will look into it. Will convert this to  a draft PR for the time being.

csandham (approved):

lib/core/src/lib/form/components/widgets/core/form-field.model.ts
dthornton-hyl: `FormFieldTypes.isValidatableType` just had one field type in it. So `isFieldValidatable()` was pretty much returning if a field was readOnly or not. Since want to validate either readOnly or not. Can pretty much do away with the isFieldValidatable() check. 

The `validator.validate()` has checks to see if the field is validation is supported (so making the isFieldValidatable check not needed) and it also checks the `isFieldOrParentHidden`

lib/core/src/lib/form/components/widgets/core/form-field.model.ts
```diff
@@ -170,23 +170,18 @@ export class FormFieldModel extends FormWidgetModel {
...
 
-        if (this.isFieldValidatable()) {
-            const validators = this.form.fieldValidators || [];
-            for (const validator of validators) {
-                if (!validator.validate(this)) {
-                    this._isValid = false;
-                    return this._isValid;
-                }
+        const validators = this.form?.fieldValidators || [];
+        for (const validator of validators) {
```
sdonnell-hyland, line 174: I'm good with this change, but lets run it by product or the Automate Front End channel.  It seems reasonable to me to require all fields to be valid regardless of being editable.  

`FormFieldTypes.isValidatableType` is kind of an odd check to me because I don't know what a `display-external-property` type field is, which appears to be the only type it's checking for, but want to confirm the removal of that check was intentional?

dthornton-hyl: Also about the isValidateType, kind of mention in comment above. It was intentional. Also, I can't just add types but I would pretty much add all types to it. But there will be types that it wouldn't know of (hxp_upload or custom fields). It looks like the validators them itself has a isSupported for the field type. So, I don't think the isValidateType is need, at least as I understand it.

DavidOlson-Hyland (approved):

jcain82 (approved):

sdonnell-hyland (approved):

PR #11837, opened 2026-04-24
AAE-44435 Fix incorrect Start process button on user task forms

amolodyh-hyland:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

On an active user task form in Automate Workspace, an extra **Start process** button intermittently appears alongside `Cancel`, `Save`, and `Submit` after a browser refresh. Clicking it wrongly completes the task. The same ghost button is also visible on completed (read-only) task forms (only `Cancel` is enabled, but the extra buttons are still rendered).

Root cause (three layered issues):

1. `FormModel.parseOutcomes()` always appends a system `START_PROCESS` outcome when no custom outcomes are defined, and `FormOutcomeModel.isVisible` defaults to `true`.
2. The cloud form template rendered every outcome in the DOM and relied solely on an `adf-form-hide-button` CSS class (`display: none`) to visually suppress `START PROCESS`. That class is declared only inside `form-renderer.component.scss` (leaks globally via `ViewEncapsulation.None`), so it races with initial paint on refresh — sometimes the button is visible before the style is applied.
3. `isOutcomeButtonVisible` short-circuited on `isFormReadOnly` **before** the `START_PROCESS_ACTION` check, so on completed task forms the button was never marked hidden at all.

Issue: https://hyland.atlassian.net/browse/AAE-44435

- **Task forms no longer carry a `START_PROCESS` outcome.** `FormCloudComponent.parseForm()` now strips `START_PROCESS_OUTCOME` from `form.outcomes` whenever the form is bound to a task (`this.taskId` is set). Start forms (process start) are unaffected.
- **Outcome rendering uses a pre-computed, filtered list.** The template iterates over `visibleOutcomes` (buttons that pass both `outcome.isVisible` and `isOutcomeButtonVisible(...)`) instead of rendering every outcome and hiding some via CSS. This removes the refresh-time race — the Start Process button is no longer in the DOM when it should not be shown.
- **`isOutcomeButtonVisible` returns `false` for `START_PROCESS_ACTION` before the read-only short-circuit.** Completed task forms no longer show the extra buttons.
- `visibleOutcomes` is recomputed when the form is set, when `readOnly` / `showCompleteButton` / `showSaveButton` change, when field values change, and on visibility refresh — no function calls in the template.

Tests added:
- `buttons-visibility.spec.ts` (new) — unit tests for the helper, including the read-only + START_PROCESS case.
- `form-cloud.component.spec.ts` — asserts `START_PROCESS_OUTCOME` is removed from task forms and that `visibleOutcomes` is populated/cleared correctly and never contains `START PROCESS`.

> - [ ] Yes
> - [x] No

Behaviour change is scoped to task forms only. Start-process forms are unaffected. The public API of `FormCloudComponent`, `FormModel`, and `FormOutcomeModel` is unchanged. Consumers iterating `form.outcomes` on task forms will no longer see a `START_PROCESS` entry — this matches the documented behaviour (task forms never show a Start Process button), but it is a shape change worth noting in review.

**Other information**:

Issue: https://hyland.atlassian.net/browse/AAE-44435

Files changed:
- `lib/core/src/lib/form/components/helpers/buttons-visibility.ts` — reorder check so `START_PROCESS_ACTION` is hidden before the read-only short-circuit.
- `lib/core/src/lib/form/components/helpers/buttons-visibility.spec.ts` — new unit tests.
- `lib/process-services-cloud/src/lib/form/components/form-cloud.component.ts` — strip `START_PROCESS_OUTCOME` from task forms in `parseForm`; add `visibleOutcomes` + `recomputeVisibleOutcomes`; wire recompute into the `form` accessor, `ngOnChanges`, `formFieldValueChanged`, and `checkVisibility`.
- `lib/process-services-cloud/src/lib/form/components/form-cloud.component.html` — iterate `visibleOutcomes` instead of every outcome; drop the `adf-form-hide-button` CSS-based hiding.
- `lib/process-services-cloud/src/lib/form/components/form-cloud.component.spec.ts` — coverage for the new behaviour.

Follow-up candidates (not in this PR):
- Consider removing the `override get form`/`set form` accessor in `FormCloudComponent` in favour of explicit `recomputeVisibleOutcomes()` calls at the 3 internal assignment sites — the setter side-effect is a hidden coupling point.
- Update `docs/features/maps/outcome-buttons-functionality.map.md` (Visibility Determination + Component Tree sections, Change Log entry) once this lands.

### DIFF
--- a/docs/process-services-cloud/components/form-cloud.component.md
+++ b/docs/process-services-cloud/components/form-cloud.component.md
@@ -89,7 +89,7 @@ The template defined inside `empty-form` will be shown when no form definition i
 | path | `string` |  | Path of the folder where the metadata will be stored. |
 | processInstanceId | `string` |  | ProcessInstanceId id to fetch corresponding form and values. |
 | readOnly | `boolean` | false | Toggle readonly state of the form. Forces all form widgets to render as readonly if enabled. |
-| showCompleteButton | `boolean` | true | Toggle rendering of the `Complete` outcome button. |
+| showCompleteButton | `boolean` | false | Toggle rendering of the `Complete` outcome button. |
 | showRefreshButton | `boolean` | true | Toggle rendering of the `Refresh` button. |
 | showSaveButton | `boolean` | true | Toggle rendering of the `Save` outcome button. |
 | showTitle | `boolean` | true | Toggle rendering of the form title. |
--- a/lib/core/src/lib/form/components/helpers/buttons-visibility.spec.ts
+++ b/lib/core/src/lib/form/components/helpers/buttons-visibility.spec.ts
@@ -1,0 +1,63 @@
+/*!
+ * @license
+ * Copyright © 2005-2025 Hyland Software, Inc. and its affiliates. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { FormOutcomeModel } from '../widgets';
+import { isOutcomeButtonVisible } from './buttons-visibility';
+
+describe('isOutcomeButtonVisible', () => {
+    const defaultProps = {
+        isFormReadOnly: false,
+        showCompleteButton: true,
+        showSaveButton: true
+    };
+
+    const outcome = (overrides: Partial<FormOutcomeModel>): FormOutcomeModel => ({ name: '', isSelected: false, ...overrides }) as FormOutcomeModel;
+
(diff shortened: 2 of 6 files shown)


DavidOlson-Hyland (approved):

dthornton-hyl (approved):

docs/process-services-cloud/components/form-cloud.component.md
```diff
@@ -89,7 +89,7 @@ The template defined inside `empty-form` will be shown when no form definition i
 | path | `string` |  | Path of the folder where the metadata will be stored. |
 | processInstanceId | `string` |  | ProcessInstanceId id to fetch corresponding form and values. |
 | readOnly | `boolean` | false | Toggle readonly state of the form. Forces all form widgets to render as readonly if enabled. |
-| showCompleteButton | `boolean` | true | Toggle rendering of the `Complete` outcome button. |
+| showCompleteButton | `boolean` | false | Toggle rendering of the `Complete` outcome button. |
```
sdonnell-hyland, line 92: double checking - this was just incorrect documentation?

sdonnell-hyland (approved):

docs/process-services-cloud/components/form-cloud.component.md
amolodyh-hyland: Yeah. It was changed to false as the default about 9 months ago but the documentation was never updated.

PR #11926, opened 2026-05-29
[AAE-46514] -  Migration to PNPM

VitoAlbano:
… for those who are trusted

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

> - [ ] Yes
> - [ ] No

**Other information**:


.github/workflows/pull-request.yml
```diff
@@ -46,12 +46,18 @@ jobs:
       - name: Ensure SHA pinned actions
         uses: hyland/github-actions-ensure-sha-pinned-actions@7d494b6b53e71f75194d74b4dd890a2266d060b5 # v2.0.0
 
-      - name: Check package-lock.json version
+      - name: Check pnpm-lock.yaml version
```
DenysVuika, line 49: we can probably drop this stage for pnpm, it was for npm when we switched between major versions

.github/workflows/pull-request.yml
```diff
@@ -174,13 +180,10 @@ jobs:
...
         with:
           cache-suffix: setup
-      - name: Install dependencies
-        if: ${{ steps.setup-env.outputs.node-modules-cache-hit != 'true' }}
-        run: npm ci
       - name: Bundle
         run: |
-          npm run bundle:js-api
-          npm run bundle:cli
+          pnpm run bundle:js-api
```
DenysVuika: you can omit the `run` with pnpm, so just `pn bundle:js-api`

package.json.tmp
DenysVuika, line 1: is this file needed?

lib/eslint-angular/main.js
```diff
@@ -1,8 +1 @@
-require('ts-node').register({
```
DenysVuika, line 1: that was the only case for using ts-node, can you please remove the dependency as well?

.github/workflows/pull-request.yml
VitoAlbano: Hmm it's checking the consistency of the version used, it's not a big issue for now , i'll remove it in a second update

.github/actions/setup/action.yml
```diff
@@ -24,20 +19,19 @@ outputs:
...
-    value: ${{ steps.node-modules-cache.outputs.cache-hit }}
 runs:
   using: "composite"
   steps:
       - name: Set consistent machine ID for Nx cache
         shell: bash
         run: echo "nx-github-actions" | sudo tee /etc/machine-id > /dev/null
-      - name: install NPM
+      - name: Setup pnpm
+        uses: pnpm/action-setup@b906affcce14559ad1aafd4ab0e942779e9f58b1 # v4.3.0
```
DenysVuika, line 29: maybe worth also pinning some version here, like 11, and enabling cache?
https://github.com/pnpm/action-setup#use-cache-to-reduce-installation-time

.github/workflows/pull-request.yml
```diff
@@ -203,13 +206,10 @@ jobs:
...
           cache-suffix: lint
           full-setup: 'false'
-      - name: Install dependencies
-        if: ${{ steps.setup-env.outputs.node-modules-cache-hit != 'true' }}
-        run: npm ci
       - name: Run lint
         env:
           BASE_REF: ${{ github.base_ref || 'develop' }}
-        run: npx nx affected --target=lint --base=origin/$BASE_REF --head=HEAD
+        run: pnpm nx affected --target=lint --base=origin/$BASE_REF --head=HEAD
```
DenysVuika, line 212: it should be `pnx` (https://pnpm.io/cli/pnx)

.github/workflows/pull-request.yml
```diff
@@ -239,14 +239,11 @@ jobs:
...
           full-setup: 'false'
-      - name: Install dependencies
-        if: ${{ steps.setup-env.outputs.node-modules-cache-hit != 'true' }}
-        run: npm ci
       - name: Build Storybook
         env:
           BASE_REF: ${{ github.base_ref || 'develop' }}
         run: |
-          npx nx affected --target=build-storybook --base=origin/$BASE_REF --head=HEAD --configuration=ci
+          pnpm nx affected --target=build-storybook --base=origin/$BASE_REF --head=HEAD --configuration=ci
```
DenysVuika, line 246: it should be `pnx` (https://pnpm.io/cli/pnx)

.github/workflows/unit-test-workflow.yml
```diff
@@ -26,16 +26,13 @@ jobs:
...
-        if: ${{ steps.setup-env.outputs.node-modules-cache-hit != 'true' }}
-        run: npm ci
       - name: Generate affected projects matrix
         id: set-matrix
         env:
           BASE_REF: ${{ inputs.base_ref }}
         run: |
           echo "Base ref is $BASE_REF"
-          AFFECTED_UNIT=$(npx nx show projects --affected --target=test --base=origin/$BASE_REF --head=HEAD --select=projects --plain --exclude=cli,stories,eslint-angular)
+          AFFECTED_UNIT=$(pnpm nx show projects --affected --target=test --base=origin/$BASE_REF --head=HEAD --select=projects --plain --exclude=cli,stories,eslint-angular)
```
DenysVuika, line 35: it should be `pnx` (https://pnpm.io/cli/pnx)

.github/workflows/unit-test-workflow.yml
```diff
@@ -72,14 +69,11 @@ jobs:
...
           full-setup: 'false'
-      - name: Install dependencies
-        if: ${{ steps.setup-env.outputs.node-modules-cache-hit != 'true' }}
-        run: npm ci
       - name: Run unit tests for ${{ matrix.project }}
         env:
           NODE_OPTIONS: "--max-old-space-size=5120"
         run: |
-          xvfb-run --auto-servernum npx nx run ${{ matrix.project }}:test
+          xvfb-run --auto-servernum pnpm nx run ${{ matrix.project }}:test
```
DenysVuika, line 76: it should be `pnx` (https://pnpm.io/cli/pnx)

.github/workflows/build-lib-workflow.yml
```diff
@@ -28,13 +28,10 @@ jobs:
         with:
           cache-suffix: build
           full-setup: 'false'
-      - name: Install dependencies
-        if: ${{ steps.setup-env.outputs.node-modules-cache-hit != 'true' }}
-        run: npm ci
```
DenysVuika, line 33: maybe still worth running `pn install --frozen-lockfile` here?

VitoAlbano: fixed

.github/workflows/pull-request.yml
VitoAlbano: `Aliases: pnpm dlx, pnpx

Fetches a package from the registry without installing it as a dependency, hotloads it, and runs whatever default command binary it exposes` -> This means that it downloads always the version (most probably the last) and it will basically invalidate the cache system as it won't be restored so we download this every time. I think for CI use case it works better to have it in the deps so we can cache / restore it

.github/actions/setup/action.yml
VitoAlbano: It is already cached and the packageManager hold the exact version "packageManager": "pnpm@11.5.0", so we should be fine here

package.json
```diff
@@ -4,7 +4,8 @@
   "version": "8.5.0",
   "author": "Hyland Software, Inc. and its affiliates",
   "scripts": {
-    "prepare": "husky",
+    "preinstall": "npx only-allow pnpm",
+    "prepare": "husky && nx run eslint-angular:build && pnpm audit --audit-level=critical",
```
DenysVuika, line 8: should be "pn eslint-angular:build"

package.json
```diff
@@ -4,7 +4,8 @@
...
     "bundle:js-api": "nx run js-api:bundle",
```
DenysVuika, line 9: should be "pn js-api:bundle"

DenysVuika (approved):

PR #11987, opened 2026-06-15
[ACS-11928] convert ReadStream to Blob before appending to FormData

adamswiderskipw:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://hyland.atlassian.net/browse/ACS-11928

Before, the raw ReadStream was appended directly to FormData (which Node 24 can't serialize); now the file is read into a Blob first.

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/js-api/src/fetchHttpClient.ts
+++ b/lib/js-api/src/fetchHttpClient.ts
@@ -367,11 +367,11 @@ export class FetchHttpClient implements HttpClient {
             const normalizedParams = FetchHttpClient.normalizeParams(formParams);
             const formData = new FormData();
             for (const [key, value] of Object.entries(normalizedParams)) {
-                const { blob, filename } = FetchHttpClient.toFormDataValue(value);
+                const { data, filename } = FetchHttpClient.toFormDataValue(value);
                 if (filename) {
-                    formData.append(key, blob, filename);
+                    formData.append(key, data, filename);
                 } else {
-                    formData.append(key, blob);
+                    formData.append(key, data);
                 }
             }
             return formData;
@@ -530,23 +530,23 @@ export class FetchHttpClient implements HttpClient {
         return false;
     }
 
-    private static toFormDataValue(value: any): { blob: any; filename?: string } {
+    private static toFormDataValue(value: any): { data: any; filename?: string } {
         if (value && typeof value === 'object' && value.path && typeof value.path === 'string' && !(value instanceof Blob)) {
             try {
                 const nodeFs = Function('return require("fs")')();
                 const nodePath = Function('return require("path")')();
                 const buffer = nodeFs.readFileSync(value.path);
                 const filename: string = nodePath.basename(value.path);
-                return { blob: new Blob([buffer]), filename };
+                return { data: new Blob([buffer]), filename };
             } catch {
-                return { blob: value };
+                return { data: value };
             }
         }
 
         if (typeof Buffer === 'function' && value instanceof Buffer) {
-            return { blob: new Blob([value]) };
+            return { data: new Blob([value]) };
         }
 
-        return { blob: value };
+        return { data: value };
     }
 }
--- a/lib/js-api/src/fetchHttpClient.ts
+++ b/lib/js-api/src/fetchHttpClient.ts
@@ -367,7 +367,12 @@ export class FetchHttpClient implements HttpClient {
             const normalizedParams = FetchHttpClient.normalizeParams(formParams);
             const formData = new FormData();
             for (const [key, value] of Object.entries(normalizedParams)) {
-                formData.append(key, value as any);
+                const { blob, filename } = FetchHttpClient.toFormDataValue(value);
+                if (filename) {
+                    formData.append(key, blob, filename);
+                } else {
+                    formData.append(key, blob);
+                }
             }
             return formData;
         }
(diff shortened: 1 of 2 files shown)


lib/js-api/src/fetchHttpClient.ts
```diff
@@ -524,4 +529,24 @@ export class FetchHttpClient implements HttpClient {
...
+                return { blob: value };
+            }
+        }
+
+        if (typeof Buffer === 'function' && value instanceof Buffer) {
+            return { blob: new Blob([value]) };
+        }
+
+        return { blob: value };
+    }
```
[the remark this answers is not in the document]
adamswiderskipw: Verdict: Won't-fix / out of scope. Acknowledge it as a known trade-off and move on.

TL;DR: Both comments are technically correct observations but neither represents a real-world bug or security vulnerability in the context of this code. They're safe to dismiss with brief explanations on the PR.

lib/js-api/src/fetchHttpClient.ts
```diff
@@ -524,4 +529,24 @@ export class FetchHttpClient implements HttpClient {
...
     }
+
+    private static toFormDataValue(value: any): { blob: any; filename?: string } {
+        if (value && typeof value === 'object' && value.path && typeof value.path === 'string' && !(value instanceof Blob)) {
+            try {
+                const nodeFs = Function('return require("fs")')();
+                const nodePath = Function('return require("path")')();
+                const buffer = nodeFs.readFileSync(value.path);
+                const filename: string = nodePath.basename(value.path);
+                return { blob: new Blob([buffer]), filename };
```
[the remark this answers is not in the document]
adamswiderskipw: Verdict: Won't-fix / out of scope. Acknowledge it as a known trade-off and move on.

TL;DR: Both comments are technically correct observations but neither represents a real-world bug or security vulnerability in the context of this code. They're safe to dismiss with brief explanations on the PR.

lib/js-api/src/fetchHttpClient.ts
```diff
@@ -524,4 +529,24 @@ export class FetchHttpClient implements HttpClient {
...
+            } catch {
+                return { blob: value };
```
AleksanderSklorz: should we really call that field as 'blob'? If you check 'if' condition in 534th line then it enters when value is not Blob so here is that really correct that we assign value as blob field or should we rename it to something different?

lib/js-api/src/fetchHttpClient.ts
```diff
@@ -524,4 +529,24 @@ export class FetchHttpClient implements HttpClient {
         }
         return false;
     }
+
+    private static toFormDataValue(value: any): { blob: any; filename?: string } {
```
AleksanderSklorz: field is called 'blob' so if that is really blob then we can set type to Blob. If that is not blob then we should not name field as blob I guess

lib/js-api/src/fetchHttpClient.ts
```diff
@@ -524,4 +529,24 @@ export class FetchHttpClient implements HttpClient {
...
+        if (value && typeof value === 'object' && value.path && typeof value.path === 'string' && !(value instanceof Blob)) {
+            try {
+                const nodeFs = Function('return require("fs")')();
```
AleksanderSklorz, line 536: Instead workarounds with Function here and line below, can you instead just import required types?

import fs from 'fs';
import path from 'path';

then you can use it as 
fs.readFileSync(value.path);
path.basename(value.path);

adamswiderskipw: Checked with copilot for validity just to be sure and got a response to leave it as it is:

```
The `Function('return require("fs")')()` pattern is intentional, not a workaround.

This file runs in **both browser and Node.js**. A static `import fs from 'fs'` would:

1. **Break Angular/browser builds** — bundlers (webpack/esbuild) would try to resolve the `fs` and `path` Node.js modules, either failing or requiring polyfill configuration
2. **Bloat the browser bundle** with unnecessary polyfills

The `Function('return require(...)')()` trick hides the `require` from static analysis by bundlers, so it's only evaluated at runtime. In Node.js it works; in browsers the `try/catch` catches the failure and falls through gracefully.

This is a well-known pattern for conditional Node.js module loading in isomorphic code.
```

lib/js-api/src/fetchHttpClient.ts
AleksanderSklorz, line 1: I see that file has corresponding unit tests file (fetchHttpClient.spec.ts). Could you please try to add/edit unit tests for this case because for our source code we try to always align unit tests?

adamswiderskipw: I added new unit tests, but please check them as I'm not very experienced with Jest

lib/js-api/src/fetchHttpClient.ts
AleksanderSklorz: ok makes sense

lib/js-api/test/fetchHttpClient.spec.ts
```diff
@@ -1097,4 +1099,67 @@ describe('FetchHttpClient', () => {
             ]);
         });
     });
+
+    describe('#toFormDataValue', () => {
+        const toFormDataValue = (FetchHttpClient as any).toFormDataValue;
```
AleksanderSklorz: We don't test private things like that. We test private logic through public functions. So for example if you want to test logic of private toFormDataValue function then check in which public place it's used. I can see it's used in buildBody function which is private too. So checking further - I see it's called in public request function. 

So please refactor your tests to test it by calling request function. 

We have even that rule in ts.instructions for copilot 

Image

lib/js-api/test/fetchHttpClient.spec.ts
```diff
@@ -1097,4 +1099,67 @@ describe('FetchHttpClient', () => {
...
+        let originalFunction: typeof global.Function;
+
+        beforeEach(() => {
+            originalFunction = global.Function;
+            global.Function = ((code: string) => {
+                if (code.includes('require')) {
+                    return () => require(code.match(/"([^"]+)"/)?.[1] || '');
+                }
+                return originalFunction(code);
+            }) as any;
```
AleksanderSklorz: instead any you can use FunctionConstructor type

AleksanderSklorz:
@adamswiderskipw there is sonar issue, can you check?
image

adamswiderskipw:
@AleksanderSklorz sonar cloud issue fixed and unit tests updated

AleksanderSklorz (approved):

PR #12023, opened 2026-07-01
[ACS-12098] firefox headless race condition viewer fix

adamswiderskipw:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://hyland.atlassian.net/browse/ACS-12098

no race condition for firefox headless CI e2e runs in viewer

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/content-services/src/lib/viewer/components/alfresco-viewer.component.spec.ts
+++ b/lib/content-services/src/lib/viewer/components/alfresco-viewer.component.spec.ts
@@ -32,7 +32,8 @@ import {
     ViewerMoreActionsComponent,
     ViewerToolbarActionsComponent,
     NoopAuthModule,
-    NoopTranslateModule
+    NoopTranslateModule,
+    UnitTestingUtils
 } from '@alfresco/adf-core';
 import { NodesApiService } from '../../common/services/nodes-api.service';
 import { UploadService } from '../../common/services/upload.service';
@@ -177,6 +178,7 @@ describe('AlfrescoViewerComponent', () => {
     let renditionService: RenditionService;
     let viewUtilService: ViewUtilService;
     let nodeActionsService: NodeActionsService;
+    let testingUtils: UnitTestingUtils;
 
     beforeEach(() => {
         TestBed.configureTestingModule({
@@ -197,6 +199,7 @@ describe('AlfrescoViewerComponent', () => {
         fixture = TestBed.createComponent(AlfrescoViewerComponent);
         element = fixture.nativeElement;
         component = fixture.componentInstance;
+        testingUtils = new UnitTestingUtils(fixture.debugElement);
         uploadService = TestBed.inject(UploadService);
         nodesApiService = TestBed.inject(NodesApiService);
         dialog = TestBed.inject(MatDialog);
@@ -398,12 +401,16 @@ describe('AlfrescoViewerComponent', () => {
         component.nodeId = 'id1';
         component.showViewer = true;
         component.versionId = null;
-        component.ngOnChanges(getSimpleChanges('id1'));
 
-        await fixture.whenStable();
+        await fixture.ngZone.run(async () => {
+            component.ngOnChanges(getSimpleChanges('id1'));
+            await fixture.whenStable();
+        });
+        fixture.detectChanges();
 
-        expect(component.fileName).toBe('file1.pdf');
-        expect(component.blobFileContent).toBe(mockBlob);
+        const viewer = testingUtils.getByDirective(ViewerComponent).componentInstance as ViewerComponent<unknown>;
+        expect(viewer.fileName).toBe('file1.pdf');
+        expect(viewer.blobFile).toBe(mockBlob);
     });
 
     it('should change display name every time node`s version changes', fakeAsync(() => {
@@ -1080,13 +1087,17 @@ describe('AlfrescoViewerComponent', () => {
 
             component.nodeId = 'node-1';
             component.showViewer = true;
-            component.ngOnChanges(getSimpleChanges('node-1'));
 
-            await fixture.whenStable();
+            await fixture.ngZone.run(async () => {
+                component.ngOnChanges(getSimpleChanges('node-1'));
+                await fixture.whenStable();
+            });
+            fixture.detectChanges();
 
-            expect(component.blobFileContent).toBe(mockBlob);
-            expect(component.urlFileContent).toBeNull();
-            expect(component.mimeType).toBe('application/pdf');
+            const viewer = testingUtils.getByDirective(ViewerComponent).componentInstance as ViewerComponent<unknown>;
+            expect(viewer.blobFile).toBe(mockBlob);
+            expect(viewer.urlFile).toBeFalsy();
+            expect(viewer.mimeType).toBe('application/pdf');
         });
 
         it('should fall back to URL-based viewing when PDF blob fetch fails', async () => {
@@ -1108,13 +1119,17 @@ describe('AlfrescoViewerComponent', () => {
 
             component.nodeId = 'node-1';
             component.showViewer = true;
-            component.ngOnChanges(getSimpleChanges('node-1'));
 
-            await fixture.whenStable();
+            await fixture.ngZone.run(async () => {
+                component.ngOnChanges(getSimpleChanges('node-1'));
+                await fixture.whenStable();
+            });
+            fixture.detectChanges();
 
-            expect(component.blobFileContent).toBeNull();
-            expect(component.urlFileContent).not.toBeNull();
-            expect(component.mimeType).toBe('application/pdf');
+            const viewer = testingUtils.getByDirective(ViewerComponent).componentInstance as ViewerComponent<unknown>;
+            expect(viewer.blobFile).toBeFalsy();
+            expect(viewer.urlFile).toBeTruthy();
+            expect(viewer.mimeType).toBe('application/pdf');
         });
 
         it('should fall back to URL-based viewing when PDF fetch returns non-ok response', async () => {
@@ -1137,13 +1152,17 @@ describe('AlfrescoViewerComponent', () => {
 
             component.nodeId = 'node-1';
             component.showViewer = true;
-            component.ngOnChanges(getSimpleChanges('node-1'));
 
-            await fixture.whenStable();
+            await fixture.ngZone.run(async () => {
+                component.ngOnChanges(getSimpleChanges('node-1'));
+                await fixture.whenStable();
+            });
+            fixture.detectChanges();
 
-            expect(component.blobFileContent).toBeNull();
-            expect(component.urlFileContent).not.toBeNull();
-            expect(component.mimeType).toBe('application/pdf');
+            const viewer = testingUtils.getByDirective(ViewerComponent).componentInstance as ViewerComponent<unknown>;
+            expect(viewer.blobFile).toBeFalsy();
+            expect(viewer.urlFile).toBeTruthy();
+            expect(viewer.mimeType).toBe('application/pdf');
         });
 
         it('should not expose intermediate null state during node setup', async () => {
@@ -1161,13 +1180,17 @@ describe('AlfrescoViewerComponent', () => {
 
             component.nodeId = 'node-1';
             component.showViewer = true;
-            component.ngOnChanges(getSimpleChanges('node-1'));
 
-            await fixture.whenStable();
+            await fixture.ngZone.run(async () => {
+                component.ngOnChanges(getSimpleChanges('node-1'));
+                await fixture.whenStable();
+            });
+            fixture.detectChanges();
 
-            expect(component.blobFileContent).toBe(mockBlob);
-            expect(component.urlFileContent).toBeNull();
-            expect(component.fileName).toBe('test.pdf');
+            const viewer = testingUtils.getByDirective(ViewerComponent).componentInstance as ViewerComponent<unknown>;
+            expect(viewer.blobFile).toBe(mockBlob);
+            expect(viewer.urlFile).toBeFalsy();
+            expect(viewer.fileName).toBe('test.pdf');
         });
     });
 });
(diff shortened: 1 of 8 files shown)


lib/core/src/lib/viewer/components/viewer-render/viewer-render.component.ts
```diff
@@ -241,15 +241,34 @@ export class ViewerRenderComponent implements OnChanges, OnInit {
...
         if (this.viewerType === 'unknown') {
             this.isLoading = false;
         }
 
-        this.extensionChange.emit(this.blobFile.type);
+        this.extensionChange.emit(blobMimeType || this.mimeType);
         this.scrollTop();
     }
 
+    /**
```
AleksanderSklorz: we don't add documentation to private functions

lib/core/viewer/pdf/src/lib/components/pdf-viewer/pdf-viewer.component.ts
```diff
@@ -292,7 +292,10 @@ export class PdfViewerComponent implements OnChanges, OnDestroy {
                 .then(() => {
                     setTimeout(() => this.scalePage('init'));
                 })
-                .catch(() => this.error.emit());
+                .catch((err) => {
+                    console.error('[ADF-PDF-VIEWER] executePdf failed:', err);
```
AleksanderSklorz: we avoid console.error in production - this is not helpful for user and if we want to display some error for user then we display message on page or some notification snackbar

lib/content-services/src/lib/viewer/components/alfresco-viewer.component.ts
```diff
@@ -368,16 +368,19 @@ export class AlfrescoViewerComponent implements OnChanges, OnInit {
...
 
-                this.blobFileContent = await response.blob();
+                blobContent = await response.blob();
                 urlFileContent = null;
             } catch (error) {
                 console.error('[ADF DEBUG] Failed to fetch PDF as blob, falling back to URL', error);
             }
         }
 
+        // Assign all bound properties atomically to prevent intermediate states
```
AleksanderSklorz: I think no need for that comment because for Angular developers it will be understandable what below code does

lib/content-services/src/lib/viewer/components/alfresco-viewer.component.ts
AleksanderSklorz, line 1: Missing unit tests for that file - for each change we always write unit tests to ensure everything works as expected and to protect against future issues

lib/core/src/lib/viewer/components/viewer-render/viewer-render.component.ts
AleksanderSklorz, line 1: same like https://github.com/Alfresco/alfresco-ng2-components/pull/12023/changes#r3511641806

lib/core/src/lib/viewer/components/viewer.component.ts
AleksanderSklorz, line 1: same like https://github.com/Alfresco/alfresco-ng2-components/pull/12023/changes#r3511641806

lib/content-services/src/lib/viewer/components/alfresco-viewer.component.spec.ts
```diff
@@ -1063,4 +1063,111 @@ describe('AlfrescoViewerComponent', () => {
...
+                )
+            );
+
+            component.nodeId = 'node-1';
+            component.showViewer = true;
+            component.ngOnChanges(getSimpleChanges('node-1'));
+
+            await fixture.whenStable();
+
+            expect(component.blobFileContent).toBe(mockBlob);
```
AleksanderSklorz: If possible we test final result so for example here you value of blobFileContent but it's not final step yet. Final step will be in html in this line

[blobFile]="blobFileContent"

So the flow is - component calls ngOnChanges, then after logic executes then taht blobFileContent changes and then angular detects change and pass new value to adf-viewer component in html by [blobFile]="blobFileContent" and that is final result in that component. 

To test that you would need to use 

UnitTestingUtils class which has getByDirective function. You can pass to that function name for class for adf-viewer component and it will read reference to that component from html. After that you can read any properties of that component so you can test if new value has been passed to it. 

Please do the same for rest places here and in other files with your tests. 

NOTE: I see this file is a bit old (as viewer is old functionality) so to see examples of UnitTestingUtils please check other places where it's used

lib/core/viewer/pdf/src/lib/components/pdf-viewer/pdf-viewer.component.spec.ts
```diff
@@ -160,6 +160,44 @@ describe('Test PdfViewer component', () => {
...
+            spyOn(component.error, 'emit');
+            spyOn(console, 'error');
+
+            const failingLoadingTask = {
+                promise: Promise.reject(new Error('PDF load failed')),
+                onPassword: null,
+                onProgress: null
+            };
+            spyOn(component as any, 'setupPdfJsWorker').and.returnValue(Promise.resolve());
+            (component as any).pdfjsLib = { getDocument: () => failingLoadingTask, GlobalWorkerOptions: {} };
```
AleksanderSklorz: we should avoid using as any only to get some private fields. In most cases there is other way to get them. For example here: 
If you check component how pdfjsLib is set then you can see that line 

private readonly pdfjsLib = inject(PDFJS_MODULE);

You can see 'inject'. If something is injected in production code then it often means that it can be injected also in tests. So instead of getting this like that you can try something like this to get that 

TestBed.inject(PDFJS_MODULE)

lib/core/viewer/pdf/src/lib/components/pdf-viewer/pdf-viewer.component.spec.ts
```diff
@@ -160,6 +160,44 @@ describe('Test PdfViewer component', () => {
...
+        it('should emit error event when PDF loading fails', fakeAsync(() => {
+            spyOn(component.error, 'emit');
+            spyOn(console, 'error');
+
+            const failingLoadingTask = {
+                promise: Promise.reject(new Error('PDF load failed')),
+                onPassword: null,
+                onProgress: null
+            };
+            spyOn(component as any, 'setupPdfJsWorker').and.returnValue(Promise.resolve());
```
AleksanderSklorz: setupPdfJsWorker is private function of component which you test so should not be mocked. We should test it's logic instead of mocking it

lib/core/viewer/pdf/src/lib/components/pdf-viewer/pdf-viewer.component.spec.ts
```diff
@@ -160,6 +160,44 @@ describe('Test PdfViewer component', () => {
         });
     });
 
+    describe('executePdf error handling', () => {
+        it('should emit error event when PDF loading fails', fakeAsync(() => {
+            spyOn(component.error, 'emit');
+            spyOn(console, 'error');
+
+            const failingLoadingTask = {
```
AleksanderSklorz: Actually this object has no type - is it possible to set PDFDocumentLoadingTask type or use 'as PDFDocumentLoadingTask'?

lib/core/viewer/pdf/src/lib/components/pdf-viewer/pdf-viewer.component.spec.ts
```diff
@@ -160,6 +160,44 @@ describe('Test PdfViewer component', () => {
...
+            flush();
+
+            expect(component.error.emit).toHaveBeenCalled();
+        }));
+
+        it('should log error to console when PDF loading fails', fakeAsync(() => {
+            spyOn(component.error, 'emit');
+            spyOn(console, 'error');
+
+            const failingLoadingTask = {
```
AleksanderSklorz: same like https://github.com/Alfresco/alfresco-ng2-components/pull/12023/changes#r3513364190

lib/core/viewer/pdf/src/lib/components/pdf-viewer/pdf-viewer.component.spec.ts
```diff
@@ -160,6 +160,44 @@ describe('Test PdfViewer component', () => {
...
+                promise: Promise.reject(new Error('PDF load failed')),
+                onPassword: null,
+                onProgress: null
+            };
+            spyOn(component as any, 'setupPdfJsWorker').and.returnValue(Promise.resolve());
```
AleksanderSklorz: for this line and line below check comments in previous test

lib/core/viewer/pdf/src/lib/components/pdf-viewer/pdf-viewer.component.spec.ts
```diff
@@ -160,6 +160,44 @@ describe('Test PdfViewer component', () => {
...
+            (component as any).pdfjsLib = { getDocument: () => failingLoadingTask, GlobalWorkerOptions: {} };
+
+            component.executePdf({ data: new ArrayBuffer(0) });
+            flush();
+
+            expect(console.error).toHaveBeenCalledWith('[ADF-PDF-VIEWER] executePdf failed:', jasmine.any(Error));
```
AleksanderSklorz: Well we don't test console code because of two reasons 
- consoles are used only by developers
- and should not be in production code 

I added one comment in pdf-viewer.component.ts saying that we should not add that console.error to code ideally

lib/core/viewer/pdf/src/lib/components/pdf-viewer/pdf-viewer.component.spec.ts
```diff
@@ -354,6 +355,54 @@ describe('Test PdfViewer component', () => {
...
+    let fixture: ComponentFixture<PdfViewerComponent>;
+    let component: PdfViewerComponent;
+
+    beforeEach(() => {
+        const failingLoadingTask = {
+            promise: Promise.reject(new Error('PDF load failed')),
+            onPassword: null,
+            onProgress: null,
+            destroy: () => Promise.resolve()
+        } as unknown as PDFDocumentLoadingTask;
```
AleksanderSklorz: as unknown seems redundant, it's enough to have just as PDFDocumentLoadingTask.

lib/core/src/lib/viewer/components/viewer-render/viewer-render.component.spec.ts
```diff
@@ -585,4 +585,68 @@ describe('ViewerComponent', () => {
...
+
+            expect(component.viewerType).toBe('pdf');
+        });
+
+        it('should fall back to mimeType input when blob type resolves to unknown', () => {
+            component.blobFile = new Blob(['content'], { type: '' });
+            component.mimeType = 'application/pdf';
+            component.ngOnChanges();
+
+            expect(component.viewerType).toBe('pdf');
```
AleksanderSklorz: Here is similar situation like I added comment in alfresco-viewer.component.spec.ts last time: 
You check component.viewerType field but it's not final result. If you analyze what happens with that field you can see that depending on that field different component is rendered in HTML so your test should check if correct element exists in HTML as we test final results. If we won't test final result then viewerType field can change but application can be still broken because somebody can change condition in html and incorrect element will be rendered but if we will test final result then we are 100% sure it protects code from future bugs. Please apply that for all component.viewerType tests 


Image

lib/core/src/lib/viewer/components/viewer.component.spec.ts
```diff
@@ -117,6 +117,47 @@ describe('ViewerComponent', () => {
...
         });
 
+        it('should strip MIME type parameters from blob type', () => {
+            const mockSimpleChanges: SimpleChanges = {
+                blobFile: new SimpleChange(null, { type: 'application/pdf;charset=utf-8' }, true)
+            };
+
+            component.ngOnChanges(mockSimpleChanges);
+
+            expect(component.mimeType).toBe('application/pdf');
```
AleksanderSklorz: Similar case like in previous comment - please test final result (html), the same for rest component.mimeType tests

AleksanderSklorz:
@adamswiderskipw btw at the beginning of that PR's comments there are 2 comments which seem not addressed (they are not closed because they are waiting for fix)

lib/core/src/lib/viewer/components/viewer-render/viewer-render.component.spec.ts
```diff
@@ -582,7 +582,77 @@ describe('ViewerComponent', () => {
...
             expect(getMainLoader()).toBeNull();
-            expect(component.viewerType).toBe('unknown');
+            expect(testingUtils.getByCSS('adf-viewer-unknown-format')).not.toBeNull();
+        });
+    });
+
+    describe('Blob MIME type handling', () => {
+        it('should resolve viewerType from blob type when blob has valid MIME type', () => {
+            component.blobFile = new Blob(['content'], { type: 'application/pdf' });
+            component.ngOnChanges();
```
AleksanderSklorz, line 592: Can you review all your tests and check if you need to call ngOnChanges when you call detectChanges? In most cases when you call detectChanges, ngOnChanges is not required. Please remove ngOnChanges when you validate

lib/content-services/src/lib/viewer/components/alfresco-viewer.component.spec.ts
```diff
@@ -1063,4 +1070,127 @@ describe('AlfrescoViewerComponent', () => {
...
+            );
+
+            component.nodeId = 'node-1';
+            component.showViewer = true;
+
+            await fixture.ngZone.run(async () => {
+                component.ngOnChanges(getSimpleChanges('node-1'));
+                await fixture.whenStable();
+            });
+            fixture.detectChanges();
```
AleksanderSklorz, line 1095: Can you review all your tests and check if you need to call ngOnChanges (and/or fixture.whenStable) when you call detectChanges? In most cases when you call detectChanges, ngOnChanges is not required. Please remove ngOnChanges when you validate

lib/core/src/lib/viewer/components/viewer.component.spec.ts
```diff
@@ -113,8 +114,57 @@ describe('ViewerComponent', () => {
             };
 
             component.ngOnChanges(mockSimpleChanges);
+            fixture.detectChanges();
```
AleksanderSklorz, line 117: same like https://github.com/Alfresco/alfresco-ng2-components/pull/12023/changes#r3528814772

AleksanderSklorz (approved):

PR #12046, opened 2026-07-09
AAE-47579 feat(core): Introduce EDIT_JSON_EDITOR injection token and add copy button to EditJsonDialogComponent

munir-fati-haji:
## Summary

This PR enhances the JSON editor dialog by introducing an injection token EDIT_JSON_EDITOR for customizable editor implementations and adding a copy-to-clipboard button for improved usability. These changes enable advanced editors (like Monaco Editor) to be used instead of the default textarea-based editor while maintaining backward compatibility.

### 1. Add EDIT_JSON_EDITOR Injection Token
- **New Token**: `EDIT_JSON_EDITOR` - allows applications to override the default JSON editor component
- **Default Behavior**: Defaults to the existing textarea-based JSON editor for backward compatibility
- **Flexibility**: Applications can provide custom editor implementations via dependency injection

### 2. Add Copy-to-Clipboard Button
- **New Feature**: Copy button in EditJsonDialogComponent toolbar
- **UX Improvement**: Users can quickly copy the entire JSON content with a single click

### 3. Update EditJsonDialogComponent
- **Enhancement**: Modified to use the `EDIT_JSON_EDITOR` injection token to render the editor
- **Backward Compatible**: Existing applications continue to work without any changes
- **Extensible**: Enables applications to swap in custom editors (Monaco, CodeMirror, etc.)

> - [ ] Bugfix
> - [x] Feature
> - [ ] Build related changes
> - [ ] Documentation

> - [ ] Yes
> - [x] No

**Other information**:


DenysVuika:
You should be adding only English locale, not other translations, those will be auto-translated

lib/core/src/lib/i18n/en.json
```diff
@@ -171,7 +171,9 @@
       },
       "EDIT_JSON": {
         "CLOSE": "Close",
-        "UPDATE": "Update"
+        "UPDATE": "Update",
+        "COPY": "Copy",
+        "COPIED": "Copied to clipboard"
```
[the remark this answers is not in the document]
munir-fati-haji: Other translations, will be auto-translated

lib/core/src/lib/dialogs/edit-json/edit-json.dialog.html
```diff
@@ -1,13 +1,17 @@
 <h1 mat-dialog-title>{{ title | translate }}</h1>
 <mat-dialog-content class="adf-edit-json-dialog-content">
-    <textarea [(ngModel)]="value" [attr.readonly]="!editable ? true : null"></textarea>
+    <ng-content>
+        <textarea [(ngModel)]="value" [attr.readonly]="!editable ? true : null"></textarea>
+    </ng-content>
 </mat-dialog-content>
```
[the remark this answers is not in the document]
munir-fati-haji: I believe this is the correct pattern for optional content projection in Angular. The fallback content inside `ng-content` renders when nothing is projected. This allows both use cases:

1. Default textarea renders when no custom editor is provided
2. Custom editor (e.g., Monaco) projects and replaces the textarea when provided

This is intentional design to support customization without breaking existing behavior.

lib/core/src/lib/dialogs/edit-json/edit-json.dialog.ts
```diff
@@ -29,26 +32,40 @@ export interface EditJsonDialogSettings {
...
 
-    editable: boolean = false;
-    title: string = 'JSON';
+    protected readonly customEditor = inject(EDIT_JSON_EDITOR);
+    protected title = this.settings?.title ?? 'JSON';
 
-    @Input()
-    value: string = '';
+    editable = this.settings?.editable ?? false;
+    readonly value = model(this.settings?.value ?? '');
```
alep85, line 51: this change breaks the public API, since value becomes a model() and ngOnInit is removed. To keep it compatible we could:
- leave value as a plain @Input() string and keep ngOnInit
- use a new private content signal for the custom editor, synced back to value
- keep the textarea as the default
- show the copy button only with @if (customEditor)
- leave the token default as null, so nothing changes when no editor is provided

lib/core/src/lib/dialogs/edit-json/edit-json.dialog.spec.ts
```diff
@@ -0,0 +1,231 @@
...
+    it('should copy an empty value when the dialog has no content', () => {
+        setup({ value: '', editable: false });
+        spyOn(clipboardService, 'copyContentToClipboard');
+
+        testingUtils.clickByDataAutomationId('adf-edit-json-dialog-copy');
+
+        expect(clipboardService.copyContentToClipboard).toHaveBeenCalledWith('', jasmine.any(String));
+        expect(clipboardService.copyContentToClipboard).toHaveBeenCalledTimes(1);
+    });
+});
```
alep85, line 237: Can we test that shows the textarea if customEditor is not set?

alep85 (changes requested):

lib/core/src/lib/dialogs/edit-json/edit-json.dialog.ts
munir-fati-haji: As discussed during our call, this does not break the public API.

`model()` is Angular's signal-based equivalent that combines `@Input()` and `@Output()` — it is fully backwards compatible as a drop-in replacement. Consumers binding to `[value]` continue to work exactly as before.

As for the other points (copy button gated by `@if (customEditor)`, `token` defaulting to `null`, `textarea` as default) — these were all already in place as we discussed.  :)

lib/core/src/lib/dialogs/edit-json/edit-json.dialog.spec.ts
munir-fati-haji: I have added an explicit test that covers the textarea fallback when no custom editor is provided as discussed :)

alep85 (approved):

PR #12060, opened 2026-07-13
[ACS-9584] Refresh after the change of name of a node

Shivangi917:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

> - [ ] Yes
> - [x] No

**Other information**:
https://hyland.atlassian.net/browse/ACS-9584
https://hyland.atlassian.net/browse/ACS-4747


lib/content-services/src/lib/viewer/components/alfresco-viewer.component.ts
```diff
@@ -265,7 +265,11 @@ export class AlfrescoViewerComponent implements OnChanges, OnInit {
             .pipe(
                 filter(
                     (node) =>
-                        node && node.id === this.nodeId && this.getNodeVersionProperty(this.nodeEntry.entry) !== this.getNodeVersionProperty(node)
+                        !!node &&
+                        node.id === this.nodeId &&
+                        !!this.nodeEntry?.entry &&
+                        (this.getNodeVersionProperty(this.nodeEntry.entry) !== this.getNodeVersionProperty(node) ||
+                            this.nodeEntry.entry.name !== node.name)
```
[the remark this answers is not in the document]
Shivangi917: I investigated that but both (nodeEntry.entry.name and node.name) have different value.

lib/content-services/src/lib/viewer/components/alfresco-viewer.component.spec.ts
```diff
@@ -585,6 +585,32 @@ describe('AlfrescoViewerComponent', () => {
             expect(component.nodeMimeType).toEqual('application/msWord');
         });
 
+        it('should update fileName when the node name changes', async () => {
+            const defaultNode: Node = {
```
AleksanderSklorz: type should not be set on both side - you have now Node type to the left side of = and to the right side of = because you have "as Node". Please remove one of those types because type should be set only on one side

lib/core/src/lib/viewer/components/viewer-render/viewer-render.component.ts
```diff
@@ -219,8 +220,10 @@ export class ViewerRenderComponent implements OnChanges, OnInit {
         this.isLoading = true;
     }
 
-    ngOnChanges() {
-        this.isLoading = true;
+    ngOnChanges(changes: SimpleChanges) {
+        if (changes['urlFile'] || changes['blobFile']) {
```
AleksanderSklorz, line 224: missing unit tests for that change

AleksanderSklorz:
@Shivangi917 I see a lot copilot comments which are not closed. Please check all comments - if they are valid then please fix them, if they are invalid then please answer there why they are invalid. After that please click "Resolve conversation". Thanks that we know which comments are addressed

lib/core/src/lib/viewer/components/viewer-render/viewer-render.component.spec.ts
```diff
@@ -519,6 +519,25 @@ describe('ViewerComponent', () => {
     describe('Spinner', () => {
         const getMainLoader = (): DebugElement => testingUtils.getByCSS('.adf-viewer-render-main-loader');
 
+        it('should set isLoading to true when urlFile changes', () => {
+            component.isLoading = false;
+            component.urlFile = 'new-file.pdf';
+
+            component.ngOnChanges({ urlFile: new SimpleChange('old-file.pdf', 'new-file.pdf', false) });
+
+            expect(component.isLoading).toBeTrue();
```
AleksanderSklorz: This test is not fully correct. Ideally we should test final result of scenario, so for example here if you will follow what happens then yes you are right that isLoading is changed BUT it's not final result. Final result is in html that when isLoading changes then for example spinner is shown or hidden. See tests below that they check it's visibility. So please check visibility instead of value of field. We always try to follow whole flow from start to end and in expect part we have 'end' part

lib/core/src/lib/viewer/components/viewer-render/viewer-render.component.spec.ts
```diff
@@ -519,6 +519,25 @@ describe('ViewerComponent', () => {
...
+        });
+
+        it('should set isLoading to true when blobFile changes', () => {
+            component.isLoading = false;
+            const blob = new Blob(['content'], { type: 'application/pdf' });
+            component.blobFile = blob;
+
+            component.ngOnChanges({ blobFile: new SimpleChange(null, blob, false) });
+
+            expect(component.isLoading).toBeTrue();
```
AleksanderSklorz: This test is not fully correct. Ideally we should test final result of scenario, so for example here if you will follow what happens then yes you are right that isLoading is changed BUT it's not final result. Final result is in html that when isLoading changes then for example spinner is shown or hidden. See tests below that they check it's visibility. So please check visibility instead of value of field. We always try to follow whole flow from start to end and in expect part we have 'end' part

AleksanderSklorz (approved):

dominikiwanekhyland (approved):

MichalKinas (approved):

PR #12106, opened 2026-07-29
Migrate to webpack-dev-server 6.0.0

DenysVuika:
Migrate webpack-dev-server to version 6.0.0
See https://github.com/webpack/webpack-dev-server/releases/tag/v6.0.0

### DIFF
--- a/package.json
+++ b/package.json
@@ -142,13 +142,13 @@
     "ts-node": "10.9.2",
     "typescript": "5.9.3",
     "undici": "8.7.0",
-    "webpack": "5.109.0"
+    "webpack": "5.109.2"
   },
   "license": "Apache-2.0",
   "engines": {
     "node": ">=24.14.0"
   },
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -148,7 +148,7 @@
   "engines": {
     "node": ">=24.14.0"
   },
-  "packageManager": "pnpm@11.9.0+sha512.bd682d5d03fe525ef7c9fd6780c6884d1e756ac4c9c9fe00c538782824310dcf90e3ddc4f53835f06dfaebd5085e41855e0bcbb3b60de2ac5bbab89e5036f03b",
+  "packageManager": "pnpm@11.17.0",
```
dsibilio: Any reason why we dropped the sha pinning for the package manager?

dsibilio: cc: @VitoAlbano

DenysVuika: it was automated update via "pn self-update"... is there a way to do that with hash?

DenysVuika: should be fixed now with corepack

DenysVuika:
@dsibilio I have regenerated the sha for the pnpm

BSekula (approved):

dsibilio (approved):

PR #12143, opened 2026-08-07
AAE-49781 Filtering out WS tasks events not related to current user

ehsan-2019:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

AAE-49781

> - [ ] Yes
> - [ ] No

**Other information**:

### DIFF
--- a/lib/process-services-cloud/src/lib/process/process-filters/components/process-filters/process-filters-cloud.component.spec.ts
+++ b/lib/process-services-cloud/src/lib/process/process-filters/components/process-filters/process-filters-cloud.component.spec.ts
@@ -341,6 +341,15 @@ describe('ProcessFiltersCloudComponent', () => {
             expect(lastValue).toBeDefined();
         });
 
+        it('should render an empty filter list instead of the loading spinner when the request fails', async () => {
+            getProcessFiltersSpy.and.returnValue(throwError(() => new Error('wrong request')));
+
+            await bindAppName();
+
+            expect(fixture.debugElement.queryAll(By.css('.adf-process-filters__entry')).length).toBe(0);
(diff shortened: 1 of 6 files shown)


eromano (changes requested):
This PR should be tested in scenario where task are assigned to groups, I have the feeling that in those cases the counter will not work anymore or anyway be refreshed only one the user those some interactions as when the option notification is off

ehsan-2019:
> This PR should be tested in scenario where task are assigned to groups, I have the feeling that in those cases the counter will not work anymore or anyway be refreshed only one the user those some interactions as when the option notification is off

I checked this scenario, and it looks like it's working well. It also reduces the number of calls significantly. Here's how it works:

- UserA and UserB are in the same group.
- UserA starts 10 tasks.
- At this point, both users see the queued tasks counter set to 10.
- UserA then claims one of the queued tasks.

This triggers two events:

- TASK_CREATED with no assignee
- TASK_ASSIGNED with UserA as the assignee

Since we allow events without an assignee to go through, the queued tasks counter gets updated for **both users**. However, by filtering out events that have an explicit assignee different from the current user, we avoid a large number of unnecessary calls.
If I understand it correctly, whenever any user performs an action, all members of the same group are notified. As a result, many users end up making calls that don't actually lead to any counter updates.
If you keep an eye on the Network tab in RC for a few minutes, you'll notice that for most requests to the count endpoint, none of the counters are refreshed. That suggests those requests can be safely filtered out.

ehsan-2019:
Since we’ve implemented a new batch endpoint for the counters and made several improvements to minimize the number of calls, I’m closing this PR.

PR #12186, opened 2026-08-20
AAE-49653 Migrating to batch count endpoint

ehsan-2019:

> - [ ] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

AAE-49653

**Background**

The frontend currently invokes separate APIs for task and process instance counters, issuing one HTTP request per filter.

A new unified endpoint will provide counts for multiple filters in a single request, regardless of the resource type.

**Objective**
Refactor the counter update flow to use the new generic batch count API.

Scope
Replace calls to:

`/query/v1/tasks/count`

`/query/v1/process-instances/count`

Introduce a shared client/service for invoking the new `/query/v1/count` endpoint.

Aggregate all filters requiring a refresh into a single request.

Update local counter state from the batch response.

Remove duplicated logic between task and process instance counter retrieval where appropriate.

> - [ ] Yes
> - [ ] No

**Other information**:


lib/process-services-cloud/src/lib/process/process-filters/components/process-filters/process-filters-cloud.component.ts
```diff
@@ -43,7 +45,12 @@ export class ProcessFiltersCloudComponent implements OnInit, OnChanges {
     @Input()
     appName: string = '';
 
-    /** (optional) From Activiti 8.7.0 forward, use the 'POST' method to get the process count */
+    /**
+     * (optional) From Activiti 8.7.0 forward, use the 'POST' method to get the process count.
+     *
+     * @deprecated only used by the backends without `POST /query/v1/count`. It will be removed,
```
VitoAlbano: Please remove this comment

lib/process-services-cloud/src/lib/process/process-filters/components/process-filters/process-filters-cloud.component.ts
```diff
@@ -79,27 +86,30 @@ export class ProcessFiltersCloudComponent implements OnInit, OnChanges {
...
     counters: { [key: string]: number } = {};
-    enableNotifications = true;
-    notificationDebounceTime = 3000;
     currentFiltersValues: { [key: string]: number } = {};
     updatedFiltersSet = new Set<string>();
+    enableNotifications = true;
+    notificationDebounceTime = 3000;
     private filtersLoadedFor?: string;
+    private countersSubscription?: Subscription;
+    private batchedCounters = true;
```
VitoAlbano, line 102: ?

lib/process-services-cloud/src/lib/process/process-filters/components/process-filters/process-filters-cloud.component.ts
```diff
@@ -167,20 +183,6 @@ export class ProcessFiltersCloudComponent implements OnInit, OnChanges {
...
     }
 
-    /**
-     * Check equality of the filter names by translating the given name strings
-     *
-     * @param name1 source name
-     * @param name2 target name
-     * @returns `true` if filter names are equal, otherwise `false`
-     */
-    private checkFilterNamesEquality(name1: string, name2: string): boolean {
```
VitoAlbano, line 177: Was this unused?

lib/process-services-cloud/src/lib/process/process-filters/components/process-filters/process-filters-cloud.component.ts
```diff
@@ -247,84 +249,140 @@ export class ProcessFiltersCloudComponent implements OnInit, OnChanges {
...
-
     isActiveFilter(filter: ProcessFilterCloudModel): boolean {
         return this.currentFilter.name === filter.name;
     }
 
+    /**
+     * @deprecated `getFilterCounters` keeps the counters in sync with the engine events, so nothing is
+     * subscribed here anymore. It will be removed in ADF 10.0.0.
+     */
     initProcessNotification(): void {
```
VitoAlbano: i don't understand why we are deprecating instead of removing, aren't we updating to the batch approach? then let's move on

lib/process-services-cloud/src/lib/process/process-filters/components/process-filters/process-filters-cloud.component.ts
```diff
@@ -247,84 +249,140 @@ export class ProcessFiltersCloudComponent implements OnInit, OnChanges {
...
     updateFilterCounter(filter: ProcessFilterCloudModel): void {
-        if (!filter?.showCounter) {
+        const filterKey = filter?.showCounter ? filter.key : undefined;
+        if (!filterKey) {
             return;
         }
 
-        this.fetchProcessFilterCounter(filter)
+        /* `defer` keeps a query that cannot be built from breaking the counters of the other filters. */
+        defer(() => this.fetchProcessFilterCounter(filter))
```
VitoAlbano, line 291: Why do we need defer here? are the query breaking each other? why?

lib/process-services-cloud/src/lib/task/task-filters/components/task-filters/task-filters-cloud.component.ts
```diff
@@ -69,9 +76,12 @@ export class TaskFiltersCloudComponent extends BaseTaskFiltersCloudComponent imp
     notificationDebounceTime = 3000;
     currentFiltersValues: { [key: string]: number } = {};
     private filtersLoadedFor?: string;
+    private countersSubscription?: Subscription;
+    private batchedCounters = true;
```
VitoAlbano, line 87: why do we need this boolean here?

lib/process-services-cloud/src/lib/task/task-filters/components/task-filters/task-filters-cloud.component.ts
```diff
@@ -42,7 +44,12 @@ import { AsyncPipe } from '@angular/common';
...
     protected readonly TASKS_ROUTE = '/task-list-cloud';
 
-    /** (optional) From Activiti 8.7.0 forward, use the 'POST' method to get the task count. */
+    /**
+     * (optional) From Activiti 8.7.0 forward, use the 'POST' method to get the task count.
+     *
+     * @deprecated only used by the backends without `POST /query/v1/count`. It will be removed,
+     * along with the 'GET' method, in ADF 10.0.0.
+     */
     @Input()
```
VitoAlbano, line 51: We can simply remove this if we do not want to support other actions

lib/process-services-cloud/src/lib/task/task-filters/components/task-filters/task-filters-cloud.component.ts
```diff
@@ -131,59 +144,54 @@ export class TaskFiltersCloudComponent extends BaseTaskFiltersCloudComponent imp
...
+     * endpoint. It will be removed in ADF 10.0.0.
      */
     updateFilterCounter(filter: TaskFilterCloudModel): void {
         if (!filter?.showCounter) {
             return;
         }
-        this.fetchTaskFilterCounter(filter)
+
+        /* `defer` keeps a query that cannot be built from breaking the counters of the other filters. */
+        defer(() => this.fetchTaskFilterCounter(filter))
```
VitoAlbano, line 174: defer also herewhich i'm not sure why w need it

lib/process-services-cloud/src/lib/task/task-filters/components/task-filters/task-filters-cloud.component.ts
```diff
@@ -131,59 +144,54 @@ export class TaskFiltersCloudComponent extends BaseTaskFiltersCloudComponent imp
...
-                .subscribe((result) => {
-                    result.forEach((taskEvent) => {
-                        this.checkFilterCounter(taskEvent.entity);
-                    });
-
-                    this.updateFilterCounters();
-                    this.filterCounterUpdated.emit(result);
-                });
-        } else {
+        if (!this.enableNotifications) {
```
VitoAlbano: This change doesn't look right, as you negated the condition just to trigger an extra return whilst you can simply use else here.

VitoAlbano (changes requested):

lib/process-services-cloud/src/lib/process/process-filters/components/process-filters/process-filters-cloud.component.ts
ehsan-2019: my goal was to deprecate this input in the next ADF version.

lib/process-services-cloud/src/lib/process/process-filters/components/process-filters/process-filters-cloud.component.ts
ehsan-2019: this is for the backward compatibility mostly.

lib/process-services-cloud/src/lib/process/process-filters/components/process-filters/process-filters-cloud.component.ts
ehsan-2019: I have just moved it to the bottom of the class.

lib/process-services-cloud/src/lib/process/process-filters/components/process-filters/process-filters-cloud.component.ts
ehsan-2019: I tried to keep supporting the old mechanism so an ADF consumer with old BE can keep using the feature. because the new endpoint will be available only on latest Activiti.

lib/process-services-cloud/src/lib/process/process-filters/components/process-filters/process-filters-cloud.component.ts
ehsan-2019: because `fetchProcessFilterCounter` can potentially throw in `this.buildQueryData` method before returning an observable, and the `catchError` can not catch it. With defer, the call is postponed until there is an observable and someone subscribes to it. from here `catchError` will able to catch the error.

lib/process-services-cloud/src/lib/task/task-filters/components/task-filters/task-filters-cloud.component.ts
ehsan-2019: with this flag we decide to use batch endpoint or normal count. in our FE/BE contract  we need to use a requestId in the payload for each filter. I am using filter.key but it's optional. so when the key is not present instead of not doing anything I use the old endpoint to update the counter. when we remove the old mechanism and update the types we can simplify this.

lib/process-services-cloud/src/lib/task/task-filters/components/task-filters/task-filters-cloud.component.ts
ehsan-2019: for the same reason above. I will try to use a more clear comment.

lib/process-services-cloud/src/lib/task/task-filters/components/task-filters/task-filters-cloud.component.ts
ehsan-2019: ok, I can merge two IFs.

lib/process-services-cloud/src/lib/task/task-filters/components/task-filters/task-filters-cloud.component.ts
ehsan-2019: I was thinking removing is a BC that could be done in the next major version. for now I marked it as deprecated.

lib/process-services-cloud/src/lib/models/filter-counters-cloud.model.ts
```diff
@@ -0,0 +1,68 @@
...
+
+export type FilterCounterEntityType = (typeof FilterCounterEntityType)[keyof typeof FilterCounterEntityType];
+
+export interface FilterCountersQuerySort {
+    field: string;
+    direction: string;
+    isProcessVariable: boolean;
+}
+
+/** A single query of the batched count request, holding the criteria of one filter. */
```
VitoAlbano: No comments in the code.

VitoAlbano (approved):

PR #12207, opened 2026-09-01
AAE-47554 Add actionlint to CI

andrphilippi:
## Summary

Adds [actionlint](https://github.com/rhysd/actionlint) as a CI gate for GitHub Actions workflows and composite actions, and fixes every issue it surfaced.

Relates to [AAE-47554](https://hyland.atlassian.net/browse/AAE-47554).

This repo uses husky + lint-staged (no pre-commit framework), so actionlint is wired in as a dedicated CI workflow instead of a pre-commit hook. shellcheck is preinstalled on the `ubuntu-latest` runners and is auto-detected by actionlint, satisfying the "shellcheck runs successfully on CI" acceptance criterion.

### Added
- `.github/workflows/actionlint.yml` — runs actionlint (pinned `v1.7.12`) on PRs that touch `.github/workflows/**` or `.github/actions/**`. No third-party actions are used (the official download script is pinned to the version tag), so the SHA-pinning policy is respected.
- `.github/actionlint.yaml` — ignores auto-generated `*.lock.yml` compiled workflows (mirrors the usual pre-commit `exclude`).

### Fixed (issues raised by actionlint)
- Removed the invalid `type` key from composite action inputs in `setup`, `before-install`, and `enable-dryrun` — `type` is only valid for `workflow_call` inputs, and its presence made the action metadata unparseable.
- Quoted `$GITHUB_OUTPUT`, `$BASE_REF`, `$GITHUB_REPOSITORY`, and `$PR_NUMBER` in run scripts across `pull-request.yml`, `unit-test-workflow.yml`, and `build-lib-workflow.yml` (SC2086).
- Removed the redundant `.github/*.md` `case` pattern already covered by `*.md` in `pull-request.yml` (SC2221/SC2222).

## Test plan
- [x] `actionlint v1.7.12` runs clean locally with shellcheck 0.11.0 on PATH (reproducing the Linux CI environment)
- [ ] `actionlint` CI job passes on this PR

[AAE-47554]: https://hyland.atlassian.net/browse/AAE-47554?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -1,0 +1,34 @@
+name: "pre-commit"
+
+on:
+  pull_request:
+    branches:
+      - develop
+      - master
+      - develop-patch*
+      - master-patch*
+    paths:
+      - '.github/workflows/**'
+      - '.github/actions/**'
+      - '.pre-commit-config.yaml'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+jobs:
+  pre-commit:
+    name: "Run pre-commit hooks"
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+
+      - name: Run pre-commit
+        uses: Alfresco/alfresco-build-tools/.github/actions/pre-commit@da99ab845e78301fcb0680d16cbc185a40a1938c # v18.24.1
+        with:
+          skip_checkout: true
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -68,16 +68,16 @@ jobs:
         run: |
           if [ "${{ github.event_name }}" != "pull_request" ]; then
             echo "Not a PR event — assuming code changed"
-            echo "code-changed=true" >> $GITHUB_OUTPUT
+            echo "code-changed=true" >> "$GITHUB_OUTPUT"
             exit 0
           fi
 
-          FILES=$(gh api /repos/$GITHUB_REPOSITORY/pulls/${{ github.event.pull_request.number }}/files --paginate --jq '.[].filename')
+          FILES=$(gh api "/repos/$GITHUB_REPOSITORY/pulls/${{ github.event.pull_request.number }}/files" --paginate --jq '.[].filename')
 
           CODE_CHANGED="false"
           while IFS= read -r file; do
             case "$file" in
-              *.md|docs/*|.github/*.md|.github/CODEOWNERS|.github/dependabot.yml|LICENSE*|NOTICE*|.editorconfig|.gitattributes)
+              *.md|docs/*|.github/CODEOWNERS|.github/dependabot.yml|LICENSE*|NOTICE*|.editorconfig|.gitattributes)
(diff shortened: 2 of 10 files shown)


.pre-commit-config.yaml
```diff
@@ -0,0 +1,6 @@
+repos:
+  - repo: https://github.com/rhysd/actionlint
+    rev: 914e7df21a07ef503a81201c76d2b11c789d3fca  # frozen: v1.7.12
+    hooks:
+      - id: actionlint
+        exclude: '\.lock\.yml$'
```
andrphilippi:
```suggestion
        exclude: ^\.github/workflows/.*\.lock\.yml$
```

.pre-commit-config.yaml
```diff
@@ -0,0 +1,6 @@
+repos:
+  - repo: https://github.com/rhysd/actionlint
+    rev: 914e7df21a07ef503a81201c76d2b11c789d3fca  # frozen: v1.7.12
+    hooks:
+      - id: actionlint
+        exclude: ^\.github/workflows/.*\.lock\.yml$
```
andrphilippi:
```suggestion
        exclude: '^\.github/workflows/.*\.lock\.yml$'
```

.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,34 @@
+name: "pre-commit"
```
atchertchian, line 1: there is already a workflow run on same events doing such checks, would be better not to add yet another workflow for this one, see https://github.com/Alfresco/alfresco-ng2-components/blob/develop/.github/workflows/pull-request.yml#L38

andrphilippi: Hi @atchertchian, pull-request uses code-changed to determine if code needs a kind of check. It is a manual check? using just the path filter in pre-commit.yml felt more natural. For sure I can adjust "Detect code changes" or introduce "Detect action changes" but that would be manual instead using a simple filter. What do you think?

atchertchian: pre-commit can cover so many things that ;path restrictions might become invalid, and this check is supposed to be quick, so I don't think this level of refinement is needed, I prefer having less workflows

andrphilippi: Moved in [86488befb272382061b498d2189e3e0e7db932a1](https://github.com/Alfresco/alfresco-ng2-components/pull/12207/changes/86488befb272382061b498d2189e3e0e7db932a1)

.github/workflows/pull-request.yml
```diff
@@ -67,27 +68,57 @@ jobs:
...
+          echo "workflows-changed=$WORKFLOWS_CHANGED" >> "$GITHUB_OUTPUT"
           echo "Code changed: $CODE_CHANGED"
+          echo "Workflows changed: $WORKFLOWS_CHANGED"
+
+  pre-commit:
+    name: "Run pre-commit hooks"
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    needs: [pre-checks]
+    if: ${{ needs.pre-checks.outputs.workflows-changed == 'true' }}
```
atchertchian: again I would run pre-commit in existing job that already checks out and does basic checks, ideally next to call to hyland/github-actions-ensure-sha-pinned-actions, no need to diffs on what changes on such quick checks

atchertchian (approved):

PR #12210, opened 2026-09-03
[ACS-12582] [ADW] Document list container is not accessible via keyboard navigation for file selection

dominikiwanekhyland:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://hyland.atlassian.net/browse/ACS-12582

> - [ ] Yes
> - [x] No

**Other information**:

### DIFF
--- a/lib/core/src/lib/datatable/components/datatable/datatable.component.ts
+++ b/lib/core/src/lib/datatable/components/datatable/datatable.component.ts
@@ -640,7 +640,7 @@ export class DataTableComponent implements OnInit, AfterContentInit, OnChanges, 
         }
 
         if (row) {
-            const rowIndex = this.data.getRows().indexOf(row) + (this.isHeaderVisible() ? 1 : 0);
+            const rowIndex = this.data.getRows().indexOf(row) + this.rowsOffset;
             this.keyManager.setActiveItem(rowIndex);
 
             const dataRowEvent = new DataRowEvent(row, mouseEvent, this);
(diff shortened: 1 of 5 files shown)


lib/core/src/lib/datatable/components/datatable/datatable.component.ts
```diff
@@ -923,6 +923,23 @@ export class DataTableComponent implements OnInit, AfterContentInit, OnChanges,
         return this.selectionMode && this.selectionMode.toLowerCase() === 'multiple';
     }
 
+    isRowKeyboardNavigable(): boolean {
```
AleksanderSklorz: we are in newer angular now so we don't need to make functions public to just use them in template. If it's used only in template then it should be protected. Public should be used only for a case when we use something outside component

lib/core/src/lib/datatable/components/datatable/datatable.component.ts
```diff
@@ -923,6 +923,23 @@ export class DataTableComponent implements OnInit, AfterContentInit, OnChanges,
...
+        return this.enableDragRows || this.multiselect || this.isSingleSelectionMode() || this.isMultiSelectionMode();
+    }
+
+    isRowActive(rowIndex: number): boolean {
```
AleksanderSklorz: we are in newer angular now so we don't need to make functions public to just use them in template. If it's used only in template then it should be protected. Public should be used only for a case when we use something outside component

lib/core/src/lib/datatable/components/datatable/datatable.component.ts
```diff
@@ -923,6 +923,23 @@ export class DataTableComponent implements OnInit, AfterContentInit, OnChanges,
...
+        const activeItemIndex = this.keyManager?.activeItemIndex ?? -1;
+        return activeItemIndex < this.rowsOffset ? rowIndex === 0 : activeItemIndex === rowIndex + this.rowsOffset;
+    }
+
+    onRowFocus(rowIndex: number): void {
```
AleksanderSklorz: we are in newer angular now so we don't need to make functions public to just use them in template. If it's used only in template then it should be protected. Public should be used only for a case when we use something outside component

lib/core/src/lib/datatable/components/datatable/datatable.component.ts
```diff
@@ -923,6 +923,23 @@ export class DataTableComponent implements OnInit, AfterContentInit, OnChanges,
...
+        this.keyManager?.updateActiveItem(rowIndex + this.rowsOffset);
+    }
+
+    private get rowsOffset(): number {
```
AleksanderSklorz: Incorrect place for getter, please check ts.instructions.md file which shows correct order 

Image

AleksanderSklorz (approved):

PR #12232, opened 2026-09-14
AAE-51793 Add retry with backoff to npm view in npm-check-bundle

dsibilio:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

The release workflow keeps failing in the npm check bundle job with `Package @alfresco/adf insights@<version> not found!`. This has hit basically every develop run since September 4.

adf insights actually publishes fine. But nx publishes it near the end of the batch since it depends on adf core and adf content services, while the check script queries it third, right after js api and adf core. So it usually gets checked before the registry has caught up with the new version. Confirmed the exact version shows up on the registry a bit later, so this is a timing race, not a real publish failure.

Full writeup in [AAE-51793](https://hyland.atlassian.net/browse/AAE-51793).

`npm-check-bundles.sh` now retries the `npm view` lookup with a short backoff (6 tries, 10 seconds apart) before giving up, instead of failing on the first miss. That gives the registry enough time to catch up so the job stops failing on a false negative.

> - [ ] Yes
> - [x] No

**Other information**:

Fixes AAE 51793. Worst case this adds about a minute to the check step, well within its 15 minute timeout.

[AAE-51793]: https://hyland.atlassian.net/browse/AAE-51793?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


scripts/github/build/npm-check-bundles.sh
```diff
@@ -34,6 +34,28 @@ error_out() {
...
+# so a package published moments ago (e.g. late in the release-npm publish order) can
+# briefly 404 here even though the publish itself succeeded. Retry before giving up.
+NPM_VIEW_RETRIES=6
+NPM_VIEW_RETRY_DELAY=10
+
+npm_view_version() {
+    local spec=$1
+    local attempt result
+    for attempt in $(seq 1 "$NPM_VIEW_RETRIES"); do
+        result=$(npm view "$spec" version 2>/dev/null)
```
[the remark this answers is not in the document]
dsibilio: Good catch, that lookup was subject to the same propagation delay, just usually masked because adf-core publishes earlier in the batch. Added a retry loop around it too in 79e462f3a2.

ehsan-2019 (approved):

.github/actions/npm-check-bundle/action.yml
```diff
@@ -26,6 +26,20 @@ runs:
           echo "TAG_NPM not set, aborting"
           exit 1
         fi
-        ADF_VERSION=$(npm view @alfresco/adf-core@${TAG_NPM} version)
+
+        # Retry to absorb npm registry propagation delay right after publish.
+        for attempt in $(seq 1 6); do
```
ehsan-2019: please replace 6 and 10 with the constants the same way you did below

ehsan-2019 (changes requested):

.github/actions/npm-check-bundle/action.yml
dsibilio: Done ✅

ehsan-2019 (approved):

PR #12244, opened 2026-09-17
AAE-52098 Fix - Rich text input mapping stopped working

tomgny:

> - [x] Bugfix
> - [ ] Feature
> - [ ] Build related changes
> - [ ] Documentation

https://hyland.atlassian.net/browse/AAE-52098

> - [ ] Yes
> - [x] No

**Other information**:


jcain82 (approved):

lib/process-services-cloud/src/lib/form/services/form-cloud-submission-values.ts
```diff
@@ -24,6 +24,9 @@ type SubmissionRow = Record<string, unknown>;
 
 const isSubmissionRow = (value: unknown): value is SubmissionRow => typeof value === 'object' && value !== null && !Array.isArray(value);
 
+const isExpressionTemplate = (authoredValue: unknown, expressions: FormExpressionService): boolean =>
+    expressions.hasExpressions(JSON.stringify(authoredValue ?? null));
```
amolodyh-hyland: This checks the whole stringified value, but `resolveRichTextExpressions` only walks `data.text`, `data.caption`, `data.content`, and `items[].content` on list blocks. So if an expression sits anywhere else, it still passes this check, we throw away the input mapping, and nothing ever resolves it.

I ran this against the branch. I used the following value on the widget:

```json
{ "blocks": [{ "type": "code", "data": { "code": "greeting = ${field.name}" } }] }
```

with an input mapping supplying a value, and here's what got submitted:

```
{"blocks":[{"type":"code","data":{"code":"greeting = ${field.name}"}}]}
```

The mapped value is gone and `${field.name}` is still sitting there unresolved. I got the same result with an image block, since we walk `data.caption` but not `data.file.url`.

I think this only needs to check the spots the resolver actually walks:

```suggestion
const isExpressionTemplate = (authoredValue: unknown, expressions: FormExpressionService): boolean =>
    hasResolvableRichTextExpressions(authoredValue, (content) => expressions.hasExpressions(content));
```

Two more edits go with that, both outside this hunk so I can't suggest them inline. The import on line 21 picks up `hasResolvableRichTextExpressions`, and `rich-text-expression-resolver.ts` gets the helper:

```ts
export const hasResolvableRichTextExpressions = (value: unknown, hasExpressions: (content: string) => boolean): boolean => {
    let resolvable = false;

    resolveRichTextExpressions(value, (content) => {
        resolvable = resolvable || hasExpressions(content);
        return content;
    });

    return resolvable;
};
```

The callback hands `content` back untouched so it only probes, and `resolveRichTextExpressions` clones by default so we're not mutating anything. Keeping it next to the resolver is what stops these two drifting apart again.

lib/process-services-cloud/src/lib/form/components/widgets/display-rich-text/display-rich-text.widget.ts
```diff
@@ -66,49 +67,38 @@ export class DisplayRichTextWidgetComponent extends BaseDisplayTextWidgetCompone
...
 
     protected reevaluateExpressions(): void {
+        this.renderAuthoredExpressionTemplate();
+    }
+
+    private getAuthoredExpressionTemplate(): string | undefined {
+        const authoredTemplate = JSON.stringify(this.field?.authoredValue ?? null);
+
+        return this.expressions.hasExpressions(authoredTemplate) ? authoredTemplate : undefined;
+    }
```
amolodyh-hyland, line 86: Same check here, same problem on the display side. I used that same code block value and `field.value.blocks[0].data.text` came back `undefined` where I expected `Mapped text`, so we throw away the mapped value and render the default instead.

```suggestion
    private getAuthoredExpressionTemplate(): string | undefined {
        const authoredValue = this.field?.authoredValue ?? null;
        const isExpressionTemplate = hasResolvableRichTextExpressions(authoredValue, (content) => this.expressions.hasExpressions(content));

        return isExpressionTemplate ? JSON.stringify(authoredValue) : undefined;
    }
```

Line 25 needs `hasResolvableRichTextExpressions` added to the resolver import too. If you share the one helper between here and `form-cloud-submission-values.ts`, display and submission can't end up disagreeing about what counts as a template.

amolodyh-hyland (commented):
Not blocking this, the direction is right and it fixes the common case.

One thing to flag though: the new check is wider than what `resolveRichTextExpressions` actually walks, so there are still values where we drop the input mapping and leave the expression unresolved. I reproduced it with a code block and with an image URL. Details and a suggested fix inline.

amolodyh-hyland:
You can apply the fix in this PR or I'll spin up another PR after you merge this one to apply the changes.

amolodyh-hyland (approved):

PR #12256, opened 2026-09-18
AAE-48048 Fix TaskListCloud triggering duplicate identical HTTP requests

DenysVuika:
## Problem
`TaskListCloudComponent` could trigger multiple identical HTTP requests (e.g. the same `process-instances/{id}/tasks` query) on a single display, because the `combineLatest([isLoadingPreferences$, isColumnSchemaCreated$, fetchProcessesTrigger$])` pipeline re-fired the request whenever any of the three source streams emitted, even when the emitted value hadn't actually changed.

## Fix
- Added `distinctUntilChanged()` to `isLoadingPreferences$` and `isColumnSchemaCreated$` so redundant same-value re-emissions no longer re-trigger the pipeline.
- Computed the request node before `switchMap` and added a `distinctUntilChanged` comparator on it, so a new HTTP call is only made when the request payload actually changes.
- Moved the "reloading" flag `tap` to run only once a genuinely new/changed request is about to be sent, avoiding a stuck loading spinner when a duplicate emission is filtered out.

## Testing
- Updated a unit test that previously asserted a duplicate identical request occurred after a column-visibility change; it now asserts the endpoint is called once when the visibility change doesn't affect the request payload.
- `nx test process-services-cloud` for `task-list-cloud.component.spec.ts` (41/41 passing) and `service-task-list-cloud.component.spec.ts` (23/23 passing).


DudaRobert (approved):

DaryaBalvanovich (approved):

ehsan-2019 (approved):
LGTM!

DiegoMasciangelo (approved):

DiegoMasciangelo (commented):
sonar seems not happy 😆

